PR #7, opened 2021-10-08
Add URLs to component manifests

kumekay:
These URLs can be shown in the UI, also it will be simpler to find the source of the component.

The question is, whether we should point to the root of the repo, or to the component directory inside

### DIFF
--- a/esp32_azure_iot_kit/idf_component.yml
+++ b/esp32_azure_iot_kit/idf_component.yml
@@ -1,5 +1,7 @@
 version: "0.0.7-alpha"
 description: Board Support Package for ESP32-Azure-IoT-Kit
+url: https://github.com/espressif/esp-bsp/tree/master/esp32_azure_iot_kit
+
 targets:
     - esp32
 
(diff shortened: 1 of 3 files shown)


tore-espressif:
@kumekay 
Sorry, I must have missed this PR :(

Meanwhile, it was decided that some component (iperf, pcap, pid_ctrl, qrcode) will not be maintained in this repo, creating a conflict with your commit. Could you please resolve it? I'll merge it then

> whether we should point to the root of the repo, or to the component directory inside

To the component directory inside makes sense to me.

kumekay:
@tore-espressif  Sure, I updated this.

One more question, which is out of scope of this PR, 

Most dependencies are defined as `>=`, this may work, but in general is not compatible with semantic versioning that we are using.

I would suggest "^1.1.1" (>=1.1.1,<2.0.0) or "~1.1.1" (>=1.1.1,<1.2.0) requirements,

like

```
dependencies:
  idf: ">=4.2"

  button:
    version: "^1.0.0"
    public: true
  es8311:
    version: "^0.0.2-alpha"
    public: true
  led_strip:
    version: "^1.1.0-alpha"
    public: true
  lvgl/lvgl_esp32_drivers:
    version: "^0.0.3"
    public: true
```

tore-espressif:
Entered #10 to track ^^

PR #14, opened 2022-01-27
ES8311: release 1.0.0

tore-espressif:
Notable changes:

- Version bumped to 1.0.0 (first stable API)
- There is no DAC ramp-up setup during initialization
- Ramp-ups for ADC and DAC are now configurable through dedicated API
- ES8311 can accepts MCLK signal from both MCLK pin (user must define its frequency) and SCK  pin

### DIFF
--- a/components/es8311/es8311.c
+++ b/components/es8311/es8311.c
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2015-2021 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2015-2022 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -21,7 +21,7 @@ typedef struct {
 } es8311_dev_t;
 
 /*
- * Clock coefficient structer
+ * Clock coefficient structure
  */
 struct _coeff_div {
     uint32_t mclk;        /* mclk frequency */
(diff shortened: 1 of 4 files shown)


tore-espressif:
@Dazza0 @L-KAYA PTAL

components/es8311/es8311.c
```diff
@@ -370,7 +370,7 @@ esp_err_t es8311_init(es8311_handle_t dev, const es8311_clock_config_t *const cl
     ret |= es8311_write_reg(dev, ES8311_SYSTEM_REG12, 0x00); // power-up DAC - NOT default
     ret |= es8311_write_reg(dev, ES8311_SYSTEM_REG13, 0x10); // Enable output to HP drive - NOT default
     ret |= es8311_write_reg(dev, ES8311_ADC_REG1C, 0x6A); // ADC Equalizer bypass, cancel DC offset in digital domain
-    ret |= es8311_write_reg(dev, ES8311_DAC_REG37, 0x48); // DAC ramprate and bypass DAC equalizer - NOT default
+    ret |= es8311_write_reg(dev, ES8311_DAC_REG37, 0x08); // Bypass DAC equalizer - NOT default
```
L-KAYA, line 373: Can we replace these `ret |=` with general check by including `esp_check.h`? Checking the error code for each line is better than together.

L-KAYA (commented):
BTW, could you check again for the configurations defined in `coeff_div`, it seems different with the es8311 driver in `esp-adf` repo.  And, please check the clock configuration as well, last time when I use the `es8311_init` to initialize it at 16KHz(If I remember right), there is a warning saying not find the configuration.

tore-espressif:
@L-KAYA Thank you very much for the review!

>  initialize it at 16KHz(If I remember right), there is a warning saying not find the configuration.

I see,  I forgot to add support for taking MCLK signal from MCLK pin, it is added now, see d19bbed6841a1d726bf0acb3e36de7df65749b8a

> BTW, could you check again for the configurations defined in coeff_div, it seems different with the es8311 driver in esp-adf repo

Yep, there are little differences in both the coefficients and the calculating algorithm, but result is the same.

PR #15, opened 2022-02-09
Components: explicit the depencies on "driver" component

o-marshmallow:
Currently, `esp-bsp` components rely on IDF's `components/{target}` to retrieve `driver` component:
```
           REQUIRES                REQUIRES
es8311 ----------------> esp32 ----------------> driver
```  
So, when `{target}` component will stop depending on `driver` component, `es8311` and all other components will fail to build.

This PR add an explicit `REQUIRES` to avoid this issue in the future.

### DIFF
--- a/components/bh1750/CMakeLists.txt
+++ b/components/bh1750/CMakeLists.txt
@@ -1,4 +1,5 @@
 idf_component_register(
     SRCS "bh1750.c"
     INCLUDE_DIRS "include"
+    REQUIRES "driver"
 )
--- a/components/bh1750/idf_component.yml
+++ b/components/bh1750/idf_component.yml
@@ -1,4 +1,4 @@
-version: "1.0.1"
+version: "1.0.2"
 description: I2C driver for BH1750 light sensor
 url: https://github.com/espressif/esp-bsp/tree/master/components/bh1750
 dependencies:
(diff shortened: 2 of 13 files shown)


o-marshmallow:
@Dazza0 @tore-espressif @L-KAYA PTAL

components/bh1750/CMakeLists.txt
```diff
@@ -1,4 +1,6 @@
 idf_component_register(
     SRCS "bh1750.c"
     INCLUDE_DIRS "include"
+    # Requires driver for I2C header file
+    PRIV_REQUIRES "driver"
```
tore-espressif, line 5: We include `driver/i2c.h` from public header files (e.g. [bh1750](https://github.com/espressif/esp-bsp/blob/master/components/bh1750/include/bh1750.h#L13))

Shouldn't this be public `REQUIRES "driver"` too?

o-marshmallow: I wasn't sure, I will set it as public then

o-marshmallow: @tore-espressif Updated!

tore-espressif: Thanks! I will merge after CI passes

o-marshmallow:
> @o-marshmallow I forgot that we need to bump version in each `idf_component.yml` for it to get uploaded to component registry... Could you please bump the bugfix version? I will merge then

No problem, I updated the versions :+1:

PR #23, opened 2022-05-02
Fix Wrover display issues

tore-espressif:
* Add display brightness control
* Avoid long delays in LVGL task
* Turn on display on init
* Explicitly call LVGL tick increment


espzav (approved):
LGTM and tested on Wrover board.

PR #25, opened 2022-05-03
bsp: Add esp_lcd_touch and supported controllers FT5x06, GT911, TT21100

espzav:
@tore-espressif PTAL

### DIFF
--- a/components/esp_lcd_touch_gt911/idf_component.yml
+++ b/components/esp_lcd_touch_gt911/idf_component.yml
@@ -1,0 +1,6 @@
+version: "0.0.1"
+description: ESP LCD Touch GT911 - touch controller GT911
+url: https://github.com/espressif/esp-bsp/tree/master/components/esp_lcd_touch_gt911
+dependencies:
+  esp_lcd_touch:
+    version: ">=0.0.1"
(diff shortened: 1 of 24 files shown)


components/esp_lcd_touch_gt911/idf_component.yml
```diff
@@ -0,0 +1,6 @@
+version: "0.0.1"
```
tore-espressif, line 1: Applies to all `.yml` files
```suggestion
version: "0.9.0"
```

Let's start with version 0.9.0

components/esp_lcd_touch_gt911/idf_component.yml
```diff
@@ -0,0 +1,6 @@
...
+description: ESP LCD Touch GT911 - touch controller GT911
+url: https://github.com/espressif/esp-bsp/tree/master/components/esp_lcd_touch_gt911
+dependencies:
+  esp_lcd_touch:
+    version: ">=0.0.1"
```
tore-espressif, line 6: Applies to all `.yml` files

Since we will probably break the API in version 1.0.0, let's define dependency like this
```suggestion
    version: "^0.9"
```

tore-espressif (changes requested):

PR #31, opened 2022-05-10
Feature/esp lcd touch support multiple interfaces

espzav:

PR #35, opened 2022-05-16
esp_lcd_touch: Used esp_lcd rx_param/tx_param for read/write (supported multiple interfaces)

espzav:
This is only part of the PR #31, because there should be merged esp_lcd_touch before other implementations.


tore-espressif:
@espzav could you please explain this PR in greater detail? Why does it require IDF v4.2.2?

PR #37, opened 2022-05-19
mag3110: Fix including esp_timer in IDF 4.1

espzav:


espzav:
@tore-espressif PTAL

examples/rainmaker_example/main/CMakeLists.txt
```diff
@@ -1,2 +1,10 @@
+set(requires esp_rainmaker nvs_flash hts221 ssd1306 json_parser app_wifi app_reset esp32_azure_iot_kit)
```
igrr, line 1: Given that hts221 and ssd1306 are public dependencies of esp32_azure_iot_kit, is it necessary to add them explicitly?

espzav: They aren't there in CMakeLists.txt. I added only these dependencies, which was missing, when I tried build.

tore-espressif (approved):

tore-espressif:
@espzav just bump mag3110 bugfix version and we can merge

PR #39, opened 2022-05-23
ci: Override path for BSPs in examples

tore-espressif:
By using local components in examples, we make sure that we compile actual code from the specific PR (rather than already uploaded component in the component service)

### DIFF
--- a/examples/display/main/idf_component.yml
+++ b/examples/display/main/idf_component.yml
@@ -2,6 +2,6 @@ description: BSP Display example
 
 dependencies:
   idf: ">=4.4"
-  esp_wrover_kit: ">=1.0.0"
-  # esp32_s2_kaluga_kit: ">=1.0.0"
-  # esp-box: ">=1.0.0"
+  esp_wrover_kit:
+    version: ">=1.0.0"
+    override_path: "../../../esp_wrover_kit"
--- a/examples/display/main/idf_component.yml
+++ b/examples/display/main/idf_component.yml
@@ -3,5 +3,5 @@ description: BSP Display example
 dependencies:
   idf: ">=4.4"
   esp_wrover_kit:
-    version: ">=1.0.0"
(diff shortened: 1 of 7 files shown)


tore-espressif:
cc @kumekay

examples/display/main/idf_component.yml
```diff
@@ -2,6 +2,6 @@ description: BSP Display example
 
 dependencies:
   idf: ">=4.4"
-  esp_wrover_kit: ">=1.0.0"
-  # esp32_s2_kaluga_kit: ">=1.0.0"
-  # esp-box: ">=1.0.0"
+  esp_wrover_kit:
+    version: ">=1.0.0"
```
kumekay, line 6: Want be better to limit versions to compatible ones?
```suggestion
    version: "^1.0.0"
```

kumekay (commented):
LGTM, but I suggest uploading to staging first, to be sure it works as expected

espzav:
I know, that we was solving it, but I don't like see this in example YML files. Because what customer will do? He will copy the example and it wouldn't working. Then he will copy all components from BSP instead of use it from component manager. 

I know, that there are no other way. Maybe some note in YML files.

espzav:
@kumekay thank you for clarification. Then it is ok for me 👍

espzav (approved):

kumekay:
@tore-espressif I'm not too fond of the idea of fallback, to me, it's too magical, nobody reads warnings. I suggest an error with clear suggestions on what to remove from which manifest to continue. Keeping the `override_path` in the manifest forever can be error prone, because project may depend on relative path.

espzav:
I agree with @kumekay, there should be error.

tore-espressif:
> there should be error

fair enough. Can we merge it as it is now?

kumekay:
@tore-espressif LGTM

PR #41, opened 2022-05-30
bsp: Add ESP32-S3-USB-OTG board

### LINKED ISSUE #36: BSP for the new ESP32-S3 OTG Kit (BSP-146)
Cannot find it anywhere

tore-espressif:
Closes #36


esp32_s3_usb_otg/esp32_s3_usb_otg.c
```diff
@@ -0,0 +1,430 @@
...
+        .mode = GPIO_MODE_OUTPUT,
+        .pull_up_en = GPIO_PULLUP_DISABLE,
+        .pull_down_en = GPIO_PULLDOWN_DISABLE,
+        .intr_type = GPIO_INTR_DISABLE
+    };
+    ESP_ERROR_CHECK(gpio_config(&led_io_config));
+    ESP_ERROR_CHECK(gpio_set_level(BSP_USB_MODE_SEL, 1));
+}
+
+void bsp_usb_host_power_mode(bsp_usb_host_power_mode_t mode, bool limit_500mA)
```
espzav, line 329: I don't know, what about `_init `functions, but this function can be called in runtime and I think, it should return `esp_err_t `and not assert on errors. I think, that all runtime functions should returns errors instead of assert. The code can working next, if this fails.

esp32_s3_usb_otg/esp32_s3_usb_otg.c
```diff
@@ -0,0 +1,430 @@
...
+    ESP_ERROR_CHECK(adc1_config_channel_atten(ADC1_CHANNEL_0, ADC_ATTEN_DB_11));
+    ESP_ERROR_CHECK(adc1_config_channel_atten(ADC1_CHANNEL_1, ADC_ATTEN_DB_11));
+    return true;
+}
+
+uint32_t bsp_voltage_battery_get(void)
+{
+    uint32_t voltage;
+    esp_adc_cal_get_voltage(ADC1_CHANNEL_1, &adc1_chars, &voltage);
+    return voltage * 2;
```
espzav, line 422: Put constant into `#define` and put some name for this constant (why `*2` ?)

esp32_s3_usb_otg/esp32_s3_usb_otg.c
```diff
@@ -0,0 +1,430 @@
...
+}
+
+uint32_t bsp_voltage_usb_get(void)
+{
+    uint32_t voltage;
+    esp_adc_cal_get_voltage(ADC1_CHANNEL_0, &adc1_chars, &voltage);
+    return voltage * 3.7f;
```
espzav, line 429: Put constant into `#define` and put some name for this constant (why `*3.7` ?)

esp32_s3_usb_otg/include/bsp/esp32_s3_usb_otg.h
```diff
@@ -0,0 +1,321 @@
...
+ *
+ * There are 2 voltages measured on ESP32-S2-USB-OTG board:
+ * 1. Battery voltage
+ * 2. Voltage on USB device connector
+ **************************************************************************************************/
+
+/**
+ * @brief Init voltage measurements
+ *
+ * ADC is configured and calibrated
```
espzav, line 295:
```suggestion
 * ADC configuration and calibration
```
No past time

espzav (commented):
Some notes added. Not tested yet and not compilated yet. I will try compilation later.

esp32_s3_usb_otg/esp32_s3_usb_otg.c
tore-espressif: This is a good point for discussion; the BSPs are not targeting production quality code, but rather quick prototyping.

So the idea here is that we don't want to bother users with error-code checking. The app will fail at runtime with clear error message.

What do you think?

esp32_s3_usb_otg/esp32_s3_usb_otg.c
```diff
@@ -0,0 +1,430 @@
...
+static lv_disp_t *lvgl_port_display_init(void)
+{
+    ESP_LOGD(TAG, "Initialize SPI bus");
+    const spi_bus_config_t buscfg = {
+        .sclk_io_num = BSP_LCD_SPI_CLK,
+        .mosi_io_num = BSP_LCD_SPI_MOSI,
+        .miso_io_num = GPIO_NUM_NC,
+        .quadwp_io_num = GPIO_NUM_NC,
+        .quadhd_io_num = GPIO_NUM_NC,
+        .max_transfer_sz = BSP_LCD_H_RES * 80 * sizeof(uint16_t),
```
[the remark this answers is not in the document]
tore-espressif: Copy paste error. I changed the buffer sizes to 1/8 of display size. (1/10 is recommended minimum)

esp32_s3_usb_otg/esp32_s3_usb_otg.c
espzav: I understand. It's a pity, that BSP is meant only for quick prototyping and not for production code. I thought, that it can help with developping (they can use part of code). 

When I am develop new device, I am using same parts of schematic as on developer boards. Then I can use the same SW there (e.g. display). It is only my opinion.

But OK, we can leave it.

PR #42, opened 2022-06-09
bsp: Add support for screen rotating on ESP-BOX.

espzav:
This is an implementation and example for screen rotating. It can be extended with gyroscope, when it will done.


tore-espressif:
@espzav Plus, looking at the code, it should be pretty easy to add the screen rotation to all BSPs with display.
That is Kaluga , Wrover and USB_OTG.

Or would you prefer separate PRs for that?

espzav:
> @espzav Plus, looking at the code, it should be pretty easy to add the screen rotation to all BSPs with display. That is Kaluga , Wrover and USB_OTG.
> 
> Or would you prefer separate PRs for that?

@tore-espressif Yes, I can add it to another boards, but I cannot try it on some boards and the example cannot be same, because there isn't touch. There could be changed rotation via some button, I think (or accelerometer/gyroscope).

esp-box/esp-box.c
```diff
@@ -321,16 +362,22 @@ void bsp_display_backlight_on(void)
...
     lvgl_port_tick_init();
     lvgl_mux = xSemaphoreCreateMutex();
     assert(lvgl_mux);
     xTaskCreate(lvgl_port_task, "LVGL task", 4096, NULL, CONFIG_BSP_DISPLAY_LVGL_TASK_PRIORITY, NULL);
+    return disp;
+}
+
+void bsp_display_rotate(lv_disp_t *disp, lv_disp_rot_t rotation)
+{
+    lv_disp_set_rotation(disp, rotation);
```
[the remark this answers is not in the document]
espzav: Isn't it same as `bsp_button_get `?

espzav:
@tore-espressif Added all BSP boards with display. Tested on ESP-BOX and WROVER KIT.

esp-box/esp-box.c
```diff
@@ -136,6 +137,46 @@ static void lvgl_port_flush_callback(lv_disp_drv_t *drv, const lv_area_t *area,
...
+        esp_lcd_touch_set_mirror_x(tp, true);
+        break;
+    case LV_DISP_ROT_90:
+        // Rotate LCD display
+        esp_lcd_panel_swap_xy(panel_handle, true);
+        esp_lcd_panel_mirror(panel_handle, false, true);
+        // Rotate LCD touch
+        esp_lcd_touch_set_mirror_y(tp, true);
+        esp_lcd_touch_set_mirror_x(tp, false);
+        break;
```
tore-espressif, line 160: According to LVGL documentation: 

> The rotation values are relative to how you would rotate the physical display in the clockwise direction. Thus, LV_DISP_ROT_90 means you rotate the hardware 90 degrees clockwise, and the display rotates 90 degrees counterclockwise to compensate.

So the 90 and 270 deg rotations should be swapper

esp_wrover_kit/esp_wrover_kit.c
```diff
@@ -98,6 +98,51 @@ static void lvgl_port_flush_callback(lv_disp_drv_t *drv, const lv_area_t *area,
...
+        break;
+    case LV_DISP_ROT_270:
+        // Rotate LCD display
+        esp_lcd_panel_swap_xy(panel_handle, true);
+#ifdef CONFIG_BSP_LCD_ILI9341
+        esp_lcd_panel_mirror(panel_handle, true, true);
+#else
+        esp_lcd_panel_mirror(panel_handle, false, true);
+#endif
+        break;
```
tore-espressif, line 141: Same as in esp-box. 90/270 rotation swapped? 
https://docs.lvgl.io/latest/en/html/porting/display.html#rotation

tore-espressif (changes requested):
@espzav I pushed a commit with fixes for USB-OTG board and Wrover-kit.

I also left comment about rotation direction, PTAL

esp_wrover_kit/esp_wrover_kit.c
espzav: Fixed on all boards.

PR #43, opened 2022-06-10
kaluga: Add camera support

tore-espressif:
Add esp32-camera component and example


examples/display_camera/main/display_camera.c
```diff
@@ -0,0 +1,60 @@
...
+    ESP_LOGI(TAG, "Camera Init done");
+
+    // Create LVGL canvas for camera image
+    lv_obj_t *camera_canvas = lv_canvas_create(lv_scr_act());
+    assert(camera_canvas);
+    lv_obj_center(camera_canvas);
+
+    // Get first frame and initialize the canvas with it
+    camera_fb_t *pic = esp_camera_fb_get();
+    assert(pic->buf);
```
espzav, line 46: add `assert(pic)` before `assert(pic->buf)`

examples/display_camera/main/display_camera.c
```diff
@@ -0,0 +1,60 @@
...
+    lv_canvas_set_buffer(camera_canvas, pic->buf, BSP_LCD_H_RES, BSP_LCD_V_RES, LV_IMG_CF_TRUE_COLOR);
+    esp_camera_fb_return(pic);
+
+    while (1) {
+        pic = esp_camera_fb_get();
```
espzav, line 51: Must it be here? Must it be called in loop? If not, it is on the top. 
I would like see something like this, if is it possible:

```
while(1)
{
esp_camera_fb_return(pic);
ESP_LOGI(TAG, "picture: %d x %d %dbyte", pic->width, pic->height, pic->len);
lv_obj_invalidate(camera_canvas); // This will force rewrite of the canvas
}
```

If not possible, use my code and add the `esp_camera_fb_get `before `esp_camera_fb_return`. I don't know this camera API, but i looks better.

espzav (commented):
I don't know the camera API and I haven't got too much time for check it now. I put some small ideas. Then ok.

examples/display_camera/main/display_camera.c
tore-espressif: Yes, esp32-camera reads frames from the camera on the background. `esp_camera_fb_get` just returns (of of possible many) filled frame buffer.

After we are done with the processing, it must be returned to esp32-camera with `esp_camera_fb_return`

PR #44, opened 2022-07-08
Add ILI9341 and GC9A01 LCD controllers into components.

espzav:
Adding two LCD controllers into components and into component manager. 
The GC9A01 is only copied from ESP-IDF example and made the component from it. 
The ILI9341 is changed for right initialization of the controller. Tested on prepared example for ESP-IDF.


tore-espressif:
Linking related discussions https://github.com/espressif/esp-idf/pull/9311

espzav:
Waiting for fix in rainmaker: https://github.com/espressif/esp-rainmaker/issues/133

tore-espressif (approved):
@espzav LGTM!

Just two things:
1. Could you add `component/lcd` folder to https://github.com/espressif/esp-bsp/blob/master/test_app/CMakeLists.txt#L9 and check that the new components are built in the CI action?
2. Not related to this PR, but since we're here: Could you please update our CI upload action as described here https://github.com/espressif/github-actions/blob/master/upload_components/README.md#deprecated-use-espressifupload-components-ci-action-instead ?

PR #53, opened 2022-08-12
esp_jpeg: Fixed unsupported MCU of tjpgd in ROM.

espzav:
I forgot to add esp32c2 to Kconfig, that there is not tjpgd in ROM code. Fixed now.


espzav:
@tore-espressif PTAL

PR #55, opened 2022-08-16
lcd: Add lcd list with controllers and component links.

espzav:


espzav:
@tore-espressif PTAL

README.md
```diff
@@ -12,6 +12,8 @@ Board support packages for development boards using Espressif's SoCs, written in
 | [ESP32-S2-Kaluga Kit](esp32_s2_kaluga_kit) | ESP32-S2 | LCD display, audio codec + power amplifier,<br>smart LED and camera | <img src="docu/pics/kaluga.png" width="150">  |
 | [ESP32-S3-USB-OTG](esp32_s3_usb_otg) | ESP32-S3 | LCD display, uSD card slot, USB-OTG | <img src="docu/pics/esp32_s3_otg.png" width="150">  |
 
+The BSP repository includes lot of LCD and Touch driver components. The list of available LCDs is [here](LCD.md).
```
tore-espressif, line 15: Could we highlight this a little more?

LCD.md
```diff
@@ -0,0 +1,19 @@
+# LCD screen and touch controllers list
```
tore-espressif, line 1: Could we add a reference to this file (a symlink?) to `componets/lcd` and `components/lcd_touch`? What do you think?

tore-espressif (commented):
@espzav Thank you for keeping track of this. I think we should make the table more visible to users

README.md
espzav: I added a headline. I don't know, how to do it better.

LCD.md
espzav: Symlink would it be nice, but I don't know, if it will work on windows right. I tried but cannot make it. Other way can be create README file for each folder and there add section with LCD list.

igrr: Creating a README file with a few sentences and a link to LCD readme sounds like a good solution.

espzav: Added README files into lcd and lcd_touch folders.

tore-espressif (approved):

PR #56, opened 2022-08-18
Draft & discussion: SquareLine generator

espzav:
Hi all, I prepared draft of SquareLine generator after tuesday discussion. This is my first python code :-)


espzav:
@tore-espressif @igrr Here is first draft of the SquareLine generator. All information are in Readme file. Please try to understand readme, if I know, that it is understandable or not. Thank you. 
@igrr please, could you try it on Mac?

SquareLine/README.md
```diff
@@ -0,0 +1,56 @@
...
+    "placeholders":
+    {
+        "PLACEHOLDER_1": "value 1",
+        "PLACEHOLDER_2": "value 2"
+    }
+```
+
+**_NOTE 1:_** Placeholders are optional.
+
+**_NOTE 2:_** When some placeholder is found in code and not found in JSON, it will leave there in a same form.
```
igrr, line 38:
```suggestion
**_NOTE 2:_** When some placeholder is found in code and not found in JSON, it will not be substituted.
```

By the way, is this behavior intentional, is it to catch typos in placeholder names?

SquareLine/gen.py
```diff
@@ -0,0 +1,232 @@
...
+            os.remove(path + "/" + filename)
+    os.rmdir(path)
+
+class SafeDict(dict):
+    def __missing__(self, key):
+        return '{' + key + '}'
+
+def replace_placeholders(file):
+    global placeholders
+    #for placeholder in placeholders:
```
radimkarnis, line 73: Is this a remnant of the previous implementation?

SquareLine/README.md
```diff
@@ -0,0 +1,56 @@
...
+* `screen_width` - Screen width (string)
+* `screen_height` - Screen height (string)
+* `screen_color_swap` - Is color bytes swapped (boolean)
+* `short_description` - Short board description (string)
+* `long_description` - Long board description (string)
+* `placeholders` - List of placeholders and values of them (JSON object)
+
+### Placeholders
+
+During the copy all files (except `image.png`), there can be replaced some placeholder by the value. Placeholder must be in compound brackets (e.g. `{PLACEHOLDER}`). The values for placeholders are saved in `manifest.json` file without compound brackets like this:
```
igrr, line 26:
```suggestion
When copying all the files (except for `image.png`), the script will replace placeholders found in these files. Placeholder must be in compound brackets (e.g. `{PLACEHOLDER}`). The placeholders and their values should be defined in `manifest.json` as follows:
```

SquareLine/README.md
```diff
@@ -0,0 +1,56 @@
+# SquareLine board generator
+
+This is simple python generator for SquareLine board packages. This generator copy all files from `common` folder to destination ZIP file and generates other files by `manifest.json` file saved in each board directory in `boards` folder.
```
igrr, line 3:
```suggestion
This is a simple python generator script for SquareLine board packages. This script copies all files from `common` folder to the destination ZIP file. It also generates other files according to the `manifest.json` file from each board directory in `boards` folder. For information on the structure of SquareLine board packages, please refer to (add link here).
```

SquareLine/README.md
```diff
@@ -0,0 +1,56 @@
...
+* `manifest.json` - Description of the board
+* `image.png` - Image of the board 380px x 300px
+* `sdkconfig.defaults` - defaults
+* `main/idf_component.yml` - Component Registry description
+
+### `manifest.json`
+
+Mandatory JSON keys in manifest:
+* `name` - Board name (string)
+* `version` - Package version (string)
```
igrr, line 15: Should we add a note, when should this version be modified?

SquareLine/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+```
+    "placeholders":
+    {
+        "PLACEHOLDER_1": "value 1",
+        "PLACEHOLDER_2": "value 2"
+    }
+```
+
+**_NOTE 1:_** Placeholders are optional.
```
igrr, line 36: I'm not sure what "optional" means here. The note below says that if the placeholder is not defined in Json, it will not be replaced in the source file. So it looks like it's not optional?

igrr (commented):
Have reviewed the readme, overall it is quite clear! I have left a few optional suggestions.

SquareLine/gen.py
espzav: Yes, I forgot remove it. Thanks for this note.

SquareLine/README.md
espzav: Yes, I want it. Because there can be used lot of other placeholders from other systems like IDF (e.g. {IDF_PATH} in CMakeLists) and these shouldn't be replaced.

SquareLine/README.md
espzav: It means, that all JSON keys must be there except placeholders. When there is missing some other key, the script fails. When there is missing some placeholder, it continues, but not replace the placeholder.

SquareLine/gen.py
```diff
@@ -0,0 +1,232 @@
...
+# Directory of all boards
+BOARDS_DIR = "boards/"
+# Other board files, which are same for all boards (there can be used placeholders)
+COMMON_DIR = "common/"
+# JSON file which describing the board (must be in each board directory)
+BOARD_GENFILE = "manifest.json"
+# Files, which are specific for each board. Will be copied from boards/BOARD/ directory
+BOARD_SPECIFIC_FILES = {"sdkconfig.defaults", "main/idf_component.yml"}
+# Generated output directory
+OUT_DIR = "espressif"
```
igrr, line 18: Can we add this directory to `.gitignore`, so that git doesn't consider the generated content as untracked files?

SquareLine/gen.py
```diff
@@ -0,0 +1,232 @@
+from genericpath import isdir
```
igrr, line 1: `genericpath` module shouldn't normally be imported. Instead you can use `os.path.isdir`

SquareLine/gen.py
```diff
@@ -0,0 +1,231 @@
...
+        f = open(file, "r")
+        str = f.read()
+        f.close()
+        str = str.format_map(SafeDict(placeholders))
+        f = open(file, "w")
+        f.write(str)
+        f.close()
+
+# Copy file and replace placeholders in copied file
+def copy_file(src_path, dst_path, translate=1):
```
igrr, line 83: Would you consider adding `placeholders` argument to this function, to `copy_directory` and to `replace_placeholders`, to avoid having a global variable?

SquareLine/gen.py
```diff
@@ -0,0 +1,231 @@
...
+class SafeDict(dict):
+    def __missing__(self, key):
+        return '{' + key + '}'
+
+def replace_placeholders(file):
+    global placeholders
+    if os.path.exists(file):
+        f = open(file, "r")
+        str = f.read()
+        f.close()
```
igrr, line 76:
```suggestion
        with open(file, "r") as f:
            str = f.read()
```

(similar for other file operations)

Another note, `str` is a built-in type name in Python. Please try picking a different variable name to avoid overwriting the built-in type.

SquareLine/gen.py
```diff
@@ -0,0 +1,231 @@
...
+def make_zip(board_name, path):
+    zip_obj = ZipFile(path + "/" + board_name + ".zip", 'w')
+    zip_dir = path + "/" + SQUARELINE_OUTDIR
+    zip_add_folder(zip_obj, zip_dir, SQUARELINE_OUTDIR)
+    zip_obj.close()
+
+def check_json_key(manifest, key):
+    if key not in manifest:
+        print_error(f"Missing {key} in manifest JSON file")
+        os._exit(1)
```
igrr, line 117:
```suggestion
        raise SystemExit(1)
```

(same for the other cases)

SquareLine/gen.py
```diff
@@ -0,0 +1,231 @@
...
+
+def get_file_version(manifest):
+    check_json_key(manifest, "version")
+    v = manifest["version"].split('.')
+    return "_v" + v[0] + "_" + v[1] + "_" + v[2]
```
igrr, line 122:
```suggestion
    return f"_v{v[0]}_{v[1]}_{v[2]};
```

SquareLine/gen.py
```diff
@@ -0,0 +1,231 @@
...
+        output_filename = board_name + file_ver
+
+        check_json_key(manifest, "placeholders")
+        placeholders = manifest["placeholders"]
+
+        # Copy common files
+        copy_directory(COMMON_DIR, output + "/" + SQUARELINE_OUTDIR)
+        # Copy specific board files
+        for file in BOARD_SPECIFIC_FILES:
+            copy_file(dir + "/" + file, output + "/" + SQUARELINE_OUTDIR + "/" + file)
```
igrr, line 181: Suggest using `os.path.join` for building paths. It will use correct type of path separator for each OS.

SquareLine/gen.py
```diff
@@ -0,0 +1,231 @@
...
+# Print usage
+def print_help():
+    print("Not enough arguments! Usage:")
+    print(" gen.py out board")
+    print("\tout: path to the generated output")
+    print("\tboard (optional): generate only specific board")
+
+
+def main():
+    if len(sys.argv) >= 2:
```
igrr, line 201: Minor recommendation, please consider using `argparse` for parsing the command line arguments. You can check most of IDF python tools for examples.

SquareLine/gen.py
```diff
@@ -0,0 +1,231 @@
...
+            #loop all boards
+            for dirname in os.listdir(BOARDS_DIR):
+                d = os.path.join(BOARDS_DIR, dirname)
+                # checking if it is a file
+                if os.path.isdir(d):
+                    process_board(dirname, output_folder + "/" + dirname, d)
+
+        print_ok(f"Generating done! Output folder: {sys.argv[1]}")
+    else:
+        print_help()
```
igrr, line 227: Should probably also `raise SystemExit(1)` in this case so that the return code is not "success".

(although if you switch to `argparse`, it will handle this automatically)

igrr (changes requested):

igrr:
@espzav I ran the script on macOS and it worked okay!

SquareLine/gen.py
```diff
@@ -0,0 +1,231 @@
+from genericpath import isdir
+import sys
+import os
+import json
+import shutil
+from colorama import Fore
```
igrr, line 6: I also had to install the `colorama` package for this script to work. I think this can be mentioned in the readme file.

igrr:
In case it helps, here are some suggestions from `pylint`. Some of them overlap with my comments above.

<details><summary>pylint --rcfile $IDF_PATH/.pylintrc gen.py</summary>

```
************* Module gen
gen.py:230:0: W0311: Bad indentation. Found 3 spaces, expected 4 (bad-indentation)
gen.py:231:0: C0305: Trailing newlines (trailing-newlines)
gen.py:48:0: W0604: Using the global statement at the module level (global-at-module-level)
gen.py:51:13: W0622: Redefining built-in 'str' (redefined-builtin)
gen.py:55:16: W0622: Redefining built-in 'str' (redefined-builtin)
gen.py:75:8: W0622: Redefining built-in 'str' (redefined-builtin)
gen.py:72:4: W0602: Using global for 'placeholders' but no assignment is done (global-variable-not-assigned)
gen.py:74:12: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
gen.py:74:12: R1732: Consider using 'with' for resource-allocating operations (consider-using-with)
gen.py:78:12: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
gen.py:78:12: R1732: Consider using 'with' for resource-allocating operations (consider-using-with)
gen.py:109:14: R1732: Consider using 'with' for resource-allocating operations (consider-using-with)
gen.py:117:8: W0212: Access to a protected member _exit of a client class (protected-access)
gen.py:151:8: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
gen.py:151:8: R1732: Consider using 'with' for resource-allocating operations (consider-using-with)
gen.py:159:38: W0622: Redefining built-in 'dir' (redefined-builtin)
gen.py:160:4: W0601: Global variable 'placeholders' undefined at the module level (global-variable-undefined)
gen.py:166:12: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
gen.py:190:8: W0212: Access to a protected member _exit of a client class (protected-access)
gen.py:166:12: R1732: Consider using 'with' for resource-allocating operations (consider-using-with)
gen.py:1:0: W0611: Unused isdir imported from genericpath (unused-import)
```
</details>

SquareLine/gen.py
dobairoland, line 6: This is not a built-in Python module. You could mention in the readme how to install dependencies and here, catch the ImportError and print a nice error message with suggestion to solve it.

SquareLine/gen.py
```diff
@@ -0,0 +1,231 @@
...
+
+global placeholders
+
+# Print green text
+def print_ok(str):
+    print(Fore.GREEN + str + Fore.WHITE)
+
+# Print red error
+def print_error(str):
+    print(Fore.RED + str + Fore.WHITE)
```
dobairoland, line 56: If you need the external package only for this kind of simple formatting then I can suggest to try to avoid it. You can find in `idf_monitor.py` of ESP-IDF the `yellow_print` and `red_print` functions and use them here.

SquareLine/gen.py
```diff
@@ -0,0 +1,231 @@
...
+
+def remove_folder(path):
```
dobairoland, line 58: You can use `shutil.rmtree` instead which does the same.

SquareLine/gen.py
```diff
@@ -0,0 +1,231 @@
...
+    for filename in os.listdir(path):
+        f = os.path.join(path, filename)
+        if os.path.isdir(f):
+            remove_folder(path + "/" + filename)
+        else:
+            os.remove(path + "/" + filename)
+    os.rmdir(path)
+
+class SafeDict(dict):
```
dobairoland, line 67: I can suggest to comment items like this. Based on the content one cannot guess the desired functionality.

SquareLine/gen.py
dobairoland: +1

SquareLine/gen.py
```diff
@@ -0,0 +1,231 @@
...
+
+# Copy file and replace placeholders in copied file
+def copy_file(src_path, dst_path, translate=1):
+    print(f"  Copying {src_path} to {dst_path}")
+    shutil.copyfile(src_path, dst_path)
+    if translate:
+        replace_placeholders(dst_path)
+
+# Copy all files in directory
+def copy_directory(from_dir, to_dir):
```
dobairoland, line 90: `shutil.copytree()`?

SquareLine/gen.py
```diff
@@ -0,0 +1,231 @@
...
+        out_f = os.path.join(to_dir, filename)
+        if os.path.isdir(f):
+            copy_directory(f, out_f)
+        else:
+            copy_file(f, out_f)
+
+def zip_add_folder(zip_obj, path, outdir):
+    for filename in os.listdir(path):
+        file = os.path.join(path, filename)
+        zip_obj.write(file, outdir + "/" + os.path.basename(file))
```
dobairoland, line 104: It is better to use `os.path.join` everywhere instead of `+ "/" +`. (repeats many times)

SquareLine/gen.py
dobairoland: or `sys.exit(1)` if you don't print a backtrace.

dobairoland (changes requested):

dobairoland:
Nice work! I've left some suggestions. It looks some of them are duplicates. Sorry for that.

SquareLine/gen.py
igrr: (unless we add `placeholders` argument to `copy_directory` as suggested above)

espzav: `copy_file` and `copy_directory` are not only copy, they will substitute all placeholders in each file.

SquareLine/gen.py
igrr: The only usage seems to be `remove_folder(output + "/" + SQUARELINE_OUTDIR)`,  seems like it's okay to remove the entire SQUARELINE_OUTDIR directory? Why do we need to keep some subdirectories?

espzav:
@dobairoland, @igrr, @radimkarnis thank you very much for the review. It is very valuable for me. 
I pushed fixes and left some comments.

SquareLine/gen.py
espzav: Yes, I need only remove `out/board/SQUARELINE_OUTDIR` with everything inside. But when I do `shutil.rmtree(out/board/SQUARELINE_OUTDIR)`, it removes all, `SQUARELINE_OUTDIR` and then `board` and then `out` directory, which contains ZIP and other files.

igrr: Huh, that's a pretty odd behavior. Let me check...

espzav:
@igrr please, is it ok now? Next week should be release of the SquareLine Studio. We should finish this before it. Which boards would it be supported? Now there are two boards and I tested on HW only ESP-BOX.

igrr (approved):
Looks good to me, please feel free to merge.

PR #61, opened 2022-09-06
kaluga: Use new led_strip driver

tore-espressif:
* Use new led_strip driver in kaluga BSP
* Build test_app with pedantic compiler flags


espzav (approved):
LGTM

PR #67, opened 2022-10-04
lvgl: ESP portation of the LVGL.

espzav:

- [x] Component contains idf_component.yml file with `url` field defined
- [x] CI passing

# Change description
This is the first draft of the LVGL portation component. It can help to initialize LVGL for any project. There is solved rotation, monochrome display usage and more... 

Discussion:
- [x] Save the initial status of the display rotation!
- [ ] The name of component?


components/esp_lvgl_port/Kconfig
```diff
@@ -0,0 +1,14 @@
...
+        default 4
+        help
+            The component will create a task that will periodically handle LVGL operation in lv_timer_handler().
+
+    config LVGL_PORT_TICK_PERIOD_MS
+    int "LVGL tick period [ms]"
+    default 5
+    range 1 100
+    help
+        Period of LVGL tick timer.
```
tore-espressif, line 13: We don't have to do it now; just an idea for future improvement:

We should once-and-for-all find a way how to get time to LVGL without periodic interrupts. e.g. with `esp_timer_get_time`

components/esp_lvgl_port/README.md
```diff
@@ -0,0 +1,89 @@
...
+        .rotation = {
+            .swap_xy = false,
+            .mirror_x = false,
+            .mirror_y = false,
+        }
+    };
+    lv_disp_t * disp = lvgl_port_add_disp(&disp_cfg);
+```
+
+**Note 1:** The component calls `esp_lcd_panel_draw_bitmap` API for send data to the screen. There must be called `lvgl_port_flush_ready(disp)` after each transaction to display. The best way is to use `on_color_trans_done` callback from `esp_lcd` IO config structure.
```
tore-espressif, line 46: 'Flush ready' should be handled inside of the lvgl_port.
I can see that currently this is hardly possible now... we talk about it later

tore-espressif (commented):
(not a full review)

I had a list of LVGL features that would require port to ESP. We don't need to implement it all now; it is just something to think about:

1. lv_log: map LVGL logs to ESP_LOGx macros
2. lv_decoder: various image decoders (JPEG, PNG....)
3. tick_timer: use interrupt-less solution. (take time from RTC with esp_timer_get_time?)
4. lv_filesystem https://docs.lvgl.io/latest/en/html/overview/file-system.html
5. Memory management: LVGL does not support full deinitialization with dynamic memory allocation. Q: is this important? should we document it?
6. Backlight control

components/esp_lvgl_port/esp_lvgl_port.c
```diff
@@ -0,0 +1,418 @@
...
+    disp_ctx->rotation.swap_xy = disp_cfg->rotation.swap_xy;
+    disp_ctx->rotation.mirror_x = disp_cfg->rotation.mirror_x;
+    disp_ctx->rotation.mirror_y = disp_cfg->rotation.mirror_y;
+
+    /* alloc draw buffers used by LVGL */
+    /* it's recommended to choose the size of the draw buffer(s) to be at least 1/10 screen sized */
+    buf1 = heap_caps_malloc(disp_cfg->buffer_size * sizeof(lv_color_t), MALLOC_CAP_DMA);
+    ESP_GOTO_ON_FALSE(buf1, ESP_ERR_NO_MEM, err, TAG, "Not enough memory for LVGL buffer (buf1) allocation!");
+    if (disp_cfg->double_buffer) {
+        buf2 = heap_caps_malloc(disp_cfg->buffer_size * sizeof(lv_color_t), MALLOC_CAP_DMA);
```
suda-morris, line 135: it will be feasible to also allow the user to provide the draw buffer instead of we allocating for them

suda-morris: and the draw buffer is not always needed to be DMA capable

espzav: I would like to add flags like: `ESP_LVGL_PORT_FLAG_BUFF_DMA`, `ESP_LVGL_PORT_FLAG_BUFF_SPIRAM` and create buffer based these flags.

components/esp_lvgl_port/README.md
espzav: I added note into readme. We need some example too.

espzav:
I was testing this component with SH1107 display and SSD1306. You can try modified example for SSD1306 on IDF repository on branch `feature/support_rotating_lcd_ssd1306`. Only copy this component to example (I will add into YML later, when this component will be merged).

tore-espressif (changes requested):
Not tested on HW yet

components/esp_lvgl_port/esp_lvgl_port.c
```diff
@@ -0,0 +1,421 @@
...
+    lvgl_port_display_ctx_t *disp_ctx = (lvgl_port_display_ctx_t *)drv->user_data;
+    assert(disp_ctx != NULL);
+    esp_lcd_panel_handle_t panel_handle = disp_ctx->handle;
+
+    /* Solve rotation screen and touch */
+    switch (drv->rotated) {
+    case LV_DISP_ROT_NONE:
+        /* Rotate LCD display */
+        esp_lcd_panel_swap_xy(panel_handle, disp_ctx->rotation.swap_xy);
+        esp_lcd_panel_mirror(panel_handle, disp_ctx->rotation.mirror_x, disp_ctx->rotation.mirror_y);
```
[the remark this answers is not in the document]
espzav: No, it is not necessary.

espzav:
@tore-espressif Thank you for review. Everything fixed. The IDF i2c_oled example was extended for using with ssh1107 too (you can select between ssd1306 and sh1107) - MR will be later after this will be done.

components/esp_lvgl_port/README.md
```diff
@@ -0,0 +1,114 @@
...
+    * Create task and timer
+    * Handle rotating
+* Add/remove display (using [`esp_lcd`](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/lcd.html))
+* Add/remove touch input (using [`esp_lcd_touch`](https://github.com/espressif/esp-bsp/tree/master/components/lcd_touch))
+
+## Usage
+
+### Initialization
+```
+    esp_err_t err = lvgl_port_init();
```
tore-espressif, line 16: This function must accepts `lvgl_port_cfg_t *` argument

components/esp_lvgl_port/esp_lvgl_port.c
espzav: Added.

tore-espressif:
@espzav LGTM. I'll have a final look once the CI passes.
And, please don't forget about the points in PR description checklist.

components/esp_lvgl_port/esp_lvgl_port.c
```diff
@@ -0,0 +1,455 @@
...
+
+    /* LVGL init */
+    lv_init();
+    /* Tick init */
+    lvgl_port_timer_period_ms = cfg->timer_period_ms;
+    ESP_RETURN_ON_ERROR(lvgl_port_tick_init(), TAG, "");
+    /* Create task */
+    lvgl_port_ctx.lvgl_mux = xSemaphoreCreateMutex();
+    ESP_GOTO_ON_FALSE(lvgl_port_ctx.lvgl_mux, ESP_ERR_NO_MEM, err, TAG, "Create LVGL mutex fail!");
+    BaseType_t res = xTaskCreate(lvgl_port_task, "LVGL task", cfg->task_stack, NULL, cfg->task_priority, NULL);
```
atanisoft, line 92: Can a config option be added for pinning this task to a specific core (default can be floating)?

components/esp_lvgl_port/include/esp_lvgl_port.h
```diff
@@ -0,0 +1,165 @@
...
+extern "C" {
+#endif
+
+/**
+ * @brief Init configuration structure
+ */
+typedef struct {
+    int task_priority;      /*!< LVGL task priority */
+    int task_stack;         /*!< LVGL task stack size */
+    int timer_period_ms;    /*!< LVGL ttimer tick period in ms */
```
atanisoft, line 24: `ttimer` -> `timer`

components/esp_lvgl_port/include/esp_lvgl_port.h
```diff
@@ -0,0 +1,165 @@
...
+} lvgl_port_cfg_t;
+
+/**
+ * @brief Rotation configuration
+ */
+typedef struct {
+    bool swap_xy;
+    bool mirror_x;
+    bool mirror_y;
```
atanisoft, line 33: should these have docs?

atanisoft:
@espzav I've done some testing using this PR locally and it seems to work great with my ILI9488 driver (SPI 4-wire + DC).

I have however observed some rather odd interactions between `esp_lcd_touch` and other components which appear to be related to usage of `portMUX_TYPE`, I'll be logging an issue once I narrow it down a bit further.

espzav:
@atanisoft Thank you very much for the review. I fixed all issues and added task pin to core selectable in menuconfig. 
Please, do you have the ILI9488 driver with esp_lcd? I should do it (this table https://github.com/espressif/esp-bsp/blob/master/LCD.md)

components/esp_lvgl_port/Kconfig
```diff
@@ -0,0 +1,15 @@
+menu "ESP LVGL PORT"
+
+    config ESP_LVGL_PORT_TASK_CORE_EN
+        bool "Enable pin LVGL task to selected core"
```
atanisoft, line 4: Any reason for Kconfig vs part of `ESP_LVGL_PORT_INIT_CONFIG`? Also this option should be hidden / disabled when `FREERTOS_UNICORE` is enabled

atanisoft:
> do you have the ILI9488 driver with esp_lcd?

https://components.espressif.com/components/atanisoft/esp_lcd_ili9488

I'll also be publishing [esp_lcd_touch_xpt2046](https://github.com/atanisoft/esp_lcd_touch_xpt2046) fairly soon.

components/esp_lvgl_port/Kconfig
espzav: I thought about it, if I should put it into init configuration structure and I think, that this is better solution for this selection. It cannot be selected in runtime. I will add FREERTOS_UNICORE depend to Kconfig.

espzav:
> > do you have the ILI9488 driver with esp_lcd?
> 
> https://components.espressif.com/components/atanisoft/esp_lcd_ili9488
> 
> I'll also be publishing [esp_lcd_touch_xpt2046](https://github.com/atanisoft/esp_lcd_touch_xpt2046) fairly soon.

That's great! Thank you very much. I will add it into our LCD table soon. I'd like test in on my HW before.

atanisoft:
> I'd like test in on my HW before.

sounds good, I've only tested it with SPI interface as I don't have one with the parallel interface (yet). One thing I have observed is with LVGL there are occasional artifacts along the edges of moving items (such as a spinner). I wasn't seeing these prior to moving over to `esp_lvgl_port` so it may just be a bug in my config for the same.

components/esp_lvgl_port/Kconfig
espzav: I thought, that the configuration structure is only for realtime settings and this seems to be ok in compilation settings only. I didn't know, that we have it in another place in configuration structure. Should I change it?

tore-espressif: If the option can be both in menuconfig or run-time config, run-time config is preferred. (Same with task's stack size, and task's priority ...).

So yes, please change it unless there is a reason to have it in menuconfig :)

espzav: Ok, changed.

components/esp_lvgl_port/include/esp_lvgl_port.h
```diff
@@ -0,0 +1,167 @@
...
+extern "C" {
+#endif
+
+/**
+ * @brief Init configuration structure
+ */
+typedef struct {
+    int task_priority;      /*!< LVGL task priority */
+    int task_stack;         /*!< LVGL task stack size */
+    int pin_to_core;        /*!< LVGL task pinned to core (-1 is floating) */
```
tore-espressif, line 24:
```suggestion
    int task_affinity;        /*!< LVGL task pinned to core (-1 is no affinity) */
```

components/esp_lvgl_port/README.md
```diff
@@ -0,0 +1,145 @@
...
+### Initialization
+```
+    const lvgl_port_cfg_t lvgl_cfg = ESP_LVGL_PORT_INIT_CONFIG();
+    esp_err_t err = lvgl_port_init(&lvgl_cfg);
+```
+
+### Add screen (IDF 5.1)
+
+Add an LCD screen to the LVGL. It can be called multiple times for adding multiple LCD screens. 
+```
```
tore-espressif, line 23:
```suggestion
``` c
```

This will add nice color to the rendered code

components/esp_lvgl_port/README.md
```diff
@@ -0,0 +1,145 @@
...
+        .flags = {
+            .buff_dma = true,
+        }
+    };
+    lv_disp_t * disp = lvgl_port_add_disp(&disp_cfg);
+```
+
+**Note:** The screens added in this function are not removed in `lvgl_port_deinit`. They must be removed by `lvgl_port_remove_disp` before deinitialization. Otherwise, there can be memory leaks!
+
+### Add screen (IDF 5.0 and IDF 4.4)
```
tore-espressif, line 53: It took me a while to spot the differences between IDF v5.1 and this chapter. 

Would it be better if we highlighted only the code that is different? That is, only the lcd_io configuration part?

tore-espressif (approved):
Few nitpicks, otherwise LGTM!

components/esp_lvgl_port/README.md
espzav: Agree, but I don't know, how to highlight for good visibility changes?

tore-espressif: Sorry, poor wording on my side :D

I was thinking about leaving the lcd_io configuration part only in this chapter. And have a comment like: 'The rest of the configuration is the same as in IDF v5.1'

PR #72, opened 2022-10-12
Add initial read into touch controllers, where is missing it.

espzav:
Added some initial read into touch controllers for check, if I2C is working.


espzav:
> What do you think about updating related BSPs (esp-box) like this: In case the touch initialization fails, the LVGL initialization continues, but informs the user that touch is not working (a log message?)
> 

I don't know, why should the touch initialization fails, but we can do it. We can do it after the esp_lvgl_port will be done and we will refactor all BSPs with it.

PR #82, opened 2022-10-26
Add new main component - esp_io_expander

Lzw655:
# Checklist for new Board Support package or Component

- [x] Component contains License
- [x] Component contains README.md
- [x] Project [README.md](../README.md) updated
- [x] Component contains idf_component.yml file with `url` field defined
- [ ] Component was added to CI [upload job](https://github.com/espressif/esp-bsp/blob/master/.github/workflows/upload_component.yml#L17)
- [ ] New files were added to CI build job
- [ ] _Optional:_ Component contains unit tests
- [ ] CI passing

# Change description
This is a new main component to driver io expander chip. I also created two devices based on it.

PR #86, opened 2022-10-27
fix some error in esp_io_expander

Lzw655:

- [x] Project [README.md](../README.md) updated
- [x] Component contains idf_component.yml file with `url` field defined
- [x] CI passing

# Change description
The last version of esp_io_expander is not perfect. Fixed some errors according to comments in  #82 .

### DIFF
--- a/components/io_expander/esp_io_expander/esp_io_expander.c
+++ b/components/io_expander/esp_io_expander/esp_io_expander.c
@@ -13,11 +13,7 @@
 
 #include "esp_io_expander.h"
 
-/* Pin num must be smaller than 32 (bit width of uint32_t) */
-#define PIN_NUM_MAX(handle)                 (((handle)->config.io_amount < IO_EXPANDER_PIN_NUM_MAX) ?  \
-                                             (handle)->config.io_amount : (IO_EXPANDER_PIN_NUM_MAX - 1))
-/* Check whether it is a valid pin number */
-#define IS_VALID_PIN_NUM(pin_num, handle)   ((pin_num) < PIN_NUM_MAX(handle))
+#define VALID_IO_COUNT(handle)      ((handle)->config.io_count <= IO_COUNT_MAX ? (handle)->config.io_count : IO_COUNT_MAX)
 
 /**
  * @brief Register type
@@ -34,11 +30,12 @@ static char *TAG = "io_expander";
 static esp_err_t write_reg(esp_io_expander_handle_t handle, reg_type_t reg, uint32_t value);
 static esp_err_t read_reg(esp_io_expander_handle_t handle, reg_type_t reg, uint32_t *value);
 
-esp_err_t esp_io_expander_set_dir(esp_io_expander_handle_t handle, esp_io_expander_pin_num_t pin_num, esp_io_expander_dir_t direction)
+esp_err_t esp_io_expander_set_dir(esp_io_expander_handle_t handle, uint32_t pin_num_mask, esp_io_expander_dir_t direction)
 {
-    ESP_RETURN_ON_FALSE(
-        IS_VALID_PIN_NUM(pin_num, handle), ESP_ERR_INVALID_ARG,
-        TAG, "Invalid pin_num(%d), max(%d)", pin_num, PIN_NUM_MAX(handle));
+    ESP_RETURN_ON_FALSE(handle, ESP_ERR_INVALID_ARG, TAG, "Invalid handle");
+    if (pin_num_mask >= BIT(VALID_IO_COUNT(handle))) {
+        ESP_LOGW(TAG, "Pin num mask out of range, bit higher than %d won't work", VALID_IO_COUNT(handle) - 1);
+    }
 
     bool is_output = (direction == IO_EXPANDER_OUTPUT) ? true : false;
     uint32_t dir_reg, temp;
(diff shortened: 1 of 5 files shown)


Lzw655:
> Thank you for changes. LGTM, please fix the test issue - there is missing end of comment `*/`

Thank you to point out the error, it's been fixed.

Lzw655:
BTW, should I change it's the version to v1.0.1?

espzav:
> BTW, should I change it's version to v1.0.1?

Yes, you should change the version before merge.

espzav (approved):
Please, change the version, after that, it is ok! Thank you

components/io_expander/esp_io_expander/include/esp_io_expander.h
```diff
@@ -19,15 +19,60 @@ extern "C" {
...
+    IO_EXPANDER_PIN_NUM_20 = 20,
+    IO_EXPANDER_PIN_NUM_21 = 21,
+    IO_EXPANDER_PIN_NUM_26 = 26,
+    IO_EXPANDER_PIN_NUM_27 = 27,
+    IO_EXPANDER_PIN_NUM_28 = 28,
+    IO_EXPANDER_PIN_NUM_29 = 29,
+    IO_EXPANDER_PIN_NUM_30 = 30,
+    IO_EXPANDER_PIN_NUM_31 = 31,
+    IO_EXPANDER_PIN_NUM_MAX,
+} esp_io_expander_pin_num_t;
```
tore-espressif, line 59: What about making this type a bitmask? This way, the user could set state of several pins at once

components/io_expander/esp_io_expander/include/esp_io_expander.h
```diff
@@ -19,15 +19,60 @@ extern "C" {
...
+    IO_EXPANDER_OUTPUT,         /*!< Output dircetion */
+} esp_io_expander_dir_t;
+
 /**
  * @brief IO Expander Configuration Type
  *
  */
 typedef struct {
-    uint8_t io_amount;                      /*!< Amout of device's IO, must be smaller than 32 */
+    uint8_t io_amount;                      /*!< Amout of device's IO, must be less than `IO_EXPANDER_PIN_NUM_MAX` */
```
tore-espressif, line 75:
```suggestion
    uint8_t io_count;                      /*!< Count of device's IO, must be less or equal than `IO_EXPANDER_PIN_NUM_MAX` */
```

Please use 'count' with countable nouns.

Should it be less or equal? `io_count = 32;` is valid, isn't it?

components/io_expander/esp_io_expander/esp_io_expander.c
```diff
@@ -14,9 +14,10 @@
 #include "esp_io_expander.h"
 
 /* Pin num must be smaller than 32 (bit width of uint32_t) */
-#define PIN_NUM_MAX(handle)                         (handle->config.io_amount > 32 ? 32 : handle->config.io_amount)
+#define PIN_NUM_MAX(handle)                 (((handle)->config.io_amount < IO_EXPANDER_PIN_NUM_MAX) ?  \
+                                             (handle)->config.io_amount : (IO_EXPANDER_PIN_NUM_MAX - 1))
```
tore-espressif, line 18: Is the calculation correct here? Why can't I set io_amount to 32?

tore-espressif (changes requested):

components/io_expander/esp_io_expander/esp_io_expander.c
Lzw655: It's wrong. Fixed.

Lzw655: @tore-espressif PTAL the new commit. Thank you very much.

components/io_expander/esp_io_expander/esp_io_expander.c
```diff
@@ -33,23 +30,25 @@ static char *TAG = "io_expander";
...
-esp_err_t esp_io_expander_set_dir(esp_io_expander_handle_t handle, uint8_t pin_num, bool is_output)
+esp_err_t esp_io_expander_set_dir(esp_io_expander_handle_t handle, uint32_t pin_num_mask, esp_io_expander_dir_t direction)
 {
-    ESP_RETURN_ON_FALSE(
-        IS_VALID_PIN_NUM(pin_num, handle), ESP_ERR_INVALID_ARG,
-        TAG, "Invalid pin_num(%d), max(%d)", pin_num, PIN_NUM_MAX(handle));
+    ESP_RETURN_ON_FALSE(handle, ESP_ERR_INVALID_ARG, TAG, "Invalid handle");
+    if (pin_num_mask >= BIT(VALID_IO_COUNT(handle))) {
+        ESP_LOGW(TAG, "Pin num mask out of range, bit higher than %d won't work", VALID_IO_COUNT(handle) - 1);
+    }
```
tore-espressif, line 38: If the `pin_num_mask` equals `0xFF00_0000` its value is greater then `BIT(32)`, which will print the warning on next line.

This condition is present on multiple places in this file, please check those too.

Lzw655: You're right. I should use `BIT64` instead of `BIT`. These conditions have been fixed in the new [commit](https://github.com/espressif/esp-bsp/pull/86/commits/d9cc48fb6b5929df2fcab8be9aa1952b69472ea6).

tore-espressif (approved):

tore-espressif:
@Lzw655 Thank you for the fixes! 
BTW, I removed version 1.0.0 from the registry

Lzw655:
> @Lzw655 Thank you for the fixes! BTW, I removed version 1.0.0 from the registry

Ok, thank you for reviewing.

PR #88, opened 2022-11-02
lcd_touch: Update FreeRTOS critical sections to use port*_CRITICAL macros instead of task*_CRITICAL macros

sudeep-mohanty:
lcd_touch: Update FreeRTOS critical sections to use port*_CRITICAL macros instead of task*_CRITICAL macros

- [ ] CI passing

# Change description
Currently the esp_lcd_touch_xx files use the taskENTER_CRITICAL and taskEXIT_CRITICAL macros to establish a critical section. They use the macros with a local spinlock variable. This is compatible with existing FreeRTOS in ESP-IDF. However, the same macros do not accept a spinlock variable when Amazon FreeRTOS-SMP is used instead of IDF-FreeRTOS. Therefore, to maintain compatibility across both kernel versions, this commit updates the critical sections to use the portENTER_CRITICAL and portEXIT_CRITICAL macros which provide the flexibility to pass local spinlock variables irrespective of the FreeRTOS kernel type used in ESP-IDF.


espzav (approved):
LGTM

tore-espressif:
> LGTM

@sudeep-mohanty Thank you very much for the update!

@espzav Should we bump esp_lcd_touch version too?

espzav:
> @espzav Should we bump esp_lcd_touch version too?

Only version of each esp_lcd_touch_xxx should be increased.

sudeep-mohanty:
@tore-espressif @espzav Thank you for the review! I have now incremented the _minor_ version of the `esp_lcd_touch_xxx` components that were updated in the PR in their respective manifest files. PTAL again and do let me know in case the version update should be done differently! Thanks!

espzav:
> @tore-espressif @espzav Thank you for the review! I have now incremented the _minor_ version of the `esp_lcd_touch_xxx` components that were updated in the PR in their respective manifest files. PTAL again and do let me know in case the version update should be done differently! Thanks!

@sudeep-mohanty Thank you. As it is no breaking change and API change, I think there should be increased only the third number. What do you thing @tore-espressif ?

tore-espressif:
> I think there should be increased only the third number

I agree. We usually use minor version for new (non-breaking) features and bugfix version for changes like this

sudeep-mohanty:
Thanks for clarification @tore-espressif. I have amended the PR to update the third number in the version now. 
(PS: I was under the impression that the third number in the version string is referred as the _patch_ version and the second number is the _minor_ version and hence the confusion 🙂)

PR #92, opened 2022-11-25
Square line/esp32 s3 lcd ev board support (BSP-230)

bubblesnake:

- [ ] CI passing

# Change description
This PR adds support to generate SquareLine project for ESP32_S3_LCD_EV_BOARD.


Lzw655:
@tore-espressif @espzav PTAL, Thank you very much!

espzav (changes requested):
@bubblesnake Thank you for the SquareLine project for EV board! I will test it today. I have some notes now:
- [ ] The image should have the resolution: 380x300px (https://docs.squareline.io/docs/obp)
- [ ] Do you add the second board with 480x480 LCD screen? 
- [ ] We changed `short_description` and `long_description` in other boards. Could you change both too, please? It seems, that your long description can be short and you can add something more to long description.

bubblesnake:
> @bubblesnake Thank you for the SquareLine project for EV board! I will test it today. I have some notes now:
> 
> * [ ]  The image should have the resolution: 380x300px (https://docs.squareline.io/docs/obp)
> * [ ]  Do you add the second board with 480x480 LCD screen?
> * [ ]  We changed `short_description` and `long_description` in other boards. Could you change both too, please? It seems, that your long description can be short and you can add something more to long description.

@espzav , thanks for the review comments.
I would like to clarify this point:
> Do you add the second board with 480x480 LCD screen?
Do you mean to add another board project for the 480x480 screen? Or we let the user to change screen size in the same SquareLine project?

espzav:
> Do you mean to add another board project for the 480x480 screen? Or we let the user to change screen size in the same SquareLine project?

I think, that there will be better to put the same project with another name (e.g. add the screen size to name of both board) and another image with another size screen settings. What do you think?

bubblesnake:
> > Do you mean to add another board project for the 480x480 screen? Or we let the user to change screen size in the same SquareLine project?
> 
> I think, that there will be better to put the same project with another name (e.g. add the screen size to name of both board) and another image with another size screen settings. What do you think?

Ok got it. I think doable. However in SquareLine Studio, there will be two different project templates. On the other hand in each project setting, user is able to change the screen resolution (bottom right corner). Will that make user confused?

image

espzav:
> Ok got it. I think doable. However in SquareLine Studio, there will be two different project templates. On the other hand in each project setting, user is able to change the screen resolution (bottom right corner). Will that make user confused? In fact the esp32_s3_lcd_ev_board right now can support up to 3 different screen sizes.

I know, that there will be two same boards. But image and name of the boards tels the right LCD. 
I see, that we support only two sizes of LCD here: https://github.com/espressif/esp-bsp/tree/master/esp32_s3_lcd_ev_board

espzav:
@bubblesnake One additional think - if there will be two boards for each LCD size, you can create another sdkconfig.defaults for each. How do you change the LCD controller from SquareLine?

SquareLine/boards/esp32_s3_lcd_ev_board/partitions.csv
```diff
@@ -0,0 +1,6 @@
+# Name,   Type, SubType, Offset,  Size, Flags
+# Note: if you have increased the bootloader size, make sure to update the offsets to avoid overlap
+nvs,      data, nvs,     0x9000,   0x6000,
+phy_init, data, phy,     ,         0x1000,
+factory,  app,  factory, ,         3500k,
+storage,  data, spiffs,  ,         7000k,
```
espzav, line 6: This file is not copied into SquareLine project.

bubblesnake:
> 

1 and 3 corrected. PTAL, thanks.

bubblesnake:
> 

In this link, I can see there are 3 more different LCD with different LCD controllers that not yet supported, and may be added in the future. In my opinion, having too many boards in the SquareLine projects is cumbersome. We can have a single board, use one of the screen in the image, and tell user how to change configurations. What do you think?

bubblesnake:
> > @bubblesnake Thank you for the SquareLine project for EV board! I will test it today. I have some notes now:
> 
> > The image should have the resolution: 380x300px (https://docs.squareline.io/docs/obp)
> > Do you add the second board with 480x480 LCD screen?
> > We changed short_description and long_description in other boards. Could you change both too, please? It seems, that your long description can be short and you can add something more to long description.
> 
> 1 and 3 corrected. PTAL, thanks.

bubblesnake:
> @bubblesnake One additional think - if there will be two boards for each LCD size, you can create another sdkconfig.defaults for each. How do you change the LCD controller from SquareLine?

@espzav , I agree. If there is one board only, we have to let user change the LCD controller in menuconfig. However it is the same for espwroverkit, which supports two LCD controllers. But we have only one SquareLine project.

SquareLine/boards/esp32_s3_lcd_ev_board/partitions.csv
bubblesnake: > This file is not copied into SquareLine project.

This file is copied into SquareLine project together with other files in the folder during generation. During testing it is found that a custom partition table is required even with a very basic UI.

espzav: Are you sure? I checked generated files and it isn't there. But we can add it to `gen.py`

espzav:
> > @bubblesnake One additional think - if there will be two boards for each LCD size, you can create another sdkconfig.defaults for each. How do you change the LCD controller from SquareLine?
> 
> @espzav , I agree. If there is one board only, we have to let user change the LCD controller in menuconfig. However it is the same for espwroverkit, which supports two LCD controllers. But we have only one SquareLine project.

@tore-espressif What do you thing?

SquareLine/boards/esp32_s3_lcd_ev_board/partitions.csv
bubblesnake: I'm so sorry. I think my memory is wrong as I can't reproduce it myself! I will try to add it to gen.py. Thanks.

bubblesnake: Done

espzav: Thank you for this change! Great!

tore-espressif (approved):

bubblesnake:
> All changes tested on HW (LCD 480x480). Generated SquareLine board, made SquareLine project and flashed to bard. Everything working properly. @bubblesnake Thank you for your work! Great job!

Thank you all for your review!

PR #96, opened 2022-12-06
CI: Add doxygen check

tore-espressif:
# Change description

Add doxygen build job to CI

### DIFF
--- a/.github/workflows/doxygen.yml
+++ b/.github/workflows/doxygen.yml
@@ -1,0 +1,16 @@
+name: Build Doxygen documentation
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+  push:
+    branches:
+      - master
+
+jobs:
+  build-docu:
+    name: Build Doxygen documentation
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - uses: mattnotmitt/doxygen-action@v1.9.5
--- a/Doxyfile
+++ b/Doxyfile
@@ -1,0 +1,18 @@
+PROJECT_NAME           = "ESP-BSP"
+PROJECT_BRIEF          = "Board support packages for development boards using Espressif's SoCs, written in C."
+OUTPUT_DIRECTORY       = ./doxygen_output
+OPTIMIZE_OUTPUT_FOR_C  = YES
+WARN_IF_UNDOCUMENTED   = NO
+INPUT                  = ./
+RECURSIVE              = YES
+EXCLUDE                = docu \
+                         examples \
+                         SquareLine \
+                         test_app
+EXCLUDE_PATTERNS       = */build/* \
+                         */managed_components/* \
+                         */dist/*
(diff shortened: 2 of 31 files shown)


tore-espressif:
@igrr @espzav PTAL.

I'm not very experienced with Doxygen, but the output seems pretty good for a first try. You can find it here: https://github.com/espressif/esp-bsp/suites/9697954427/artifacts/464694116

Doxyfile
```diff
@@ -0,0 +1,2774 @@
...
+IGNORE_PREFIX          =
+
+#---------------------------------------------------------------------------
+# Configuration options related to the HTML output
+#---------------------------------------------------------------------------
+
+# If the GENERATE_HTML tag is set to YES, doxygen will generate HTML output
+# The default value is: YES.
+
+GENERATE_HTML          = YES
```
igrr, line 1293: We probably don't need HTML output, just XML one is enough?

Doxyfile
```diff
@@ -0,0 +1,2774 @@
...
+GENERATE_LEGEND        = YES
+
+# If the DOT_CLEANUP tag is set to YES, doxygen will remove the intermediate
+# files that are used to generate the various graphs.
+#
+# Note: This setting is not only used for dot files but also for msc temporary
+# files.
+# The default value is: YES.
+
+DOT_CLEANUP            = YES
```
igrr, line 2774: I think the default Doxyfile is a bit too long to work with. Since we don't need most of Doxygen features (just the XML generation) I would suggest starting from something simpler; for example, https://github.com/espressif/esp-idf/blob/master/docs/doxygen/Doxyfile or https://github.com/espressif/esp-modbus/blob/master/docs/Doxyfile.

components/bh1750/include/bh1750.h
```diff
@@ -4,6 +4,11 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+/**
+ * @file
+ * @brief BH1750 driver
+ */
```
igrr, line 10: Note that file-level comments are only used when Doxygen generates HTML output; when we use Doxygen XML files along with Breathe or another tool, this will probably not be used.

.github/workflows/doxygen.yml
```diff
@@ -0,0 +1,19 @@
...
+      - master
+
+jobs:
+  build-docu:
+    name: Build Doxygen documentation
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - uses: mattnotmitt/doxygen-action@v1.9.5
+      - uses: actions/upload-artifact@v2
```
igrr, line 17: Not sure if this is necessary; as in, is this artifact used anywhere?

espzav:
@tore-espressif I tried and check HTML output. It looks nice.

tore-espressif (commented):
Thank you for the review @igrr all fixed

components/bh1750/include/bh1750.h
igrr: Sure, we can keep that.

PR #100, opened 2022-12-12
Added badges with versions to READMEs of components

XDanielPaul:
## Description
Added badges with versions in README files to following components:

- esp-box
- esp-32_azure_iot_kit
- esp32_azure_iot_kit
- esp32_s3_eye
- esp32_s3_lcd_ev_board
- esp32_s3_usb_otg
- esp_wrover_kit

## Edit
Added badge with latest version in README files to remaining components in [components](https://github.com/espressif/esp-bsp/tree/master/components) subfolder and it's subfolders

### DIFF
--- a/components/bh1750/README.md
+++ b/components/bh1750/README.md
@@ -1,5 +1,7 @@
 # Component: BH1750
 
+[![Component Registry](https://components.espressif.com/components/espressif/bh1750/badge.svg)](https://components.espressif.com/components/espressif/bh1750)
+
 * This component will show you how to use I2C module read external i2c sensor data, here we use BH1750 light sensor(GY-30 module).
 * BH1750 measurement mode:
     * one-time mode: bh1750 just measure only one time when receieved the one time measurement command, so you need to send this command when you want to get intensity value every time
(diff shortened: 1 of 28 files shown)


igrr:
It would have been slightly nicer if the badge said `Latest version: vX.Y.Z` instead of `Version:  vX.Y.Z`. If we were to look at some old commit like https://github.com/espressif/esp-bsp/tree/8e739dd9ef35e58a06bf91226b4da5f7c5d92db8/esp_wrover_kit, the badge would still say `Version: 1.3.1` while the actual version of the component in that commit would be 1.2.1.

(Not a blocker for merging this PR, since the badge is defined on the IDF Component Registry side.)

igrr (approved):

XDanielPaul:
> thank you @XDanielPaul

No problem!

tore-espressif:
We have also lots of components in [components](https://github.com/espressif/esp-bsp/tree/master/components) subfolder, but we can update that later.

tore-espressif (approved):
Thank you @XDanielPaul

PR #101, opened 2022-12-13
esp-box: add home button handle (BSP-234)

espressif2022:

- [ ] CI passing

# Change description

- add TaskHandle judgment to prevent mutex deadlock.

- add the num of framebuf lines and quantity definition.
    application is out of memory

- add touch button reading function.
    only the read interface is added, and user handles this button in the application

- modify GPIO error.


tore-espressif:
Thank you very much @espressif2022 !
Could you pleas bump version in [here](https://github.com/espressif/esp-bsp/blob/master/esp-box/idf_component.yml) to 2.2.0 ?

espressif2022:
> Thank you very much @espressif2022 ! Could you pleas bump version in [here](https://github.com/espressif/esp-bsp/blob/master/esp-box/idf_component.yml) to 2.2.0 ?

already done.

esp-box/esp-box.c
```diff
@@ -155,6 +155,26 @@ static void bsp_touchpad_read(lv_indev_drv_t *indev_drv, lv_indev_data_t *data)
     }
 }
 
+static void bsp_button_read(lv_indev_drv_t *indev_drv, lv_indev_data_t *data)
+{
+    uint8_t g_home_btn_val;
+    static uint8_t last_btn = 0;
```
tore-espressif, line 161: Does this variable ever change to 'not zero'?

esp-box/esp-box.c
```diff
@@ -323,7 +343,9 @@ static lv_disp_t *lvgl_port_display_init(void)
 static esp_err_t lvgl_port_indev_init(void)
 {
     static lv_indev_drv_t indev_drv_tp;
+    static lv_indev_drv_t indev_drv_btn;
```
tore-espressif, line 346: How is the user supposed to assign this button to specific coordinates on the screen, if the handler is not exposed to him?

See docs https://docs.lvgl.io/latest/en/html/porting/indev.html#button

tore-espressif (changes requested):
@espressif2022 The button press is correctly detected on my ESP-BOX. However, I don't see a way how to map this button to my GUI application.

@espzav PTAL too.

esp-box/esp-box.c
espressif2022: > Does this variable ever change to 'not zero'?
Probably not. It is more appropriate to delete here.

esp-box/esp-box.c
espressif2022: > How is the user supposed to assign this button to specific coordinates on the screen, if the handler is not exposed to him?
> 
> See docs https://docs.lvgl.io/latest/en/html/porting/indev.html#button
> 
> and mainly this function [lv_indev_set_button_points](https://docs.lvgl.io/8.3/overview/indev.html?highlight=lv_indev_set_button_points#_CPPv426lv_indev_set_button_pointsP10lv_indev_tA_K10lv_point_t)

yes, it't use 'lv_indev_set_button_points' to virtual a fixed coordinate area
`lv_obj_get_click_area(btn_return, &a);lv_indev_set_button_points(ui_get_button_indev(), points_array);`

espzav:
@espressif2022 Thank you for this PR. It is good work, but I don't understand, why we need use this button to click on some coordinates on screen. I thought, that the button can be used as a back button or home button or something like that. For this feature, we don't need button which simulate touch on some coordinates.
It can be disabled and used the button in other way?

espressif2022:
> @espressif2022 Thank you for this PR. It is good work, but I don't understand, why we need use this button to click on some coordinates on screen. I thought, that the button can be used as a back button or home button or something like that. For this feature, we don't need button which simulate touch on some coordinates. It can be disabled and used the button in other way?

Your question is the function of this point ? It is defined in our [specifications](https://github.com/espressif/esp-box/blob/master/docs/getting_started.md).
Physically, this button point is not within the effective touch range[0,0-320,240], it can't be treated as a coordinate point.

esp-box/esp-box.c
espzav: @espressif2022 Please, where can I find the function `ui_get_button_indev()`. Have you got any working example please?

espzav:
> Your question is the function of this point ? It is defined in our [specifications](https://github.com/espressif/esp-box/blob/master/docs/getting_started.md). Physically, this button point is not within the effective touch range[0,0-320,240], it can't be treated as a coordinate point.

I understand. Ok. But when I want to use BSP and I don't want to map it to some coordinates, there should be some disable feature (e.g. Kconfig).

esp-box/esp-box.c
espressif2022: > 

You can refer to this [file](https://github.com/espressif/esp-box/blob/master/examples/factory_demo/main/gui/ui_about_us.c)
It find the handle register by `lv_indev_drv_init`, here is `indev_drv_btn`. 
This is not a standard TP, the general TP won't have this button

espzav: I still don't see any chance, how to get `indev_drv_btn` for use it in BSP example. There is missing simillar function like `ui_get_button_indev()` in BSP.

espressif2022: > I still don't see any chance, how to get `indev_drv_btn` for use it in BSP example. There is missing simillar function like `ui_get_button_indev()` in BSP.

I'm a bit confused that what we need to do here about this button event?
BSP reads the button event, how to handle this event is upto the application?

espzav: Application missing pointer to `indev_drv_btn`. Which is needed by `lv_indev_set_button_points(...)`.

espressif2022: > Application missing pointer to `indev_drv_btn`. Which is needed by `lv_indev_set_button_points(...)`.

If `indev_drv_btn` is null, application cann't execute `lv_indev_set_button_points`, home button is useless.

tore-espressif:
Hi @espressif2022 , 

Unfortunately, myself and @espzav still don't see a way how a user of this BSP can map the button to his GUI application.

Could you please provide us with working/compilable project or code snippet, that would demonstrate the button's usage?

Thank you for your patience!

espressif2022:
> Hi @espressif2022 ,
> 
> Unfortunately, myself and @espzav still don't see a way how a user of this BSP can map the button to his GUI application.
> 
> Could you please provide us with working/compilable project or code snippet, that would demonstrate the button's usage?
> 
> Thank you for your patience!

I draw a description as follows, it's ok？
image

tore-espressif:
Hello @espressif2022 thank you for your explanation.

Yes, this will technically work.

Just a question: Wouldn't it be better to add the button to the [bsp_button API](https://github.com/espressif/esp-bsp/blob/master/esp-box/include/bsp/esp-box.h#L91)? This way, the user could get the state of the button with [bsp_button_get](https://github.com/espressif/esp-bsp/blob/master/esp-box/include/bsp/esp-box.h#L91) and then decide what to do next.

What I'm trying to say is that the fact that we map the button as `indev` to LVGL prevents us from using the button like 'normal button'. The typical usecase does not need to map the button to coordinates on the display. What do you think?

cc @espzav

espressif2022:
> 

Hello, @tore-espressif @espzav 
Here, the home button can be placed on the application, and then user uses a timer to poll the event change, and the user decides how to handle it.
In fact, I also have a ESP32-S3-BOX-Lite board with 3 button, without TP. If the button is not registered in the lv_indev_drv of lvgl, applicaton will be very inconvenient.
I want to put these two bsps in this one package manager, and manage all these buttons here（ADC, GPIO, touch button），Currently there is a local commit. I submit it in this PR, let's discuss how to deal with it is better?

esp-box/esp-box.c
```diff
@@ -218,8 +219,8 @@ static lv_disp_t *bsp_display_lcd_init(void)
     const lvgl_port_display_cfg_t disp_cfg = {
         .io_handle = io_handle,
         .panel_handle = panel_handle,
-        .buffer_size = BSP_LCD_H_RES * 50,
-        .double_buffer = true,
+        .buffer_size = BSP_LCD_H_RES * BSP_LCD_DRAW_BUF_HEIGHT,
+        .double_buffer = BSP_LCD_DRAW_BUF_DOUBLE,
         .hres = BSP_LCD_H_RES,
         .vres = BSP_LCD_V_RES,
         .monochrome = false,
```
espzav, line 226: I am sorry, I wasn't enough test rotation before, there are bad values in screen rotation initialization. There should be this:
```
/* Rotation values must be same as used in esp_lcd for initial settings of the screen */
        .rotation = {
            .swap_xy = false,
            .mirror_x = true,
            .mirror_y = true,
        },
```

Could you please fix in this review?

esp-box/esp-box.c
```diff
@@ -370,3 +371,22 @@ bool bsp_button_get(const bsp_button_t btn)
 {
     return !(bool)gpio_get_level(btn);
 }
+
+bool bsp_touch_button_get(void)
```
espzav, line 375: I don't like this new function. I rather to see add `BSP_BUTTON_MAIN` into enum `bsp_button_t`. And then handle it in `bsp_button_get` (in condition).

esp-box/esp-box.c
```diff
@@ -370,3 +371,22 @@ bool bsp_button_get(const bsp_button_t btn)
...
+{
+    uint8_t home_btn_val = 0x00;
+
+    assert(indev);
+    lv_indev_drv_t *indev_drv = indev->driver;
+    assert(indev_drv);
+    lvgl_port_touch_ctx_t *touch_ctx = (lvgl_port_touch_ctx_t *)indev_drv->user_data;
+    assert(touch_ctx);
```
espzav, line 383: This is not working! Please put `static esp_lcd_touch_handle_t tp;` out to top of file and then use it like this:
```
    uint8_t home_btn_val = 0x00;
    assert(tp);

    esp_lcd_touch_get_button_state(tp, 0, &home_btn_val);
```

espzav (changes requested):
Cannot be compiled. I suggested some solution and left some comments.

esp-box/esp-box.c
espressif2022: ok, I will modify it tomorrow and submit it together

> I am sorry, I wasn't enough test rotation before, there are bad values in screen rotation initialization. There should be this:
> 
> ```
> /* Rotation values must be same as used in esp_lcd for initial settings of the screen */
>         .rotation = {
>             .swap_xy = false,
>             .mirror_x = true,
>             .mirror_y = true,
>         },
> ```
> 
> Could you please fix in this PR?

ok, I will modify it tomorrow and submit it together

espzav (commented):
Great work! Thank you. Only one comment left and LGTM.

espzav (approved):

esp-box/esp-box.c
```diff
@@ -56,6 +56,7 @@ static const char *TAG = "ESP-BOX";
 #endif
 
 static lv_disp_t *disp;
+static esp_lcd_touch_handle_t tp;   // LCD touch handle
```
igrr, line 59: Not a blocker for this PR, just for the future: it is recommended to use the `s_` prefix for static variables at file scope.

(https://docs.espressif.com/projects/esp-idf/en/latest/esp32/contribute/style-guide.html#naming)

esp-box/esp-box.c
```diff
@@ -368,5 +367,14 @@ esp_err_t bsp_button_init(const bsp_button_t btn)
 
 bool bsp_button_get(const bsp_button_t btn)
 {
-    return !(bool)gpio_get_level(btn);
+    if (btn == BSP_BUTTON_MAIN) {
+        uint8_t home_btn_val = 0x00;
+        assert(tp);
+
+        esp_lcd_touch_get_button_state(tp, 0, &home_btn_val);
```
igrr, line 374: Do we need to wrap this part with `#ifdef CONFIG_ESP_LCD_TOUCH_MAX_BUTTONS > 0`?

Alternatively, `_Static_assert(CONFIG_ESP_LCD_TOUCH_MAX_BUTTONS > 0, "Touch buttons must be supported for this BSP");`

esp-box/include/bsp/esp-box.h
```diff
@@ -273,6 +274,9 @@ esp_err_t bsp_spiffs_unmount(void);
 #define BSP_LCD_PIXEL_CLOCK_HZ     (40 * 1000 * 1000)
 #define BSP_LCD_SPI_NUM            (SPI3_HOST)
 
+#define BSP_LCD_DRAW_BUF_HEIGHT     (10)
```
igrr, line 277: This changes the buffer height from 50 to 10. Was this supposed to be part of the commit related to home button handling?

igrr (approved):
A few minor comments, LGTM otherwise. Feel free to address them in a follow-up PR.

esp-box/include/bsp/esp-box.h
espzav: @igrr Thank you for this point. I overlooked it - I don't understand, why it was changed in this PR. It is working with 50 too.

esp-box/esp-box.c
espzav: I think, that we didn't use it in all BSPs.

esp-box/esp-box.c
espzav: I agree, there should be `#ifdef` or `#warning`.

esp-box/include/bsp/esp-box.h
espressif2022: Because this bsp is mainly adapted to esp-idf 5.0. Compared with 4.4, the sram reserved for application is smaller. 
50 lines will lead to insufficient sram, and the program cannot run normally now.

igrr: Sounds fair, but could you put that change into a separate commit, with the message explaining why this is changed?  It would help someone looking at `git blame` output later if this change isn't part of an unrelated commit.

espressif2022: ok, i'll rearrange it, thanks.

esp-box/esp-box.c
```diff
@@ -368,5 +371,16 @@ esp_err_t bsp_button_init(const bsp_button_t btn)
...
+#if (CONFIG_ESP_LCD_TOUCH_MAX_BUTTONS > 0)
+    if (btn == BSP_BUTTON_MAIN) {
+        uint8_t home_btn_val = 0x00;
+        assert(tp);
+
+        esp_lcd_touch_get_button_state(tp, 0, &home_btn_val);
+
+        return home_btn_val ? true : false;
+    } else {
+        return !(bool)gpio_get_level(btn);
```
espzav, line 383: I think, this shopuldn't be inside the `#if (CONFIG_ESP_LCD_TOUCH_MAX_BUTTONS > 0)`. 
My suggestion:
```
    if (btn == BSP_BUTTON_MAIN) {
#if (CONFIG_ESP_LCD_TOUCH_MAX_BUTTONS > 0)
        uint8_t home_btn_val = 0x00;
        assert(tp);

        esp_lcd_touch_get_button_state(tp, 0, &home_btn_val);
        return home_btn_val ? true : false;
#else
       ESP_LOGE(TAG, "Some error message.....");
       return false;
#endif
    } else {
        return !(bool)gpio_get_level(btn);
    }
```

espressif2022: Good suggestion, please check the latest one again, thanks.

tore-espressif (approved):

PR #103, opened 2022-12-19
SquareLine: Added license file.

espzav:
Add license file into each board package for SquareLine.

### DIFF
--- a/SquareLine/common/LICENSE
+++ b/SquareLine/common/LICENSE
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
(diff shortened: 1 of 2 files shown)


espzav:
@igrr PTAL

igrr:
LGTM, but let's wait for @kisvegabor to confirm the location of the file.

espzav:
> LGTM, but let's wait for @kisvegabor to confirm the location of the file.

@igrr Thank you. Now, the LICENSE file will be in root package for each board and in ZIP file too. I think, it can be in ZIP file, because ZIP file is used for generating project folder and the LICENSE will be copied there too.

kisvegabor:
We need the license in the top level in each repo. So next to the `.slb` file.

However, adding it to the zip is also a good idea.

espzav:
> We need the license in the top level in each repo. So next to the `.slb` file.
> 
> However, adding it to the zip is also a good idea.

It will be in the top level after this PR. 
@igrr I think, we can merge it, do you get me approve please?

igrr (approved):

kisvegabor:
Thank you, I'll add the licenses to our repos.

PR #115, opened 2023-01-06
esp_lvgl_port: lvgl_port_lock deadlock (BSP-249)

espressif2022:

- [ ] CI passing

# Change description
 LVGL and external user event will operate the same API, `lvgl_port_lock` is used to protect this API.
 When lvgl event calls this kind of API, it will cause a secondary lock, cause deadlock.

### DIFF
--- a/components/esp_lvgl_port/esp_lvgl_port.c
+++ b/components/esp_lvgl_port/esp_lvgl_port.c
@@ -89,7 +89,7 @@ esp_err_t lvgl_port_init(const lvgl_port_cfg_t *cfg)
     lvgl_port_timer_period_ms = cfg->timer_period_ms;
     ESP_RETURN_ON_ERROR(lvgl_port_tick_init(), TAG, "");
     /* Create task */
-    lvgl_port_ctx.lvgl_mux = xSemaphoreCreateMutex();
+    lvgl_port_ctx.lvgl_mux = xSemaphoreCreateRecursiveMutex();
     ESP_GOTO_ON_FALSE(lvgl_port_ctx.lvgl_mux, ESP_ERR_NO_MEM, err, TAG, "Create LVGL mutex fail!");
 
     BaseType_t res;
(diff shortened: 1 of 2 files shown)


tore-espressif (commented):
@espressif2022 Thank you for pointing this out.

However, I don't think this is a proper solution. There is a path in the `lvgl_port_lock` function, that does not lock anything, which is confusing.In this case, the developer should know which functions are being called from LVGL context, that does not need any locking protection. I.e. LVGL events (or tasks) do not need to call the lock function as they are being called from LVGL locked context.

We can add an `assert` to the lock function that would check whether this task has already taken the lock.

espressif2022:
> @espressif2022 Thank you for pointing this out.
> 
> However, I don't think this is a proper solution. There is a path in the `lvgl_port_lock` function, that does not lock anything, which is confusing.In this case, the developer should know which functions are being called from LVGL context, that does not need any locking protection. I.e. LVGL events (or tasks) do not need to call the lock function as they are being called from LVGL locked context.
> 
> We can add an `assert` to the lock function that would check whether this task has already taken the lock, if you find it useful

hello @tore-espressif 
Now some applications will have the following logic:
LVGL task `->` Click event or others `->` cb -> fun_xx_public(modify lvgl widget)
APP event `->` fun_xx_public(modify lvgl widget)

Many examples use the above operations, and I hope to add this modification to match these existing designs.
Otherwise, many related changes will be involved.

I also don't recommend having external events to control LVGL widget directly. 
Some other GUI framework also prohibit such operations, and each widget function have an assert judgment on whether it is its own task.

espressif2022:
@tore-espressif 
Hello, will this modification be considered next? 
Do you have any suggestions?

tore-espressif:
> Hello, will this modification be considered next?
Do you have any suggestions?

Hello @espressif2022 thank you for your patience.

We discussed this usecase again. Could you please change the lock type to 'recursive mutex'? It should allow the LVGL task to take the mutex multiple times, allowing you to call the same function from your application code. 

Please let me know whether it works for you and whether you agree, thank you!

espressif2022:
> > Hello, will this modification be considered next?
> > Do you have any suggestions?
> 
> Hello @espressif2022 thank you for your patience.
> 
> We discussed this usecase again. Could you please change the lock type to 'recursive mutex'? It should allow the LVGL task to take the mutex multiple times, allowing you to call the same function from your application code.
> 
> Please let me know whether it works for you and whether you agree, thank you!

Ok, I'll modify and test it out. Thanks for your suggestion

tore-espressif (approved):
Thank you @espressif2022 !

Is this ready for merging?

PR #117, opened 2023-01-07
Add MPU6050 ISR support (BSP-251)

### LINKED ISSUE #109: Adding ISR support to MPU6050 sensor module (FEATURE) (BSP-241)
I have used the MPU6050 component in a custom project, and it works great. However, I believe that there are a few capabilities of the sensor not yet implemented by the component, such as interrupt configuration and handling.

One of the features of my project involved waking up the ESP32 from deep sleep whenever the MPU6050 detected motion. Initially, I wanted to implement it using the ULP to communicate directly with the MPU6050 via I2C, but then I found out about "Motion detection" interrupts which can be enabled on the sensor to trigger a wakeup from EXT0 or EXT1. Based on the [MPU6050 datasheet](https://netsonic.fi/en/files/MPU6050-datasheet.pdf) and [this post from the Arduino Stack Exchange](https://arduino.stackexchange.com/questions/48424/how-to-generate-hardware-interrupt-in-mpu6050-to-wakeup-arduino-from-sleep-mode), I learned that registers `0x1F`, `0x37`, `0x38 `and `0x3A `can be used to configure interrupt behavior of the MPU6050. Yet, as far as I know, the component provided in this repo does not support reading or writing to these registers. In the end, I implemented an extension module to support `DATA_READY` and `MOTION` interrupts for my project. 

Does it make sense to enhance the MPU6050 component to make it capable of enabling interrupts in the MPU6050, registering ISR handlers and getting the interrupt status? If so, I would like to contribute to the component with the work that I've already implemented. I would greatly appreciate your opinion on this. 

(I am opening this issue since the CONTRIBUTING.md mentions that an issue should be created before contributing large(ish) scale changes)

Frozen-Burrito:
# Change description

- Include new functions to setup and enable interrupts produced by the MPU6050 on different events (mainly `DATA_READY` at the moment, see notes below).
- Add new definitions for registers, types and enumerations used to configure and handle interrupts triggered by the MPU6050.
- Update component README.md with more information about the sensor, mention ISR support and add links to the register map and the sensors example.

# Notes
Since the driver does not yet support FiFo or I2C sensor slaves, interrupts by `FIFO_OVERFLOW` and `I2C_MST_INT` are not produced by the MPU6050. If necessary, these driver features could be implemented in the future. 

I decided not to include `MOTION_DETECT` interrupt support yet, because there is conflicting information on this feature from different MPU6050 register map and datasheet revisions.

Closes #109

### DIFF
--- a/components/mpu6050/README.md
+++ b/components/mpu6050/README.md
@@ -1,6 +1,40 @@
-# Component: MPU6050
+# MPU6050 Driver Component
 
 [![Component Registry](https://components.espressif.com/components/espressif/mpu6050/badge.svg)](https://components.espressif.com/components/espressif/mpu6050)
 
- * C driver for MPU6050 gyroscope and accelerometer.
- * See [datasheet](https://invensense.tdk.com/wp-content/uploads/2015/02/MPU-6000-Datasheet1.pdf).
+C driver for Invensense MPU6050 6-axis gyroscope and accelerometer based on I2C communication.
+
+## Features
+
+- Get 3-axis accelerometer and 3-axis gyroscope data, either raw or as floating point values. 
+- Read temperature from MPU6050 internal temperature sensor.
+- Configure gyroscope and accelerometer sensitivity.
+- MPU6050 power down mode.
+- Support for MPU6050 interrupt generation when data ready (occurs each time a write to all sensor data registers has been completed).  
+
+## Important Notes
+
+- Keep in mind that MPU6050 I2C address depends on the level of its AD0 pin (9) (0x68 when low, 0x69 when high).
+- In order to receive MPU6050 interrupts, its INT pin (12) must be conneced to a GPIO on the ESP32. 
+
+## Limitations
+
+- Only I2C communication is supported.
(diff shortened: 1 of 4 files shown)


tore-espressif (approved):
Very nice work @Frozen-Burrito , I'd like to experiment with it this week and I'll merge (or comment) then. 
Thank you!

Frozen-Burrito:
Ok, thank you for your help! I have been considering providing an example to showcase ISR support for the mpu6050, once the changes have been merged. Should I update the already existing "sensors example", while still using the timer-based read? Or would it be better to provide a separate example for this?

tore-espressif:
> Ok, thank you for your help! I've been thinking about providing an example to showcase ISR support for the mpu6050, once the changes have been merged. Should I update the already existing "sensors example", while still using the timer-based read? Or would it be better to provide a separate example for this?

That's a good idea, unfortunately the Azure-IoT-kit does not have the ISR pin routed to ESP32. So i'm just wondering how would the user actually try this new feature. 

Otherwise LGTM!

PR #124, opened 2023-02-06
esp_lvgl_port: Add public for esp_lcd_touch

espzav:
Added public for esp_lcd_touch in esp_lvgl_port for using LCD touch header outside.

### DIFF
--- a/components/esp_lvgl_port/include/esp_lvgl_port.h
+++ b/components/esp_lvgl_port/include/esp_lvgl_port.h
@@ -14,8 +14,11 @@
 #include "esp_err.h"
 #include "esp_lcd_panel_io.h"
 #include "esp_lcd_panel_ops.h"
-#include "esp_lcd_touch.h"
 #include "lvgl.h"
+
+#if __has_include ("esp_lcd_touch.h")
+#include "esp_lcd_touch.h"
+#endif
 
 #ifdef __cplusplus
 extern "C" {
@@ -59,6 +62,7 @@ typedef struct {
     } flags;
 } lvgl_port_display_cfg_t;
 
+#if __has_include ("esp_lcd_touch.h")
 /**
  * @brief Configuration touch structure
  */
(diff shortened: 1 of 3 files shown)


espzav:
@tore-espressif PTAL, there is only one small change.

tore-espressif:
Thanks, @espzav . Could you please add some reasoning behind this change?

I think that for majority of usecases esp_lcd_touch should not be public. If the users needs it for some custom purposes, he can always add the `esp_lcd_touch` as dependency to his `main` component

espzav:
> Thanks, @espzav . Could you please add some reasoning behind this change?
> 
> I think that for majority of usecases esp_lcd_touch should not be public. If the users needs it for some custom purposes, he can always add the `esp_lcd_touch` as dependency to his `main` component

I am working on change BSPs to use LVGL port and I had problem with Kaluge kit, which has only LCD without touch. But the touch header is used in `esp_lvgl_port.h`, which is included in BSP.

tore-espressif:
> I am working on change BSPs to use LVGL port and I had problem with Kaluge kit, which has only LCD without touch. But the touch header is used in `esp_lvgl_port.h`, which is included in BSP.

I see. Could you please check if we can use `__has_include` macros, to configure `esp_lvgl_port` without touch? https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005finclude.html
Something like this:
``` c
#if __has_include ("esp_lcd_touch.h")
#  include "esp_lcd_touch.h"
#endif

...
#if __has_include ("esp_lcd_touch.h")
/* Touch specific code is defined here */
#endif
```

If it is not possible, we can merge this, as it is

espzav:
@tore-espressif thank you very much for suggestion! I reworked, it should be better. Thank you very much!

components/esp_lvgl_port/include/esp_lvgl_port.h
```diff
@@ -59,13 +62,15 @@ typedef struct {
...
  * @brief Configuration touch structure
  */
 typedef struct {
     lv_disp_t *disp;    /*!< LVGL display handle (returned from lvgl_port_add_disp) */
     esp_lcd_touch_handle_t   handle;   /*!< LCD touch IO handle */
 } lvgl_port_touch_cfg_t;
+#endif
 
 /**
  * @brief Touch IO configuration structure
```
tore-espressif, line 76: Not related to your changes, but could you please update this comment?

tore-espressif (approved):
One off-topic comment, otherwise LGTM!

components/esp_lvgl_port/include/esp_lvgl_port.h
espzav: Thank you. Fixed.

PR #125, opened 2023-02-08
esp_lvgl_port: Add maximum time in ms for sleep LVGL task.

espzav:
There is another minor change in LVGL port, which we need for `esp32_s3_eye` BSP. We need to not sleep during the LVGL task. I added a maximum time in ms, for which the task can sleep.


espzav:
> @espzav Could you please set the timeout to 500, in case the user passes 0? This will ensure backward compatibility for cases where users do not use `ESP_LVGL_PORT_INIT_CONFIG()` macro

Good point! Fixed. Thank you @tore-espressif

tore-espressif (approved):
LGTM!

PR #126, opened 2023-02-08
bsp: Use LVGL port for all BSPs with display.

espzav:
Used LVGL port component in all BSPs with display (except RGB display).


esp-box/esp-box.c
```diff
@@ -18,47 +18,14 @@
 #include "bsp/esp-box.h"
 #include "esp_lcd_touch_tt21100.h"
 #include "esp_lvgl_port.h"
+#include "bsp_err_check.h"
```
tore-espressif, line 22: Could we also increase the default draw buffer size? base on our performance measuremens?

tore-espressif (approved):
LGTM!

Only Azure Kit and ESP-BOX don't have updated  versions

esp-box/esp-box.c
espzav: I think, that we want to do it in some review before, but here it is: https://github.com/espressif/esp-bsp/pull/101#pullrequestreview-1238794515

tore-espressif: Ah, OK

espzav:
@tore-espressif Thank you  for review!

PR #127, opened 2023-02-10
bsp: Fixed compilation error missing lcd touch header in EV board.

espzav:
After the last changes, the EV board is not working (not compilable).


Lzw655:
@espzav Hi, I'm improving the s3-lcd-ev-board component, which will be uploaded next week.

PR #128, opened 2023-02-10
lcd_touch: Add pulbic for all touch components.

espzav:
Some types from esp_lcd_touch is used and this component must be as public in all touch components.


tore-espressif (approved):
Thank you for the quick fix

PR #129, opened 2023-02-13
Bsp/s3 lcd ev board improve

Lzw655:

- [ ] CI passing

# Change description
* Configurations:
    * Support to set number of frame buffers for RGB
    * Support to use bounce-buffer for RGB
    * Support to change the minimum delay time of LVGL task
    * Support to change the stack size of LVGL task
    * Support to enable anti-tearing function simply by using multiple buffers
* RGB Anti-tearing:
    * Optimize direct mode
    * Use triple-buffer in full-refresh mode

Also, update README and sdkconfig of `display_lvgl_demos` example.


tore-espressif:
@Lzw655 thanks for the PR. I haven't tried it on HW yet, though

@suda-morris could you please have a look too?

Lzw655:
> @Lzw655 thank you for this review. I tried on HW with 800x480 and it is working good. But I don't understand other settings of the RGB (buffers, tearing etc.). When I tried change settings, in all combination was really worst that default settings (in FPS). Other LGTM.
> 
> Next note, it will be nice to add RGB LCD support into LVGL port. I would like do it in future, but I must to more understand before it.

Thank you for testing it! When using the anti-tearing function, compared with the default settings, it will definitely lead to a lower FPS. Because this function uses PSRAM as the frame buffer while default settings use SRAM. And, I also expect RGB LCD support in LVGL port.

Lzw655:
@espzav @tore-espressif Can we merge it now? After this, we will update examples and the factory bin in esp-dev-kits.

PR #131, opened 2023-02-15
esp_lvgl_port: Backport changes for IDF v4.4

espzav:
Updated LVGL port component with latest backported changes in IDF 4.4.4


espzav:
@tore-espressif PTAL

components/esp_lvgl_port/esp_lvgl_port.c
```diff
@@ -23,12 +23,6 @@
 #include "esp_lcd_touch.h"
 #endif
 
-#if ESP_IDF_VERSION < ESP_IDF_VERSION_VAL(5, 0, 1)
```
igrr, line 26: You still need to check if IDF is 4.4.4 or earlier (e.g. v4.4.3), don't you?

components/esp_lvgl_port/esp_lvgl_port.c
```diff
@@ -23,7 +23,7 @@
...
+#if ESP_IDF_VERSION < ESP_IDF_VERSION_VAL(4, 4, 4)
```
tore-espressif, line 26: Unfortunately, the feature with esp-lcd `esp_lcd_panel_io_register_event_callbacks` is not included in v5.0.0. 

So it should work on:  <v4.4.4, v5.0.0) ∪ <v5.0.1, ∞)

tore-espressif: The CI did not catch it because we test on release _branches_ rather than on release _tags_

tore-espressif (changes requested):
We must make sure that the code is compatible with esp-idf tag v5.0

components/esp_lvgl_port/esp_lvgl_port.c
espzav: Yes, fixed.

components/esp_lvgl_port/esp_lvgl_port.c
espzav: You are right. Fixed.

espzav:
Tested with master, v5.0, v4.4.4, v4.4.3

tore-espressif (approved):
LGTM!

PR #132, opened 2023-02-16
esp_lvgl_port: Add support for buttons as a input device.

espzav:
I am working on BSP for ESP-BOX-Lite and this board hasn't got touch screen. I decided for use the buttons for control LVGL objects and the best place for handling is LVGL port component. Here is my implementation.

### DIFF
--- a/components/esp_lvgl_port/README.md
+++ b/components/esp_lvgl_port/README.md
@@ -33,7 +33,7 @@ This part is necessary only in IDF 5.0 and older:
     static bool notify_lvgl_flush_ready(esp_lcd_panel_io_handle_t panel_io, esp_lcd_panel_io_event_data_t *edata, void *user_ctx)
     {
         lv_disp_t ** disp = (lv_disp_t **)user_ctx;
-        lvgl_port_flush_ready(*disp);
+        lvgl_port_flush_ready((*disp)->driver);
         return false;
     }
     
@@ -76,9 +76,12 @@ Main part of the code (IDF version independent):
         }
     };
     disp = lvgl_port_add_disp(&disp_cfg);
-```
+    
+    /* ... the rest of the initialization ... */
 
-**Note:** The screens added in this function are not removed in `lvgl_port_deinit`. They must be removed by `lvgl_port_remove_disp` before deinitialization. Otherwise, there can be memory leaks!
+    /* If deinitializing LVGL port, remember to delete all displays: */
+    lvgl_port_remove_disp(disp);
+```
 
 ### Add touch input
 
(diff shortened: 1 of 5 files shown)


components/esp_lvgl_port/include/esp_lvgl_port.h
```diff
@@ -20,6 +20,10 @@
 #include "esp_lcd_touch.h"
 #endif
 
+#if __has_include ("iot_button.h")
```
igrr, line 23: Since you have `button` component in the dependencies list of esp_lvgl_port, when can this condition be false?

components/esp_lvgl_port/include/esp_lvgl_port.h
```diff
@@ -73,6 +77,18 @@ typedef struct {
 } lvgl_port_touch_cfg_t;
 #endif
 
+#if __has_include ("iot_button.h")
+/**
+ * @brief Configuration of the navigation buttons structure
+ */
+typedef struct {
+    lv_disp_t *disp;                /*!< LVGL display handle (returned from lvgl_port_add_disp) */
+    const button_config_t *button_prev;   /*!< Navigation button for preview */
```
igrr, line 86:
```suggestion
    const button_config_t *button_prev;   /*!< Navigation button for previous */
```

components/esp_lvgl_port/idf_component.yml
```diff
@@ -8,3 +8,6 @@ dependencies:
     public: true
   esp_lcd_touch:
     version: "^1.0"
+  button:
```
igrr, line 11: Could we avoid adding the dependency here, and add it in the BSP component, instead? Otherwise anyone who uses `esp_lvgl_port` will also use `button` component, even if they don't need it.

espzav: Thank you for the review! This is what I wanted, when I put `#if __has_include ("iot_button.h")` in code. But when I remove the `button` component from YML, it is not working. Linker cannot find the functions. Maybe, it can be better for selectable from Kconfig.

igrr: You can do something like this in CMakeLists.txt of esp_lvgl_port:

```cmake
    idf_build_get_property(build_components BUILD_COMPONENTS)
    if("espressif__button" IN_LIST build_components)
        target_link_libraries(${COMPONENT_LIB} PRIVATE idf::espressif__button)
    endif()
```

and then add the button component only in the BSP's idf_component.yml, not in esp_lvgl_port idf_component.yml.

espzav: Thank you very much! That's working!

espzav:
I added one fix for rotation: When the screen is swapped by default, the rotation was fine, but rotating was clockwise and it should be counterclockwise.

components/esp_lvgl_port/include/esp_lvgl_port.h
espzav: As we solved it in another thread, this should left here.

espzav:
@tore-espressif PTAL

components/esp_lvgl_port/README.md
```diff
@@ -99,6 +99,50 @@ Add touch input to the LVGL. It can be called more times for adding more touch i
 
 **Note:** The inputs added in this function are not removed in `lvgl_port_deinit`. They must be removed by `lvgl_port_remove_touch` before deinitialization. Otherwise, there can be memory leaks!
 
+### Add buttons input
+
+Add buttons input to the LVGL. It can be called more times for adding more buttons inputs for different displays.
```
igrr, line 104: Maybe now we can add a sentence here that this feature is available if `espressif/button` component is included.

components/esp_lvgl_port/README.md
```diff
@@ -99,6 +99,50 @@ Add touch input to the LVGL. It can be called more times for adding more touch i
...
+        .button_prev = &bsp_button_config[0],
+        .button_next = &bsp_button_config[1],
+        .button_enter = &bsp_button_config[2]
+    };
+
+    /* Add buttons input (for selected screen) */
+    lvgl_port_add_navigation_buttons(&btns);
+```
+
+**Note 1:** The inputs added in this function are not removed in `lvgl_port_deinit`. They must be removed by `lvgl_port_remove_navigation_buttons` before deinitialization. Otherwise, there can be memory leaks!
```
igrr, line 142: Optional suggestion: can we convert this note into a part of the code snippet above? For example:

```c++
    /* Add buttons input (for selected screen) */
    lv_indev_t* buttons_handle = lvgl_port_add_navigation_buttons(&btns);
    
    /* ... the rest of the initialization ... */

    /* If deinitializing LVGL port, remember to delete the buttons: */
    lvgl_port_remove_navigation_buttons(buttons_handle);
```

components/esp_lvgl_port/README.md
```diff
@@ -99,6 +99,50 @@ Add touch input to the LVGL. It can be called more times for adding more touch i
...
+
+**Note 2:** When you use navigation buttons for control LVGL objects, there must be used LVGL groups. More info in [LVGL documentation](https://docs.lvgl.io/master/overview/indev.html?highlight=lv_indev_get_act#keypad-and-encoder).
```
igrr, line 144:
```suggestion
**Note:** When you use navigation buttons for control LVGL objects, these objects must be added to LVGL groups. See [LVGL documentation](https://docs.lvgl.io/master/overview/indev.html?highlight=lv_indev_get_act#keypad-and-encoder) for more info.
```

components/esp_lvgl_port/README.md
espzav: Great! Thank you! Added to all parts (touch, display) instead of notes.

espzav:
@igrr @tore-espressif Note: I think, that there will be nice to add support for some encoder component (I cannot find any encoder component - add to iot_button could be nice) :-)

igrr:
There's an example written by @suda-morris: https://github.com/espressif/esp-idf/tree/master/examples/peripherals/pcnt/rotary_encoder/main, maybe it can be converted into a component.

igrr (approved):

components/esp_lvgl_port/esp_lvgl_port.c
```diff
@@ -56,6 +56,24 @@ typedef struct {
...
 #endif
 
+#if __has_include ("iot_button.h")
+
+typedef enum {
+    LVGL_PORT_NAV_BTN_PREV,
+    LVGL_PORT_NAV_BTN_NEXT,
+    LVGL_PORT_NAV_BTN_ENTER,
+    LVGL_PORT_NAV_BTN_CNT,
+} lvgl_port_nav_btns_t;
```
tore-espressif, line 66: I don't really understand this. What if I need more or less than exactly 3 buttons?

espzav: These are only navigation buttons. How can you use more than these buttons for navigation? LVGL support only these three buttons for navigation (simulate encoder using). I have an example (modified ESP-BOX examples), but it is using new esp-box-lite, which is not done.

tore-espressif:
@espzav Can we include https://github.com/espressif/esp-bsp/pull/133 in this PR too?

cc @kutu

components/esp_lvgl_port/esp_lvgl_port.c
tore-espressif: I didn't know much about LVGL's navigation. All clear now, thanks!

kutu:
I don't mind, you can delete my pull request.

PR #133, opened 2023-02-17
esp_lvgl_port: README parameter must be driver (BSP-274)

kutu:
in README, in function `notify_lvgl_flush_ready`
parameter must be driver, not the display


espzav:
Fixed in #132.

PR #134, opened 2023-02-20
esp_lvgl_port: change init config to use pdTICKS_TO_MS (BSP-275)

kutu:
Disclamer: I'm newbie in C and lvgl

with 100hz tick rate setting
timer_period_ms = currently 5ms is 0 ticks

and task_max_sleep_ms 500ms is too much

I created rtos task where I show/hide colon symbol every 1sec (I'm using `xTaskDelayUntil(&last_wakeup, pdMS_TO_TICKS(1000));`)
I spend quite some time figuring out why my clock colon symbol was not rendering at rate of 1sec
with timer_period_ms=500ms setting, colon symbol was flashing every 1050ms, and when time debt of 50ms overlap 500ms, it was rendering again

so basically delay was:
1050 - colon appear
1050 - colon disappear
1050 - appear
450 - disappear
1050 - appear
1050 - disappear

I looked at other lvgl ports, and found that
`lv_tick_inc()` should be invoked as short as possible, usually every 1ms
and
`lv_timer_handler()` should be every 5ms

with proposed settings, tick will be 10ms and timer handler will be 50ms
my colon is now rendering consistently around every 1sec

The problem lays down in `_lv_disp_refr_timer()` function
If you configure lvgl settings to show fps counter or memory monitor, then timer is not paused, and you will end up with refresh rate that you set in settings - so no visible problems here
But if you don't use those, timer gets paused, and function `lv_timer_handler()` in `lvgl_port_task()` returns 0xffffffff `task_delay_ms` until next refresh, which is shortened by `task_max_sleep_ms` to 500ms

maybe better approach would be to put this numbers into KConfig

### DIFF
--- a/components/esp_lvgl_port/include/esp_lvgl_port.h
+++ b/components/esp_lvgl_port/include/esp_lvgl_port.h
@@ -93,13 +93,13 @@ typedef struct {
  * @brief LVGL port configuration structure
  *
  */
-#define ESP_LVGL_PORT_INIT_CONFIG() \
-    {                               \
-        .task_priority = 4,       \
-        .task_stack = 4096,       \
-        .task_affinity = -1,      \
-        .task_max_sleep_ms = 500, \
-        .timer_period_ms = 5,     \
+#define ESP_LVGL_PORT_INIT_CONFIG()            \
+    {                                          \
+        .task_priority = 4,                    \
+        .task_stack = 4096,                    \
+        .task_affinity = -1,                   \
+        .task_max_sleep_ms = pdTICKS_TO_MS(5), \
+        .timer_period_ms = pdTICKS_TO_MS(1),   \
     }
 
 /**


tore-espressif:
Hello @kutu ,
the `ESP_LVGL_PORT_INIT_CONFIG()` macro is there only to provide you with convenient initialization of `esp_lvgl_port` component.
In case you need to change the configuration for your custom project, you can define your own `lvgl_port_cfg_t` and pass it to `lvgl_port_init`.

Would that help your problem?
Cheers, Tomas

kutu:
i currently do like this:
```c
lvgl_port_cfg_t lvgl_cfg = ESP_LVGL_PORT_INIT_CONFIG();
lvgl_cfg.task_max_sleep_ms = pdTICKS_TO_MS(5);
lvgl_cfg.timer_period_ms = pdTICKS_TO_MS(1);
```
i think default config should work without issues out of the box, people who will stumble upon same issue, will have to dig into to understand that they need to change that 500ms `task_max_sleep_ms`

as i said in my pr, everything works fine if you use FPS meter, but then you remove it and your display needs to update something only every 1s (not often) then problem starts

tore-espressif:
@kutu thanks for clarification!
@espzav could you please take a look when you're free?

espzav:
Hello @kutu,
thank you for this PR. I think, that this depend on situation. I think, that the task should sleep as much as possible (we plan add more sleep for decreasing power consuption). 
I disabled FPS and memory monitor. If I use for example touchscreen, there isn't problem with refreshing. When you want to any animations, you can use animations inside LVGL and then you shouldn't have this problem.  

And about using the macro `pdTICKS_TO_MS(5)` it is ok to use, but when you change the tick rate in menuconfig, then the value 5 wouldn't be 500 ms. It can be 5000 ms or 50 ms. I don't know, if this is ok.

kutu:
i choose `pdTICKS_TO_MS` to make sure that ticks delay is not 0

currently if i choose `pdMS_TO_TICKS(1)` it will be 0 ticks

i noticed issue in simple clock, when you have colon symbol flashing every 1sec, and nothing else is happening on the screen

if you try to make such example, you might notice that colon is flashing relatively every 1s, but every 10secs or so, you will notice it flashes with 500ms delay, and then all repeats.

while investigating my "colon flashing" issue, i looked into others implementation of lvgl port, and found this
https://github.com/lvgl/lv_port_esp32/blob/master/main/main.c#L155
```c
    while (1) {
        /* Delay 1 tick (assumes FreeRTOS tick is 10ms */
        vTaskDelay(pdMS_TO_TICKS(10));
```
and https://github.com/lvgl/lv_port_esp32/blob/master/main/main.c#L213
```c
lv_tick_inc(LV_TICK_PERIOD_MS);
```
which is 1ms

so, after changing default value of `task_max_sleep_ms` from 500ms to 50ms, issue was resolved

but i also understand that config `ESP_LVGL_PORT_INIT_CONFIG` is just here for quick start, and i can change it to my needs,

all i wanted is to resolve issue that i have to other people, who might not dig that far to find out why their colon is blinking not consistently

you can close this issue if this 500ms is intended and suit to most users

tore-espressif:
For time sensitive applications, you can use [esp_timer](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/esp_timer.html) which has microsecond resolutions and very low latency.

`lvgl_cfg.task_max_sleep_ms` and `lvgl_cfg.timer_period_ms` are configurable options, so we won't update defaults for now.

The whole concept of power saving and related LVGL task timing will be revisited after LVGL v9 is released, thanks!

Issue #121, opened 2023-02-02
ESP32-S3-Korvo-2: Generic LCD initialization function is not available (BSP-269)


TTauriainenQt:
Public function for generic LCD initialization is not available.

Proposed API-function prototype:
`esp_lcd_panel_handle_t bsp_lcd_init(esp_lcd_panel_io_color_trans_done_cb_t on_color_trans_done);`

Functionality would be same as in `static lv_disp_t *bsp_display_lcd_init(void)`
lines 206 - 258 + setting callback function to `esp_lcd_panel_io_spi_config_t` + returning pointer to the lcd panel handle.

https://github.com/espressif/esp-bsp/blob/ee79fd887a1a331bf9eb3935aa0e602780aa80fc/esp32_s3_korvo_2/esp32_s3_korvo_2.c#L206-L258

Alternative proposal:

Add two public API-function similar to ESP-BOX LCD BSP.

```
esp_lcd_panel_handle_t bsp_lcd_init(void);
void bsp_lcd_set_cb(esp_lcd_panel_io_color_trans_done_cb_t on_color_trans_done);
```

For reference
https://github.com/espressif/esp-box/blob/master/components/bsp/src/peripherals/bsp_lcd.c#L115
https://github.com/espressif/esp-box/blob/master/components/bsp/src/peripherals/bsp_lcd.c#L227

tore-espressif:
@TTauriainenQt do you need deinitialization function too?
That would free all the memory and HW resources?

tore-espressif:
Closed by #136

PR #136, opened 2023-02-22
bsp: Proposal of LCD API

tore-espressif:
This PR allows LCD initialization without LVGL. The API is in a separate file `bsp/lcd.h` to indicate that it is not needed in usual usacases.

I'd like to discuss few issues before I implement it for all BSPs:
1. Is it OK (at least for now) to omit the deinit functions? _EDIT:_  Seems to be OK for now
2. ~Should we extend the backlight_init function with variable LEDC channels and timers? (the user can, after all, implement his own backlight control)~ 

Any ideas welcome!

Related #121

### DIFF
--- a/esp_wrover_kit/include/bsp/display.h
+++ b/esp_wrover_kit/include/bsp/display.h
@@ -1,0 +1,42 @@
+/*
+ * SPDX-FileCopyrightText: 2023 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: CC0-1.0
+ */
+
+/**
+ * @file
+ * @brief BSP LCD
+ *
+ * This file offers API for basic LCD control.
+ * It is useful for users who want to use the LCD without the default Graphical Library LVGL.
+ *
+ * For standard LCD initialization, you can call all-in-one function bsp_display_start().
+ */
+
+#pragma once
+#include "esp_lcd_types.h"
+
+/**
+ * @brief Create new display panel
+ *
+ * For maximum flexibility, this function performs only reset and initialization of the display.
+ * You must turn on the display explicitly by calling esp_lcd_panel_disp_on_off().
+ * The display's backlight is not turned on either. You can use bsp_display_backlight_on/off(),
+ * bsp_display_brightness_set() (on supported boards) or implement your own backlight control.
+ *
+ * If you want to free resources allocated by this function, you can use esp_lcd API, ie.:
+ *
+ * \code{.c}
+ * esp_lcd_panel_del(panel);
+ * esp_lcd_panel_io_del(io);
+ * spi_bus_free(spi_num_from_configuration);
+ * \endcode
+ *
+ * @param[out] ret_panel esp_lcd panel handle
+ * @param[out] ret_io    esp_lcd IO handle
+ * @return
+ *      - ESP_OK                On success
+ *      - Else                  esp_lcd failure
+ */
+esp_err_t bsp_display_new(esp_lcd_panel_handle_t *ret_panel, esp_lcd_panel_io_handle_t *ret_io);
--- a/esp-box/CMakeLists.txt
+++ b/esp-box/CMakeLists.txt
@@ -3,5 +3,5 @@ idf_component_register(
     INCLUDE_DIRS "include"
     PRIV_INCLUDE_DIRS "priv_include"
     REQUIRES driver spiffs
-    PRIV_REQUIRES esp_timer esp_lcd esp_lcd_touch
+    PRIV_REQUIRES esp_lcd
 )
--- a/esp-box/esp-box.c
+++ b/esp-box/esp-box.c
@@ -4,18 +4,19 @@
  * SPDX-License-Identifier: CC0-1.0
  */
 
-#include "esp_timer.h"
 #include "driver/gpio.h"
 #include "driver/ledc.h"
 #include "driver/spi_master.h"
 #include "esp_err.h"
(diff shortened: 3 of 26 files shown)


esp_wrover_kit/include/bsp/display.h
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * @brief BSP display configuration structure
+ */
+typedef union {
+    struct {
+        int num;               /*!< Index of SPI peripheral to use */
+        int clk_hz;            /*!< SPI clock in Hz */
+        int max_transfer_size; /*!< SPI max transfer size in bytes */
+    } spi;
+} bsp_display_config_t;
```
espzav, line 29: Why we need this structure? The BSP knows these values. All these values should be used from BSP header file.

esp_wrover_kit/include/bsp/display.h
```diff
@@ -0,0 +1,64 @@
...
+ * \endcode
+ *
+ * @param[in]  config    Configuration of display HAL
+ * @param[out] ret_panel esp_lcd panel handle
+ * @param[out] ret_io    esp_lcd IO handle
+ * @return
+ *      - ESP_OK                On success
+ *      - Else                  esp_lcd failure
+ */
+esp_err_t bsp_display_new(const bsp_display_config_t *config, esp_lcd_panel_handle_t *ret_panel, esp_lcd_panel_io_handle_t *ret_io);
```
espzav, line 53: I think, that this function should be independent. Without `config`. 
If you can change the board with another LCD, then it will be very simple - there should be same API without any other settings.

esp_wrover_kit/include/bsp/display.h
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * SPDX-License-Identifier: CC0-1.0
+ */
+
+/**
+ * @file
+ * @brief BSP LCD
+ *
+ * This file offers API for basic LCD control.
+ * It is useful for users who want to use the LCD without a Graphical Library or without the default LVGL library.
```
espzav, line 12:
```suggestion
 * It is useful for users who want to use the LCD without the default Graphical Library LVGL.
```

espzav (commented):
@tore-espressif Thank you! Great work! Some comments left.

esp_wrover_kit/include/bsp/display.h
tore-espressif: Yes, that might be better. 

However, we should implement a way of changing these HAL parameters (SPI num, clk and max_transfer_size). Currently they are hardcoded in the header, which cannot be changed if you are using idf-component-manager.

Would it be OK to change it via menuconfig?

espzav: Yes, we should put these params into menuconfig.

tore-espressif:
@espzav Changes compared to the initial proposal:
* Backlight is now initialized (but not turned on)
* Display is reset and initialized (but not turned on)
TODOs:
- [ ] Check backlight function on Korvo 2

tore-espressif:
@TTauriainenQt PTAL!

espzav:
Hi @TTauriainenQt,
here is the example of usage:
```
static bool _flush_ready_callback(esp_lcd_panel_io_handle_t panel_io, esp_lcd_panel_io_event_data_t *edata, void *user_ctx)
{
    ....
    return false;
}

...

esp_lcd_panel_handle_t lcd_panel;
esp_lcd_panel_io_handle_t lcd_io;
err = bsp_display_new(&lcd_panel, &lcd_io);

/* Register done callback */
const esp_lcd_panel_io_callbacks_t cbs = {
    .on_color_trans_done = _flush_ready_callback,
};
esp_lcd_panel_io_register_event_callbacks(lcd_io, &cbs, NULL);

...
```

esp32_s3_lcd_ev_board/esp32_s3_lcd_ev_board.c
```diff
@@ -20,6 +20,7 @@
 #include "lvgl.h"
 
 #include "bsp/esp32_s3_lcd_ev_board.h"
+#include "bsp/display.h"
```
espzav, line 23: Here is missing implementation of the `bsp_display_new`

espzav (commented):
@tore-espressif Thank you for this PR! Great work! Thank you for fix some minor issues in other BSPs. Some notes:
1. Same as LCD should be new function for touch init. Do you plan it in next PR?
2. The EV board is missing the `bsp_display_new` function.
3. Did you test all on hardware?

Thanks!

esp32_s3_lcd_ev_board/esp32_s3_lcd_ev_board.c
tore-espressif: I'll leave the LCD_EV board for follow-up MR. I don't understand the RGB driver enough, yet.

tore-espressif:
> @tore-espressif Thank you for this PR! Great work! Thank you for fix some minor issues in other BSPs. Some notes:
> 
> 1. Same as LCD should be new function for touch init. Do you plan it in next PR?
> 2. The EV board is missing the `bsp_display_new` function.
> 3. Did you test all on hardware?
> 
> Thanks!

@espzav Thank you very much for the review!
1. Touch will be done it another PR, so this one isn't too big
2. I'll leave the EV board for follow-up MR, I don't understand the RGB driver enough, yet
3. Tested on Korvo-2, Kaluga, Wrover and USB-OTG. All work OK!

espzav (approved):

PR #139, opened 2023-03-06
bsp: Add generic touch init functions

### LINKED ISSUE #122: ESP32-S3-Korvo-2: Generic touch controller initialization function is not available (BSP-270)
Public function for generic touch controller initialization is not available.

Proposed API-function prototype:
`esp_lcd_touch_handle_t bsp_lcd_touch_init(void);
`

Functionality would be same as in `static lv_indev_t *bsp_display_indev_init(lv_disp_t *disp)` lines 286 - 306 + returning pointer to the touch handle.

https://github.com/espressif/esp-bsp/blob/ee79fd887a1a331bf9eb3935aa0e602780aa80fc/esp32_s3_korvo_2/esp32_s3_korvo_2.c#L286-L306

tore-espressif:
# TODO:

# Change description

Add API for touchscreen initialization without LVGL.

Closes #122


esp-box/idf_component.yml
```diff
@@ -1,4 +1,4 @@
-version: "2.3.0"
+version: "2.4.0"
```
espzav, line 1: It isn't breaking change. Do we need change this version. I think, that enough for this change is version 2.3.1. (same for all BSPs edited in this PR)

esp-box/include/bsp/touch.h
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+/**
+ * @file
+ * @brief BSP Touchscreen
+ *
+ * This file offers API for basic touchscreen initialization.
+ * It is useful for users who want to use the touchscreen without the default Graphical Library LVGL.
+ *
+ * For standard LCD initialization, you can call all-in-one function bsp_display_start().
```
espzav, line 14:
```suggestion
 * For standard LCD initialization with LVGL graphical library, you can call all-in-one function bsp_display_start().
```
Could you change it in `display.h` too, please?

espzav: Same for all same files...

espzav (approved):
@tore-espressif good work! Thank you. Only some minor notes, if you don't want change it, you can leave it. LGTM

esp-box/idf_component.yml
tore-espressif: Thank you for the review @espzav 

Breaking changes should be indicated by major version bump (->3.0.0).

But I agree, let's bump to 2.3.1

Issue #145, opened 2023-03-13
Display color format / bits per pixel do not have public defines (BSP-283)


TTauriainenQt:
Display color format / bits per pixel is not available from display.h or from esp-bsp.h ( via board specific header ).

In the BSP implementation these values are hardcoded.

Example from Korvo-2 BSP:
https://github.com/espressif/esp-bsp/blob/160c0d66d851c3e247de4c535dc8ddad5a7e585f/esp32_s3_korvo_2/esp32_s3_korvo_2.c#L235-L236

tore-espressif:
@TTauriainenQt yes, these values are constant and specific to the display that is shipped with korvo-2 board. Do you need to change these values?

Issue #144, opened 2023-03-10
Cannot use display.h from C++ file  (BSP-282)


TTauriainenQt:
https://github.com/espressif/esp-bsp/blob/160c0d66d851c3e247de4c535dc8ddad5a7e585f/esp32_s3_korvo_2/include/bsp/display.h#L42

Suggested fix:
```
#ifdef __cplusplus
extern "C" {
#endif
esp_err_t bsp_display_new(esp_lcd_panel_handle_t *ret_panel, esp_lcd_panel_io_handle_t *ret_io); 
#ifdef __cplusplus
}
#endif
```

tore-espressif:
@TTauriainenQt I'm sorry, I must have missed that. Thanks for the issue?

PR #147, opened 2023-03-14
bsp: Fix display headers for use in C++ and update BSP constants

### LINKED ISSUE #143: Cannot use touch.h from C++ file (BSP-281)
https://github.com/espressif/esp-bsp/blob/160c0d66d851c3e247de4c535dc8ddad5a7e585f/esp32_s3_korvo_2/include/bsp/touch.h#L34

Suggested fix:

```
#ifdef __cplusplus
extern "C" {
#endif
esp_err_t bsp_touch_new(esp_lcd_touch_handle_t *ret_touch); 
#ifdef __cplusplus
}
#endif
```

espzav:
Fixed display and touch headers for using with C++ and updated BSP header with LCD constants.

Closes #143, #144, #145

### DIFF
--- a/esp-box/include/bsp/display.h
+++ b/esp-box/include/bsp/display.h
@@ -17,6 +17,23 @@
 #pragma once
 #include "esp_lcd_types.h"
 
+/* LCD color formats */
+#define ESP_LCD_COLOR_FORMAT_RGB565    (1)
+#define ESP_LCD_COLOR_FORMAT_RGB888    (2)
+
+/* LCD display color format */
+#define BSP_LCD_COLOR_FORMAT        (ESP_LCD_COLOR_FORMAT_RGB565)
+/* LCD display color bytes endianess */
+#define BSP_LCD_BIGENDIAN           (1)
+/* LCD display color bits */
+#define BSP_LCD_BITS_PER_PIXEL      (16)
+/* LCD display color space */
+#define BSP_LCD_COLOR_SPACE         (ESP_LCD_COLOR_SPACE_BGR)
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 /**
  * @brief Create new display panel
  *
@@ -40,3 +57,7 @@
  *      - Else                  esp_lcd failure
  */
 esp_err_t bsp_display_new(esp_lcd_panel_handle_t *ret_panel, esp_lcd_panel_io_handle_t *ret_io);
+
+#ifdef __cplusplus
+}
+#endif
--- a/esp32_s3_usb_otg/include/bsp/esp32_s3_usb_otg.h
+++ b/esp32_s3_usb_otg/include/bsp/esp32_s3_usb_otg.h
@@ -126,7 +126,9 @@ esp_err_t bsp_sdcard_unmount(void);
 #define BSP_LCD_V_RES              (240)
 #define BSP_LCD_PIXEL_CLOCK_HZ     (40 * 1000 * 1000)
 #define BSP_LCD_SPI_NUM            (SPI3_HOST)
-#define BSP_LCD_BUFF_SIZE          (BSP_LCD_H_RES * 30) // Frame buffer size in pixels
+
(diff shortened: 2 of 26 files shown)


espzav:
@TTauriainenQt please, is it enough for your Qt portation? As you mention, you need to know the target format i.e. RGB565 / RGB888 / ARGB8888, we haven't got this in our BSPs. All LCDs are set to RGB565.

espzav:
Hi @TTauriainenQt, Thank you for your clarification. But if I go to ILI9341 driver implementation (https://github.com/espressif/esp-bsp/blob/master/components/lcd/esp_lcd_ili9341/esp_lcd_ili9341.c) I see, that the information `bits_per_pixel` is strictly for color format (line 73).

I think that you need the `BSP_LCD_COLOR_SPACE`, because we set it in the driver too (line 60). This is information about input color data.

It seems that we haven't got any API for setting endianness in the driver. A didn't find any info in the driver about set endianness. We are using default. This is the reason why we cannot provide this information to you from BSP.

tore-espressif:
@TTauriainenQt That is a good point, for most of our examples we solve it in LVGL that can swap the upper and lower bytes (the config is here https://github.com/espressif/esp-bsp/blob/master/examples/display/sdkconfig.defaults#L6).

Some display controllers provide an option to accept both low- and high-endian data, but there are some restrictions. Typically that it does not work for all interfaces (SPI, parallel) or for all color formats. That is why we use the default endianness and swap the bytes manually.

esp32_s3_usb_otg/include/bsp/esp32_s3_usb_otg.h
```diff
@@ -126,7 +126,11 @@ esp_err_t bsp_sdcard_unmount(void);
 #define BSP_LCD_V_RES              (240)
 #define BSP_LCD_PIXEL_CLOCK_HZ     (40 * 1000 * 1000)
 #define BSP_LCD_SPI_NUM            (SPI3_HOST)
-#define BSP_LCD_BUFF_SIZE          (BSP_LCD_H_RES * 30) // Frame buffer size in pixels
+#define BSP_LCD_BITS_PER_PIXEL     (16)
+#define BSP_LCD_COLOR_SPACE        (ESP_LCD_COLOR_SPACE_RGB)
```
tore-espressif, line 130: Could we move these definitions to `display.h`? (color space and bits per pixel)

espzav: Yes, I will do it.

espzav:
Hello @TTauriainenQt and @tore-espressif I updated the header files. Moved to `display.h` and added new constants. The `BSP_LCD_COLOR_FORMAT` taking one of these values:
```
/* LCD color formats */
#define ESP_LCD_COLOR_FORMAT_RGB565    (1)
#define ESP_LCD_COLOR_FORMAT_RGB888    (2)
```
I think, we should move these color formats definition to esp_lcd in future as a enum. 

The endian will be used in LVGL port in future for set the color swap in runtime instead of `CONFIG_LV_COLOR_16_SWAP=y` (it is not released in LVGL yet)

esp-box/include/bsp/display.h
```diff
@@ -16,6 +16,24 @@
 
 #pragma once
 #include "esp_lcd_types.h"
+#include "sdkconfig.h"
```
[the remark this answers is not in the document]
espzav: Yes, removed.

espzav:
@TTauriainenQt I didn't change the size of drawing buffer. I only move all constants to header. Now, there is `#define BSP_LCD_DRAW_BUFF_SIZE     (BSP_LCD_H_RES * 50)`  and before the `BSP_LCD_H_RES * 50` was used as a size of buffer in C file. The buffer can be higher, but the ESP32S3 have too small RAM and if you increase the buffer, it is working, but there isn't too much RAM for other things. 

@tore-espressif We should think about mechanism to set the size of buffer from custommer project? What about move it to KConfig?

PR #148, opened 2023-03-21
bsp: Allow to change size of the draw buffer

espzav:
After discussion, there should be way, how to change the draw buffer size in bsp. This is the solution.

### DIFF
--- a/esp_wrover_kit/include/bsp/display.h
+++ b/esp_wrover_kit/include/bsp/display.h
@@ -40,6 +40,14 @@ extern "C" {
 #endif
 
 /**
+ * @brief BSP display configuration structure
+ *
+ */
+typedef struct {
+    int max_transfer_sz;    /*!< Maximum transfer size, in bytes. */
+} bsp_display_config_t;
+
+/**
  * @brief Create new display panel
  *
  * For maximum flexibility, this function performs only reset and initialization of the display.
@@ -55,13 +63,14 @@ extern "C" {
  * spi_bus_free(spi_num_from_configuration);
  * \endcode
  *
+ * @param[in]  config    display configuration
  * @param[out] ret_panel esp_lcd panel handle
  * @param[out] ret_io    esp_lcd IO handle
  * @return
- *      - ESP_OK                On success
- *      - Else                  esp_lcd failure
+ *      - ESP_OK         On success
+ *      - Else           esp_lcd failure
  */
-esp_err_t bsp_display_new(esp_lcd_panel_handle_t *ret_panel, esp_lcd_panel_io_handle_t *ret_io);
+esp_err_t bsp_display_new(const bsp_display_config_t *config, esp_lcd_panel_handle_t *ret_panel, esp_lcd_panel_io_handle_t *ret_io);
 
 #ifdef __cplusplus
 }
--- a/esp-box/esp-box.c
+++ b/esp-box/esp-box.c
@@ -188,9 +188,10 @@ esp_err_t bsp_display_backlight_on(void)
     return bsp_display_brightness_set(100);
 }
 
-esp_err_t bsp_display_new(esp_lcd_panel_handle_t *ret_panel, esp_lcd_panel_io_handle_t *ret_io)
+esp_err_t bsp_display_new(const bsp_display_config_t *config, esp_lcd_panel_handle_t *ret_panel, esp_lcd_panel_io_handle_t *ret_io)
 {
     esp_err_t ret = ESP_OK;
+    assert(config != NULL && config->max_transfer_sz > 0);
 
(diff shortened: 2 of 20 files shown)


espzav:
@tore-espressif PTAL

esp-box/include/bsp/esp-box.h
```diff
@@ -274,9 +274,12 @@ esp_err_t bsp_spiffs_unmount(void);
 #define BSP_LCD_PIXEL_CLOCK_HZ     (40 * 1000 * 1000)
 #define BSP_LCD_SPI_NUM            (SPI3_HOST)
 
-#define BSP_LCD_DRAW_BUFF_SIZE     (BSP_LCD_H_RES * 10)
 #define BSP_LCD_DRAW_BUFF_DOUBLE   (0)
 
+#ifndef BSP_LCD_DRAW_BUFF_SIZE
```
igrr, line 279: Just out of curiosity, why is it not a value which is set at run time? We aren't allocating this buffer statically anyway, right? (Or, if making it a compile time setting, why not via a Kconfig option?)

espzav:
> Hello @espzav , so how is the user supposed to configure `BSP_LCD_DRAW_BUFF_SIZE`?

When you define the `BSP_LCD_DRAW_BUFF_SIZE` before the `esp-box.h` hedaer file.

espzav:
> > > Hello @espzav , so how is the user supposed to configure `BSP_LCD_DRAW_BUFF_SIZE`?
> > 
> > 
> > @tore-espressif When you define the `BSP_LCD_DRAW_BUFF_SIZE` before the `esp-box.h` hedaer file.
> 
> Ah, but this way, the definition will never make its way to `esp_box.c`, right? :D

 @tore-espressif yes, you are right. Then, you can define into Cmake :-) I absolutely agree, that this is not good solution. I am sorry. It seems, that we must change API and add it into runtime params :-(

esp-box/include/bsp/esp-box.h
tore-espressif: I am in favor of the first solution -> runtime settings.

The initial API was drafted with an idea of as simple as possible functions, often with no parameters, so the user could just call a few functions to initialize his development board. The requirement has changed now, as we want to ship BSP with various GUI projects. Let's now introduce a breaking change, that will be open for extensions and closed for modifications :)

espzav: @tore-espressif Ok, I am working on it. Thank you

espzav:
@igrr and @tore-espressif I changed it to runtime API. I made it only for one board, please, could you check it and I will copy it to all BSPs after that.

tore-espressif:
> @igrr and @tore-espressif I changed it to runtime API. I made it only for one board, please, could you check it and I will copy it to all BSPs after that.

So the LVGL users that call `bsp_display_start` will not have an option to change the draw buffer size, right?

espzav:
> > @igrr and @tore-espressif I changed it to runtime API. I made it only for one board, please, could you check it and I will copy it to all BSPs after that.
> 
> So the LVGL users that call `bsp_display_start` will not have an option to change the draw buffer size, right?

Yes, that's right. The important is allow to change the transfer size in `bsp_display_new`. Do you think, that we should add it into  `bsp_display_start`?

esp_wrover_kit/include/bsp/display.h
```diff
@@ -55,13 +63,23 @@ extern "C" {
...
+esp_err_t bsp_display_new(bsp_display_config_t *config, esp_lcd_panel_handle_t *ret_panel, esp_lcd_panel_io_handle_t *ret_io);
+
+/**
+ * @brief BSP display default configuration
+ *
  */
-esp_err_t bsp_display_new(esp_lcd_panel_handle_t *ret_panel, esp_lcd_panel_io_handle_t *ret_io);
+#define BSP_DISPLAY_CONFIG()           \
+    {                                       \
+        .max_transfer_sz = BSP_LCD_DRAW_BUFF_SIZE*sizeof(uint16_t), \
```
tore-espressif, line 81: This definition relies on `BSP_LCD_DRAW_BUFF_SIZE`, which is defined in different header file: `esp_wrover_kit.h`

tore-espressif (changes requested):
Only one comment, otherwise LGTM. Let's wait for @igrr

esp_wrover_kit/include/bsp/display.h
espzav: Right, I included header `esp-bsp.h`.

espzav:
@tore-espressif I reworked it. Now it is changed for all boards. PTAL

tore-espressif (approved):
LGTM!

PR #150, opened 2023-03-30
lcd_touch: Add interrupt callback into esp_lcd_touch

espzav:
Improved esp_lcd_touch: Added support for interrupt callback. It can be registered when initializing or later (later is for LVGL port for future use).
This feature was requested by Qt developers (BSP-284).

### DIFF
--- a/components/lcd_touch/esp_lcd_touch/esp_lcd_touch.c
+++ b/components/lcd_touch/esp_lcd_touch/esp_lcd_touch.c
@@ -7,10 +7,14 @@
 #include <stdlib.h>
 #include "freertos/FreeRTOS.h"
 #include "freertos/task.h"
+#include "driver/gpio.h"
 #include "esp_system.h"
 #include "esp_err.h"
+#include "esp_check.h"
 #include "esp_log.h"
 #include "esp_lcd_touch.h"
+
+static const char *TAG = "TP";
 
 /*******************************************************************************
 * Function definitions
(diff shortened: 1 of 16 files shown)


components/lcd_touch/esp_lcd_touch/esp_lcd_touch.c
```diff
@@ -195,3 +199,27 @@ esp_err_t esp_lcd_touch_del(esp_lcd_touch_handle_t tp)
 
     return ESP_OK;
 }
+
+esp_err_t esp_lcd_touch_register_interrupt_callback(esp_lcd_touch_handle_t tp, esp_lcd_touch_interrupt_callback_t callback)
+{
+    esp_err_t ret = ESP_OK;
+    assert(tp != NULL);
+    assert(callback != NULL);
```
tore-espressif, line 207: Could we disable the interrupt in case user passes callback==NULL?

components/lcd_touch/esp_lcd_touch/esp_lcd_touch.c
```diff
@@ -195,3 +199,27 @@ esp_err_t esp_lcd_touch_del(esp_lcd_touch_handle_t tp)
...
+    /* Interrupt pin is not selected */
+    if (tp->config.int_gpio_num == GPIO_NUM_NC) {
+        return ESP_ERR_INVALID_ARG;
+    }
+
+    tp->config.interrupt_callback = callback;
+
+    /* Add GPIO ISR handler */
+    ret = gpio_install_isr_service(0);
+    ESP_RETURN_ON_ERROR(ret, TAG, "GPIO ISR install failed");
```
tore-espressif, line 218: This is a system-wide service. User might have already installed it, so `ESP_ERR_INVALID_STATE` should not be treated as error

espzav: Good point, thank you. I fixed it.

components/lcd_touch/esp_lcd_touch/esp_lcd_touch.c
espzav: I changed it. Now, it is possible :-)

espzav:
@tore-espressif Thank you for the review. Please, could you check my latest changes? 
I tested only tt21100 and GT911, because I haven't got HW for others. 
@Lzw655 please, have you got any HW for test other touch controllers? Could you help with test and review please? (Mainly the right value for `.intr_type = GPIO_INTR_NEGEDGE,` for each touch controller. Seems, that ussually it is same as I put it, some controllers can set it.

Lzw655 (approved):
@espzav Hi, thank you for adding this useful function. Due to all my HWs don't use the INT pin, I can't actually test it. But according to their datasheets, I think this function can work well. LGTM!

espzav:
@Lzw655 Thank you for review! 
@tore-espressif can we merge it?

tore-espressif (approved):
LGTM!

PR #155, opened 2023-04-05
esp_lvgl_port: Add support for USB HID mouse/keyboard as an input dev…

espzav:
This is the implementation of the USB HID component to LVGL port (using keayboard and mouse). Before register mouse or keyboard, there must be initialized USB. 
I have prepared very simple example with mouse and keyboard with esp32_s3_usb_otg BSP. I tested it on 7" screen too.  

This PR waiting for hid component (after it will be added to component registry, it can be finished)


espzav:
Example "esp-bsp\examples\display_usb_hid" usage:
Copy `hid` and `esp_lvgl_port` components into components.
Add "hid" into CMakeLists.txt into `REQUIRES`.

tore-espressif:
USB Host HID driver released https://components.espressif.com/components/espressif/usb_host_hid

espzav:
@tore-espressif PTAL
example is here: https://github.com/espressif/esp-bsp/tree/example/usb_hid/examples/display_usb_hid

components/esp_lvgl_port/README.md
```diff
@@ -152,6 +152,43 @@ Add buttons input to the LVGL. It can be called more times for adding more butto
 
 **Note:** When you use navigation buttons for control LVGL objects, these objects must be added to LVGL groups. See [LVGL documentation](https://docs.lvgl.io/master/overview/indev.html?highlight=lv_indev_get_act#keypad-and-encoder) for more info.
 
+### Add USB HID keyboard and mouse input
+
+Add mouse and keyboard input to the LVGL. This feature is available only when the component `espressif/hid` was added into the project.
```
tore-espressif, line 157:
```suggestion
Add mouse and keyboard input to the LVGL. This feature is available only when the component `espressif/usb_host_hid` was added into the project.
```

components/esp_lvgl_port/esp_lvgl_port.c
```diff
@@ -784,6 +916,323 @@ static void lvgl_port_btn_up_handler(void *arg, void *arg2)
...
+    esp_err_t ret;
+
+    /* USB HID is already initialized */
+    if (lvgl_port_ctx.hid_ctx.task) {
+        return &lvgl_port_ctx.hid_ctx;
+    }
+
+    /* USB HID host driver config */
+    const hid_host_driver_config_t hid_host_driver_config = {
+        .create_background_task = true,
```
tore-espressif, line 931: Now we have 3 tasks to handle HID events:
1. USB Host library task (must be present in the application)
2. HID driver task
3. LVGL/HID glue task

I hope that 2 and 3 can be combined. Could you please help with this @roma-jam?

roma-jam: Hey @espzav, @tore-espressif 
Yes, there is an option to decrease the task creation and merging tasks 2 and 3 with the following steps: 
1. Use the `create_background_task = false` in config during the `hid_host_install()`
2. After this, we need to call `esp_err_t hid_host_handle_events(uint32_t timeout_ms)` from some task, I assume that could be the LVGL/HID glue task. 

Hope that can help.
If there is anything else, I am ready to help, just poke me.

components/esp_lvgl_port/esp_lvgl_port.c
```diff
@@ -784,6 +916,323 @@ static void lvgl_port_btn_up_handler(void *arg, void *arg2)
...
+    lvgl_port_usb_hid_ctx_t *ctx = (lvgl_port_usb_hid_ctx_t *)arg;
+    hid_host_device_handle_t hid_device_handle = NULL;
+    vgl_port_usb_hid_event_t msg;
+
+    assert(ctx);
+
+    ctx->running = true;
+
+    while (ctx->running) {
+        if (xQueueReceive(ctx->queue, &msg, pdMS_TO_TICKS(50))) {
```
roma-jam, line 1116: :thinking: 
Yes, it seems that this is a good place to call both of them: 

    while (ctx->running) {
        hid_host_handle_events(pdMS_TO_TICKS(50));
        if (xQueueReceive(ctx->queue, &msg, pdMS_TO_TICKS(50))) { 
            ... 
        }
    }

components/esp_lvgl_port/esp_lvgl_port.c
igrr: What do you folks think of not creating the HID task in LVGL port layer? Given the purpose of LVGL port component, it's a bit unexpected that it would handle initialization of HID. Could we either 1) let the application initialize the HID and register some callback of BSP, or 2) move this functionality into a separate optional component, esp_lvgl_port_hid, which would encapsulate all the logic related to HID bindings to LVGL.

components/esp_lvgl_port/esp_lvgl_port.c
igrr: Could we avoid waking up the system every 50ms and block on an event indefinitely? (Using queue sets, `select` or a similar approach.)
Otherwise it will be pretty hard to make this implementation compatible with power management features such as tickless idle and automatic light sleep.

components/esp_lvgl_port/esp_lvgl_port.c
```diff
@@ -784,6 +916,323 @@ static void lvgl_port_btn_up_handler(void *arg, void *arg2)
...
+            switch (msg.event) {
+            case HID_HOST_DRIVER_EVENT_CONNECTED:
+                /* Handle mouse or keyboard */
+                if (dev.proto == HID_PROTOCOL_KEYBOARD || dev.proto == HID_PROTOCOL_MOUSE) {
+                    const hid_host_device_config_t dev_config = {
+                        .callback = lvgl_port_usb_hid_host_interface_callback,
+                        .callback_arg = ctx
+                    };
+
+                    ESP_ERROR_CHECK( hid_host_device_open(hid_device_handle, &dev_config) );
```
igrr, line 1129: I seem to remember we have introduced some helper macros for error checking in BSPs. Could we use them here, to propagate the error to the user instead of aborting?

components/esp_lvgl_port/include/esp_lvgl_port.h
```diff
@@ -228,6 +246,39 @@ lv_indev_t *lvgl_port_add_navigation_buttons(const lvgl_port_nav_btns_cfg_t *but
 esp_err_t lvgl_port_remove_navigation_buttons(lv_indev_t *buttons);
 #endif
 
+
+#if __has_include ("usb/hid_host.h")
```
igrr, line 250: I see that you are following an already established pattern, just a note about it: it seems that this #if prevents Doxygen from seeing the declaration of the function, so it's not going to be visible in the API reference, once we have it.

components/esp_lvgl_port/esp_lvgl_port.c
```diff
@@ -35,11 +43,41 @@ static const char *TAG = "LVGL";
...
+        uint32_t last_key;
+        bool     pressed;
+    } kb;
+} lvgl_port_usb_hid_ctx_t;
+
+typedef struct {
+    hid_host_device_handle_t hid_device_handle;
+    hid_host_driver_event_t event;
+    void *arg;
+} vgl_port_usb_hid_event_t;
```
igrr, line 69: Typo, "vgl"

components/esp_lvgl_port/esp_lvgl_port.c
```diff
@@ -23,6 +23,14 @@
 #include "esp_lcd_touch.h"
 #endif
 
+#if __has_include ("usb/hid_host.h")
```
igrr, line 26: Instead of repeating "had_include" a bunch of times, could we define a macro such as "WITH_USB_HID" in CMake if the HID component is present, and then check that simpler macro?

components/esp_lvgl_port/esp_lvgl_port.c
espzav: lvgl_port can be used without BSP, then we cannot use BSP error checking.

igrr: In this case, let's propagate errors to the caller, instead of aborting.

components/esp_lvgl_port/esp_lvgl_port.c
espzav: @igrr When I put `hid_host_handle_events` into this task, the `xQueueReceive` cannot be blocked indefinitely.

components/esp_lvgl_port/esp_lvgl_port.c
espzav: I wanted to return errors, but here it is in task. I can add error callback and call it here, when error occur.

components/esp_lvgl_port/esp_lvgl_port.c
espzav: I added macros into header file.

components/esp_lvgl_port/esp_lvgl_port.c
espzav: @roma-jam I tried `create_background_task = false` and call `hid_host_handle_events` from  LVGL/HID glue task. After that, the function `hid_class_request_set_idle` crashes:
```
E (5897) hid-host: Control Transfer Timeout
E (5897) hid-host: hid_control_transfer(801): Unable to HALT EP
ESP_ERROR_CHECK failed: esp_err_t 0x102 (ESP_ERR_INVALID_ARG) at 0x42009830
```

components/esp_lvgl_port/esp_lvgl_port.c
espzav: I am thinking about the tasks for USB HID and we can make new component, asi @igrr said, it can be something similar like `esp_lvgl_port` for USB HID. But I think, that the best will be put tasks inside the HID component, instead of creating new component. What do you think?

roma-jam: Personally, agreed. 
I think that we do not need a separate task at all here. The usage of event_loop can solve this problem and provide the callback which can be used easily (the same way, how the `BT` and `BLE` transport levels are work in `esp_hid`). By the way, this types of optimization (and others, which helps to save some memory) are planned and in process right now. 
As well as possibility to add layouts for a keyboards, 
 
So, the other question, is the LVGL component can be used not only USB HID, but with the bluetooth HID devices also?  
With usage of `esp_hid` component?

components/esp_lvgl_port/esp_lvgl_port.c
roma-jam: @espzav Thanks for the report. Working on it.

components/esp_lvgl_port/esp_lvgl_port.c
espzav: The bluetooth HID devices are not supported now in LVGL port. We can add it in future. 
Now, I propose leave the tasks in LVGL port like it is and change it in future, when USB HID component will be improved for better handling tasks. 
@igrr @tore-espressif What do you think?

tore-espressif: @espzav @roma-jam There are many new questions that rose from this usecase. Many of them require breaking changes and will take some time.

I'm OK with releasing it as-is for now, so we can demonstrate the feature to our users and optimize in follow-up MRs

espzav: @roma-jam Please, when is the update of the USB HID planned?

espzav:
@tore-espressif I updated and tested with example. You can test it with example from PR #250 
PTAL

tore-espressif (approved):
Thank you for the updates!

Let's merge it as it is for now.
Follow-up: Once we update HID driver with cleaner interface, we will also update this component, so the `esp_lvgl_port` does not install the HID driver

PR #156, opened 2023-04-06
lcd: add st7796 lcd driver

suda-morris:
Added ST7796 LCD driver, requested by one customer.


espzav:
@suda-morris please, is there any reason, why is this LCD driver not supported with IDF4.4? All another LCD controllers are supported for IDF4.4 and newer.
When we leave support only for >=IDF5.0, then must be modified `test_app/CMakeLists.txt` and exclude this driver from tests for IDF4.4.

components/lcd/esp_lcd_st7796/esp_lcd_st7796.c
```diff
@@ -0,0 +1,296 @@
...
+    st7796->y_gap = y_gap;
+    return ESP_OK;
+}
+
+static esp_err_t panel_st7796_disp_on_off(esp_lcd_panel_t *panel, bool on_off)
+{
+    st7796_panel_t *st7796 = __containerof(panel, st7796_panel_t, base);
+    esp_lcd_panel_io_handle_t io = st7796->io;
+    int command = 0;
+
```
espzav, line 288: If we change it to use with IDF4.4, here must be this:
```
    on_off = !on_off;
#endif
```

components/lcd/esp_lcd_st7796/esp_lcd_st7796.c
```diff
@@ -0,0 +1,296 @@
...
+    st7796->reset_level = panel_dev_config->flags.reset_active_high;
+    st7796->base.del = panel_st7796_del;
+    st7796->base.reset = panel_st7796_reset;
+    st7796->base.init = panel_st7796_init;
+    st7796->base.draw_bitmap = panel_st7796_draw_bitmap;
+    st7796->base.invert_color = panel_st7796_invert_color;
+    st7796->base.set_gap = panel_st7796_set_gap;
+    st7796->base.mirror = panel_st7796_mirror;
+    st7796->base.swap_xy = panel_st7796_swap_xy;
+    st7796->base.disp_on_off = panel_st7796_disp_on_off;
```
espzav, line 100: If we change it to use with IDF4.4, here must be this:
```
    st7796->base.disp_off = panel_ili9341_disp_on_off;
#else
    st7796->base.disp_on_off = panel_ili9341_disp_on_off;
#endif
```

components/lcd/esp_lcd_st7796/idf_component.yml
```diff
@@ -0,0 +1,5 @@
+version: "1.0.0"
+description: ESP LCD ST7796 driver
+url: https://github.com/espressif/esp-bsp/tree/master/components/lcd/esp_lcd_st7796
+dependencies:
+  idf: ">=5.0"
```
espzav, line 5: I think, that there is not reason to change to use it only with IDF5.0 and newer. I think, that it can be used in IDF4.4 too. When you change it, please, could you test it with IDF4.4?

components/lcd/esp_lcd_st7796/esp_lcd_st7796.c
```diff
@@ -0,0 +1,296 @@
...
+    if (invert_color_data) {
+        command = LCD_CMD_INVON;
+    } else {
+        command = LCD_CMD_INVOFF;
+    }
+    esp_lcd_panel_io_tx_param(io, command, NULL, 0);
+    return ESP_OK;
+}
+
+static esp_err_t panel_st7796_mirror(esp_lcd_panel_t *panel, bool mirror_x, bool mirror_y)
```
espzav, line 240: Please, did you test this driver with LVGL rotation? For example this example: `esp-bsp\examples\display_rotation`

espzav (changes requested):
@suda-morris Thank you for add this LCD driver! I left some comments and questions.

components/lcd/esp_lcd_st7796/esp_lcd_st7796.c
suda-morris: These commands are part of standard DCS (Display Command Set), and yes, the rotation can work as expect.

suda-morris:
> @suda-morris Thank you for add this LCD driver! I left some comments and questions.

Thank you @espzav for the quick review!

espzav (approved):
@suda-morris Thank you for changes! LGTM

tore-espressif (approved):

Issue #160, opened 2023-04-17
ESP-BSP allows API-breaking changes for its sub-components (BSP-301)


TT-Qt:
Selecting a specific ESP-BSP version wont protect against API-breaking changes as sub-components are allowed to fetch such changes.

https://github.com/espressif/esp-bsp/blob/bd0efea277b965123ec8252de69d815b21492d4e/esp32_s3_korvo_2/idf_component.yml#L11

tore-espressif:
Hello @TTauriainenQt , we use semantic versioning (major.minor.bugfix) for BSP components, so breaking changes should be introduced only between major versions.

The `^1` syntax allows automatic updates to new versions in this range <1.0.0; 2.0.0), so there should be no breaking change injected into your project. Have you encoutered a situtation where a breaking change was introduced unintentiaonally? We can yank the version and provide a fix.

Anyway, for production code, we recommend fixed versions of _all_ components. For example, although the latest supported version of `esp_lcd_touch_tt21100` by the Korvo2 BSP is 1.0.7 (at the time of writing), you can fix it by adding `espressif/esp_lcd_touch_tt21100: "1.0.5"` to you main `idf_component.yml` file

tore-espressif:
@espzav Seems like your https://github.com/espressif/esp-bsp/pull/150 did not update minimum required version of `esp_lcd_touch` for touch drivers that use the newly introduced API `esp_lcd_touch_register_interrupt_callback`.

Could you please release new versions of affected components? We will then remove the components with invalid dependencies list from IDF component registry, so issues like this one don't repeat.

espzav:
@tore-espressif and @TTauriainenQt I apologize for this issue. Here is the PR with fix: #161

tore-espressif:
#161 is merged and faulty versions removed from component registry.

The issue should be fixed now @TTauriainenQt

PR #161, opened 2023-04-19
esp_lcd_touch: Fixed minimum required esp_lcd_touch version.

espzav:

- [x] CI passing

# Change description
Fix for issue #160

### DIFF
--- a/components/lcd_touch/esp_lcd_touch_cst816s/idf_component.yml
+++ b/components/lcd_touch/esp_lcd_touch_cst816s/idf_component.yml
@@ -1,8 +1,8 @@
-version: "1.0.2"
+version: "1.0.2~1"
 description: ESP LCD Touch CST816S - touch controller CST816S
 url: https://github.com/espressif/esp-bsp/tree/master/components/lcd_touch/esp_lcd_touch_cst816s
 dependencies:
   idf: ">=4.4.2"
   esp_lcd_touch:
-    version: "^1.0"
+    version: "^1.0.4"
     public: true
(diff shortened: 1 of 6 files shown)


components/lcd_touch/esp_lcd_touch_cst816s/idf_component.yml
```diff
@@ -1,8 +1,8 @@
-version: "1.0.2"
+version: "1.0.3"
 description: ESP LCD Touch CST816S - touch controller CST816S
 url: https://github.com/espressif/esp-bsp/tree/master/components/lcd_touch/esp_lcd_touch_cst816s
 dependencies:
   idf: ">=4.4.2"
   esp_lcd_touch:
-    version: "^1.0"
+    version: ">=1.0.4"
```
tore-espressif, line 7: Would `^1.0.4` be safer? In case we introduce a breaking change in esp_lcd_touch v2?

components/lcd_touch/esp_lcd_touch_cst816s/idf_component.yml
```diff
@@ -1,8 +1,8 @@
-version: "1.0.2"
+version: "1.0.3"
```
tore-espressif, line 1: We could bump to `1.0.2~1` to indicate that no change in the code was made

components/lcd_touch/esp_lcd_touch_cst816s/idf_component.yml
espzav: Yes, you are right! Fixed.

PR #153, opened 2023-04-04
Examples/optimization

tore-espressif:
# ESP-BSP Pull Request checklist

- [x] CI passing

# Change description

1. Fix format warnings and remove `-Wno-format` compile flag
2. Remove few examples

We want to remove few rudimentary examples to offload the maintainers, so they can focus in improving existing examples

Closes #50

PR #163, opened 2023-04-21
fix: Leftover from https://github.com/espressif/esp-bsp/pull/153

tore-espressif:
The removed examples from https://github.com/espressif/esp-bsp/pull/153 were not removed from readmes and idf_component.yml files.

Let's wait until https://github.com/espressif/upload-components-ci-action/issues/10 is resolved before merging this


espzav (approved):
LGTM

PR #165, opened 2023-04-25
bsp: Add ESP-BOX Lite and Korvo-2 into SquareLine generator.

espzav:
Add missing boards into SquareLine generator.


tore-espressif (approved):
Thank you for the update @espzav

PR #169, opened 2023-04-26
bsp: Fixed LCD SPI maximum transfer size.

espzav:
After changes in `esp_lcd`, there is no check on maximum transfer size in SPI. It must be changed in BSP. Here is a fix.

### DIFF
--- a/esp_wrover_kit/esp_wrover_kit.c
+++ b/esp_wrover_kit/esp_wrover_kit.c
@@ -21,6 +21,8 @@
 
 static const char *TAG = "Wrover";
 
+#define LCD_SPI_MAX_DATA_SIZE (32768)
+
 sdmmc_card_t *bsp_sdcard = NULL;    // Global uSD card handler
 
 esp_err_t bsp_leds_init(void)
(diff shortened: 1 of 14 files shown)


tore-espressif:
@espzav Could you please add some background information and link to the commit in esp-idf that requires this change?
It will be very useful when we need to revisit this change in the future :)

espzav:
> @espzav Could you please add some background information and link to the commit in esp-idf that requires this change? It will be very useful when we need to revisit this change in the future :)

We have internal ticket in Jira: BSP-316, there is internal link to PR and discussion about this issue. Is it enough or do we need more here too?

esp_wrover_kit/esp_wrover_kit.c
```diff
@@ -202,7 +205,7 @@ static lv_disp_t *bsp_display_lcd_init(void)
     esp_lcd_panel_io_handle_t io_handle = NULL;
     esp_lcd_panel_handle_t panel_handle = NULL;
     const bsp_display_config_t bsp_disp_cfg = {
-        .max_transfer_sz = BSP_LCD_DRAW_BUFF_SIZE * sizeof(uint16_t),
+        .max_transfer_sz = 10 * BSP_LCD_H_RES,
```
suda-morris, line 208: not sure how much performance we will improve if we increase this size. Should we have a small test on it and choose a prpoer one?  10 lines seems is the recommendation for LVGL draw buffer

tore-espressif:
@suda-morris @espzav @igrr I'm wondering whether this change is necessary, when we plan to release another change to all supported IDF change that should guarantee no behavior change

espzav:
It was fixed in IDF.

Issue #176, opened 2023-06-06
How to exclude esp_lcd_touch when depending on esp_lvgl_port?


huhaifan:
I noticed that esp_lvgl_port has dynamic include esp_lcd_touch
```
#if __has_include ("esp_lcd_touch.h")
#include "esp_lcd_touch.h"
#endif
```
But esp_lcd_touch dependency is forced included in idf_component.yml
```
dependencies:
  esp_lcd_touch:
    version: ^1.0
```

PR #177, opened 2023-06-06
esp_lvgl_port: Removed touch dependency from LVGL port.

espzav:
Removed touch dependency from LVGL port. It is only loaded, when touch is necessary from the project. 
Fix issue #176


tore-espressif:
I got following error when I tried to build `display_audio_photo` example
```
C:/github_repos/esp-bsp/esp-box/esp-box.c:433: undefined reference to `lvgl_port_add_touch'
```

we will probably have to define it as 'download only' dependency
```
dependencies:
  esp_lcd_touch:
    version: ^1.0
    require: no
```

And then define the dependency in CMakeLists.txt for BSPs that actually need touch

tore-espressif:
More information about the `require` field is here https://docs.espressif.com/projects/idf-component-manager/en/latest/reference/manifest_file.html#common-fields-for-all-dependency-types

 EDIT: But is seems that our trick with `#if __has_include ("esp_lcd_touch.h")` won't work correctly ...

espzav:
@tore-espressif please, how did you tested? I tried copy the component from `managed_components` to `components` folder. And it was working for ESP-BOX and for ESP-BOX -Lite wasn't downloaded component `esp_lcd_touch`. Compilation was ok for both (not tested on ESP-BOX HW, I haven't got it, my ESP-BOX is now on the cupboard for Maker faire). 

For iot_button, we have this in CMakeLists.txt, maybe it should be there for touch too:
```
if("espressif__button" IN_LIST build_components)
    target_link_libraries(${COMPONENT_LIB} PRIVATE idf::espressif__button)
endif()
```

tore-espressif:
I update esp-box idf-component.yml like this and build the example
```
  esp_lvgl_port:
    version: ^1.1.1
    path: C:\\github_repos\\esp-bsp\\components\\esp_lvgl_port
    public: true
```

but maybe I'm doing something wrong :P

espzav:
@tore-espressif Thank you. I tried yours solution and I have same result as you had. I added condition into CMakeLists.txt and it seems, that it is working.

components/esp_lvgl_port/CMakeLists.txt
```diff
@@ -3,4 +3,7 @@ idf_component_register(SRCS "esp_lvgl_port.c" INCLUDE_DIRS "include" REQUIRES "e
 idf_build_get_property(build_components BUILD_COMPONENTS)
 if("espressif__button" IN_LIST build_components)
     target_link_libraries(${COMPONENT_LIB} PRIVATE idf::espressif__button)
+endif()
+if("espressif__esp_lcd_touch" IN_LIST build_components)
```
tore-espressif, line 7: One last point: the touch component will be named `espressif__esp_lcd_touch` in case it is coming from IDF component manager. But it will be name `esp_lcd_touch` when using it from local path (ie for debugging purposes).

Could you please check for both names in this condition? and do the same update for `espressif__button`?

espzav: Done.

PR #180, opened 2023-06-12
add new lcd panel IO - 3-wire SPI

Lzw655:

- [ ] CI passing

# Change description
Most RGB LCD panels use [3-wire SPI ](https://focuslcds.com/3-wire-spi-parallel-rgb-interface-fan4213/)(no DC line) to initialize. But ESP's SPI interface doesn't support transmitting 9-bit data directly (see [issue](https://github.com/espressif/esp-idf/issues/8487)). 

Due to the less stringent timing requirements, this can be achieved through software simulation (with GPIO or IO expander). And based on the new LCD panel IO, we can upload some RGB drivers like `st7701` and `gc9a01`.

### DIFF
--- a/.github/workflows/upload_component.yml
+++ b/.github/workflows/upload_component.yml
@@ -18,7 +18,7 @@ jobs:
             esp32_azure_iot_kit;esp32_s2_kaluga_kit;esp_wrover_kit;esp-box;esp32_s3_usb_otg;esp32_s3_eye;esp32_s3_lcd_ev_board;esp32_s3_korvo_2;esp-box-lite;
             components/bh1750;components/ds18b20;components/es8311;components/es7210;components/fbm320;components/hts221;components/mag3110;components/mpu6050;components/ssd1306;components/esp_lvgl_port;
             components/lcd_touch/esp_lcd_touch;components/lcd_touch/esp_lcd_touch_ft5x06;components/lcd_touch/esp_lcd_touch_gt911;components/lcd_touch/esp_lcd_touch_tt21100;components/lcd_touch/esp_lcd_touch_gt1151;components/lcd_touch/esp_lcd_touch_cst816s;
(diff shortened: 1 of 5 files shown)


Lzw655:
@tore-espressif @espzav Hi, I create a new LCD panel IO which is used to initialize RGB LCD. PTAL, thank you very much!
And I'm going to add some drivers of RGB LCD and replace the code in BSP esp32-s3-lcd-ev-board after all.

espzav:
@Lzw655 Thank you for this component. I think, that better will be update `esp_lcd` component in IDF for support 9-bit SPI communication rather than new component. 

The LCD driver for `gc9a01` is now available with esp_lcd: https://github.com/espressif/esp-bsp/tree/master/components/lcd/esp_lcd_gc9a01

What do you think: @suda-morris ?

espzav:
Some notes:
1. The `esp_lcd` IO now support 3-wire SPI mode - flag `sio_mode`
2. I think, that `esp_lcd` IO support without D/C line too
3. 9-bit data is not supported now (it will be nice to add direct into `esp_lcd`)
4. IO expander is not supported now (I don't know, if this will be ok to add direct to `esp_lcd` - thinking...)

Lzw655:
> @Lzw655 Thank you for this component. I think, that better will be update `esp_lcd` component in IDF for support 9-bit SPI communication rather than new component.
> 
> The LCD driver for `gc9a01` is now available with esp_lcd: https://github.com/espressif/esp-bsp/tree/master/components/lcd/esp_lcd_gc9a01
> 
> What do you think: @suda-morris ?

Sorry, `gc9a01` is a mistake. I mean `gc9503`...

suda-morris:
Hi, since this component is not related to any "Device", so the esp-bsp repo is not the best place. We can move it to the iot-solution repo. This makes sense because it's a solution to abstract the usage of GPIO (including IO expanded ones) bit-bang SPI interface.

PR #185, opened 2023-06-28
lcd: add gc9503 lcd driver

Lzw655:

- [ ] CI passing

# Change description
Add a new RGB LCD. It is used by ESP32-S3-LCD-Ev-Board BSP.


Lzw655:
@tore-espressif @espzav PTAL, thanks!

components/lcd/esp_lcd_gc9503/idf_component.yml
```diff
@@ -0,0 +1,11 @@
+version: "1.0.0"
+description: ESP LCD GC9503
+url: https://github.com/espressif/esp-bsp/tree/master/components/lcd/esp_lcd_gc9503
+dependencies:
+  idf: ">=4.4"
+  esp_io_expander:
+    version: "^1"
+    public: true
```
tore-espressif, line 8: I don't see where this compont is used, or why it is needed ?

components/lcd/esp_lcd_gc9503/include/esp_lcd_gc9503.h
```diff
@@ -0,0 +1,79 @@
...
+ * @file
+ * @brief ESP LCD: GC9503
+ */
+
+#pragma once
+
+#include "esp_lcd_types.h"
+#include "esp_lcd_panel_rgb.h"
+
+#include "esp_lcd_panel_io_additions.h"
```
tore-espressif, line 16: Could we move this include to the .c file and make esp_lcd_panel_io_additions a private dependency?

tore-espressif (changes requested):
Thank you very much for the PR @Lzw655 ! I left few comments

Could you please also add this new driver to this table? https://github.com/espressif/esp-bsp/blob/master/LCD.md

components/lcd/esp_lcd_gc9503/idf_component.yml
Lzw655: Sorry, it's a mistake. I also remove the `esp_lcd_panel_io_additions` which is unnecessary.

components/lcd/esp_lcd_gc9503/include/esp_lcd_gc9503.h
Lzw655: Yeah, I directly remove it.

tore-espressif (approved):
LGTM!

There is a bug in CI that is being fixed by idf_tools team, let's wait until it is resolved

Lzw655:
@tore-espressif Thank you for reviewing! I modify someplace according to your comments. But there is a CI error below. Can you help me to check it?
```
2023-06-28 08:31:36  WARNING >>> /opt/esp/idf/tools/check_python_dependencies.py:12: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
2023-06-28 08:31:36    ERROR >>> Build succeeded with warnings
```

Lzw655:
Another problem: This component just supports esp32s3. How can I avoid some checks for other chips?

components/lcd/esp_lcd_gc9503/include/esp_lcd_gc9503.h
```diff
@@ -0,0 +1,77 @@
...
+ * @note  This function first initialize GC9503 with vendor specific initialization, then calls `esp_lcd_new_rgb_panel()` to create a RGB LCD panel.
+ * @note  Vendor specific initialization can be different between manufacturers, should consult the LCD supplier for initialization sequence code.
+ *
+ * @param[in]  io_handle  LCD panel IO handle
+ * @param[in]  rgb_config Pointer to RGB panel timing configuration structure
+ * @param[out] ret_panel  Returned LCD panel handle
+ * @return
+ *      - ESP_OK: Success, otherwise returns ESP_ERR_xxx
+ */
+esp_err_t esp_lcd_new_panel_gc9503(esp_lcd_panel_io_handle_t io_handle, const esp_lcd_rgb_panel_config_t *rgb_config, esp_lcd_panel_handle_t *ret_panel);
```
tore-espressif, line 32: One more comment: this API is different to what usual esp_lcd driver looks like: 
``` c
esp_lcd_new_panel_xxx(const esp_lcd_panel_io_handle_t io, const esp_lcd_panel_dev_config_t *panel_dev_config, esp_lcd_panel_handle_t *ret_panel);
```
Your API depends on RGB implementation, while the esp_lcd driver should be IO agnostic. This has few drawbacks:
* Can be compiled only on ESP32-S3
* We can't use usual esp_lcd API like `esp_lcd_panel_reset()` ...

Lzw655: Here are explanations for your concerns:
* The GC9503 only has RGB and MIPI (which is not supported) interface. So it should be compiled only on ESP32-S3.
* Actually, we can use the most esp_lcd APIs. Because the `ret_panel`  is created by `esp_lcd_new_rgb_panel()` which is in [RGB driver](https://github.com/espressif/esp-idf/blob/67552c31dac8cd94fb0d63192a538f4f984c5b6e/components/esp_lcd/src/esp_lcd_panel_rgb.c#L69). And the driver has implemented many esp_lcd APIs.

tore-espressif: Thank you for the explanation! I'll find a way how to build this for S3 only in CI

tore-espressif:
> @tore-espressif Thank you for reviewing! I modify someplace according to your comments. But there is a CI error below. Can you help me to check it?
> 
> ```
> 2023-06-28 08:31:36  WARNING >>> /opt/esp/idf/tools/check_python_dependencies.py:12: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
> 2023-06-28 08:31:36    ERROR >>> Build succeeded with warnings
> ```

You can rebase on top of this https://github.com/espressif/esp-bsp/pull/188 to fix the problem

components/lcd/esp_lcd_gc9503/include/esp_lcd_gc9503.h
tore-espressif: in test_app/CMakeLists.txt you must exclude this component from build. Something like this: (not tested)
```
if(NOT "${IDF_TARGET}" STREQUAL "esp32s3")
    set(EXCLUDE_COMPONENTS "esp_lcd_gc9503")
endif()
```

PR #186, opened 2023-06-28
framebuffer: esp-box/esp-box-lite use framebuf height form Kconfig (BSP-331)

espressif2022:

- [ ] CI passing

# Change description
In some scenarios, the remaining SRAM is insufficient, make the num of framebuf height to configurable.

### DIFF
--- a/esp-box/Kconfig
+++ b/esp-box/Kconfig
@@ -60,6 +60,20 @@ menu "Board Support Package"
         help
             LEDC channel is used to generate PWM signal that controls display brightness.
             Set LEDC index that should be used.
+
+        config BSP_LCD_DRAW_BUF_HEIGHT
+        int "LCD framebuf height"
+        default 100
+        range 10 240
+        help
+            Framebuf is used for lvgl rendering output.
+
+        config BSP_LCD_DRAW_BUF_DOUBLE
+        int "LCD framebuf number"
+        default 0
+        range 0 1
+        help
+            Whether to enable double framebuf.
     endmenu
     
     config BSP_I2S_NUM
--- a/esp-box/Kconfig
+++ b/esp-box/Kconfig
@@ -69,9 +69,8 @@ menu "Board Support Package"
             Framebuf is used for lvgl rendering output.
 
         config BSP_LCD_DRAW_BUF_DOUBLE
-        int "LCD framebuf number"
-        default 0
-        range 0 1
+        bool "LCD double framebuf"
+        default n
         help
             Whether to enable double framebuf.
     endmenu
--- a/esp-box-lite/Kconfig
+++ b/esp-box-lite/Kconfig
@@ -69,9 +69,8 @@ menu "Board Support Package"
             Framebuf is used for lvgl rendering output.
 
         config BSP_LCD_DRAW_BUF_DOUBLE
-        int "LCD framebuf number"
-        default 0
-        range 0 1
+        bool "LCD double framebuf"
+        default n
         help
             Whether to enable double framebuf.
     endmenu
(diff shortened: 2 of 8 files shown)


tore-espressif:
Hello @espressif2022 , you are right, the frame buffer is not easily configurable now.

We were thinking about making it configurable in runtime (as opposed to you suggestion with Kconfig).
We wanted to introduce new API, that would extend the current `bsp_display_start()` like this:
``` c
lv_disp_t *bsp_display_start_with_config(lvgl_port_cfg_t *cfg);
```

This would allow users to configure their `esp_lvgl_port` and framebuffer settings on runtim, what do you think?

cc @espzav

espressif2022:
> Hello @espressif2022 , you are right, the frame buffer is not easily configurable now.
> 
> We were thinking about making it configurable in runtime (as opposed to you suggestion with Kconfig). We wanted to introduce new API, that would extend the current `bsp_display_start()` like this:
> 
> ```c
> lv_disp_t *bsp_display_start_with_config(lvgl_port_cfg_t *cfg);
> ```
> 
> This would allow users to configure their `esp_lvgl_port` and framebuffer settings on runtim, what do you think?
> 
> cc @espzav
@tore-espressif ，good suggestion.
This is needed for spi screens, probably useless for RGB screens. 
This modification is only for the screen of SPI interface, will it affect the consistency of the interface(`bsp_display_start()`)?

tore-espressif:
> This is needed for spi screens, probably useless for RGB screens.
This modification is only for the screen of SPI interface, will it affect the consistency of the interface(bsp_display_start())?

Yep, RGB screens are not yet supported by `esp_lvgl_port` , it is planned in following months.

So, unless this PR is not in a hurry, I'd suggest waiting a little. Then we can come up with API that would allow configuration of framebuffers for all LCDs (SPI, RGB...)

espressif2022:
> > This is needed for spi screens, probably useless for RGB screens.
> > This modification is only for the screen of SPI interface, will it affect the consistency of the interface(bsp_display_start())?
> 
> Yep, RGB screens are not yet supported by `esp_lvgl_port` , it is planned in following months.
> 
> So, unless this PR is not in a hurry, I'd suggest waiting a little. Then we can come up with API that would allow configuration of framebuffers for all LCDs (SPI, RGB...)

You are thoughtful，but there's a little urgency now.
Recently, many developers urged us to update esp-box-lite, and now the examples of esp-box-lite have been componentized and are waiting to be released.
At present, on some examples, the RAM is insufficient, and rainmaker fails to run, so submit this PR.

esp-box/Kconfig
```diff
@@ -60,6 +60,20 @@ menu "Board Support Package"
...
+
+        config BSP_LCD_DRAW_BUF_HEIGHT
+        int "LCD framebuf height"
+        default 100
+        range 10 240
+        help
+            Framebuf is used for lvgl rendering output.
+
+        config BSP_LCD_DRAW_BUF_DOUBLE
+        int "LCD framebuf number"
```
tore-espressif, line 72: Please change to bool in both Kconfig files

tore-espressif (approved):
> At present, on some examples, the RAM is insufficient, and rainmaker fails to run, so submit this PR.

Understood. I left one commit about using bool instead of int in Kconfig. Apart from that LGTM!

tore-espressif:
@espressif2022 I rebased your branch on top of https://github.com/espressif/esp-bsp/pull/188, which should fix the CI

espressif2022:
> @espressif2022 I rebased your branch on top of #188, which should fix the CI

ok.

PR #187, opened 2023-06-29
esp_lvgl_port: add encoder support (BSP-332)

espressif2022:

- [ ] CI passing

# Change description
add encoder event


espressif2022:
Hello, @espzav 
 The component of `esp_lv_port` is split into an independent PR here.

espzav:
@espressif2022 please, could you rebase your branch to top - there is fixed CI

espzav (approved):
@espressif2022 Thank you very much for this change! LGTM

components/esp_lvgl_port/include/esp_lvgl_port.h
```diff
@@ -77,7 +80,13 @@ typedef struct {
...
 #endif
 
-#if __has_include ("iot_button.h")
+#if __has_include ("iot_knob.h")
+typedef struct {
+    lv_disp_t *disp;    /*!< LVGL display handle (returned from lvgl_port_add_disp) */
+    const knob_config_t *encoder_a_b;
+    const button_config_t *encoder_enter;  /*!< Navigation button for enter */
+} lvgl_port_encoder_cfg_t;
+#elif __has_include ("iot_button.h")
```
tore-espressif, line 89: Why is #elif here, instead of #if?

Let's say, my hardware has one rotary encoder (so I use iot_knob) but it also has few HW buttons (iot_button). But if #elif, these two features are mutualy exclusive, right?

espressif2022: > these two features are mutualy exclusive

Because the cfg here is used for `lv_indev_drv_register`, and `lvgl` does not support multiple indrv_drv for the time being, so I think this is mutually exclusive here.

espzav: Are you sure, that LVGL do not support multiple input devices of one type? Because there can be multiple input devices (e.g. touch + encoder).

espressif2022: > Are you sure, that LVGL do not support multiple input devices of one type? Because there can be multiple input devices (e.g. touch + encoder).



> Are you sure, that LVGL do not support multiple input devices of one type? Because there can be multiple input devices (e.g. touch + encoder).

I made a mistake here,  _lv_indev_ll is a link list, you can register multiple inputs here.
So both sides are handled independently ?

espzav: Yes, I dont't know, how is it, when you use twice Encoder (buttons + encoder), but I think, that it should work.

espressif2022: > Yes, I dont't know, how is it, when you use twice Encoder (buttons + encoder), but I think, that it should work.

ok, I've pused a new commit. I test it on board `esp32-c3-board` and `esp-box-lite`,  no conflict between them.

espzav: Thank you very much! I'll test it and merge it after that.

PR #188, opened 2023-06-29
fix(ci): Ignore pkg_resources warning

tore-espressif:
Ignore pkg_resources warning coming from updated `pip`


espzav (approved):
@tore-espressif thank you! LGTM

PR #190, opened 2023-07-14
add lvgl task size and task priority config for lvgl task (BSP-334)

Jerry-ESP:

In some application, the default 4096 lvgl task stack size is not enough, so we need config this task stack size

# Change
Add a menuconfig entry to config the lvgl task stack size and priority

### DIFF
--- a/components/esp_lvgl_port/Kconfig
+++ b/components/esp_lvgl_port/Kconfig
@@ -1,0 +1,14 @@
+menu "ESP Lvgl Port"
+
+    config LVGL_PORT_TASK_STACK_SIZE
+        int "Lvgl task stack size"
+        default 4096
+        help
+            Set the default lvgl port task stack size.
+
+    config LVGL_PORT_TASK_PRIORITY
+        int "LVGL task priority"
+        default 4
+        help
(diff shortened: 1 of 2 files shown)


espzav (changes requested):
I think, that this is bad. Because user doesn't need use init macro and when he calls init with own structure, then KConfig changes will be confusing.
We want to add new function in BSP API, which will have param with  configuration structure for display initialization, where will be all params (whole structure) for LVGL port. We have it in plan.

espzav:
Here is PR for ESP-BOX BSP, which adding function for initialize display with params: https://github.com/espressif/esp-bsp/pull/192
Is it ok for you?

espressif2022:
> Here is PR for ESP-BOX BSP, which adding function for initialize display with params: #192 Is it ok for you?

@Jerry-ESP ，I think it's ok. there add an independent interface.

Jerry-ESP:
> Here is PR for ESP-BOX BSP, which adding function for initialize display with params: #192 Is it ok for you?

Yes, it is OK, thanks a lot, wait for your PR merged.

PR #195, opened 2023-07-20
Add pre-commit for copyright

espzav:
# Change description
Added pre-commit hook for copyright/license check


.pre-commit-config.yaml
```diff
@@ -17,3 +17,9 @@ repos:
         types_or: [c, c++]
         args: ['--fix=lf']
         description: Forces to replace line ending by the UNIX 'lf' character
+
+- repo: https://github.com/espressif/check-copyright/
+  rev: v1.0.2
```
igrr, line 22: You can use the latest 1.0.3 here

espzav:
@igrr @leeebo PTAL

.github/workflows/pre-commit.yml
```diff
@@ -8,6 +8,10 @@ jobs:
   pre-commit:
     runs-on: ubuntu-latest
     steps:
+    - name: Install copyright check
+      shell: bash
+      run: |
+        pip install git+https://github.com/espressif/check-copyright.git@master
```
igrr, line 14: Could you try removing these 4 lines? Since the copyright check is part of the pre-commit, the package will be installed by pre-commit framework automatically into its own virtual environment.

igrr (approved):
One minor comment about the CI part, LGTM otherwise.

leeebo (approved):
LGTM!

.github/workflows/pre-commit.yml
espzav: I added this, because the CI failed without it (it was only in pre-commit)

igrr: Probably you were installing check-copyright==1.0.2? It should work with 1.0.3 without these 4 lines.

espzav:
@igrr Thank you very much for review and for all helps!

PR #196, opened 2023-07-26
Kaluga v3

tore-espressif:
Kaluga v3:
* Introduce backward compatibility with IDF v4.4.5
* Audio uses esp_codec_dev
* Added button initializer function
* Added native support for SPIFFS

There are other small changes:
* Now we run upload-component with dry-run option during PRs. This way we can catch broken component manager manifest files before merging to the master branch
* Trim builds: Build only components required by the example apps. This speeds up build times significantly
* Small typos and comments fixes


espzav (commented):
I didn't test it yet. I will wait for your IDF4/5 reworking. And I haven't got board for test now. I left some comments, other LGTM.

- [x] Please, could you add new display API in this PR too: `lv_disp_t *bsp_display_start_with_config(const bsp_display_cfg_t *cfg);` ?
- [x] After reworking to IDF4/5, please don't forget updet file `.build-test-rules.yml`. Please, could you add `wrover-kit` and `azure-kit` there too?

tore-espressif:
@espzav Thank you for your patience with this PR.

From your last review, I only fixed the recording audio path for IDF v5.x. and added PR description. PTAL!

esp32_s2_kaluga_kit/include/bsp/esp32_s2_kaluga_kit.h
```diff
@@ -74,83 +74,45 @@
...
+ * esp_codec_dev_set_out_vol(spk_codec_dev, DEFAULT_VOLUME);
+ * esp_codec_dev_open(spk_codec_dev, &fs);
+ * esp_codec_dev_write(spk_codec_dev, wav_bytes, wav_bytes_len);
+ * esp_codec_dev_close(spk_codec_dev);
+ * \endcode
+ **************************************************************************************************/
 
 /**
- * @brief Init audio
+ * @brief Initialize speaker codec device
```
espzav, line 104: When I compare it with another boards, here are missing two functions:

```
esp_err_t bsp_audio_init(const i2s_config_t *i2s_config);
#else
esp_err_t bsp_audio_init(const i2s_std_config_t *i2s_config);
#endif
```

```
const audio_codec_data_if_t *bsp_audio_get_codec_itf(void);

```

Can we make it same?

esp32_s2_kaluga_kit/esp32_s2_kaluga_kit.c
```diff
@@ -1,24 +1,32 @@
...
 #include "esp_lvgl_port.h"
+#include "esp_codec_dev_defaults.h"
 #include "bsp_err_check.h"
 
 static const char *TAG = "Kaluga";
 
+static bool i2c_initialized = false;
+
+// Function prototypes
+const audio_codec_data_if_t *bsp_audio_init(void);
```
espzav, line 28: I think, that we should keep it in public API (for make BSP independent on `esp_codec_dev` - something like displays)

esp32_s2_kaluga_kit/esp32_s2_kaluga_kit.c
```diff
@@ -29,7 +37,7 @@ static const touch_pad_t bsp_touch_button[TOUCH_BUTTON_NUM] = {
     TOUCH_BUTTON_GUARD,      /*!< Guard ring for waterproof design. If this pad is touched, other pads no response.*/
 };
 
-const button_config_t bsp_button_config[BSP_BUTTON_NUM] = {
+static const button_config_t bsp_button_config[BSP_BUTTON_NUM] = {
     {
         .type = BUTTON_TYPE_ADC,
         .adc_button_config.adc_channel = ADC_CHANNEL_5, // ADC1 channel 5 is GPIO6
```
espzav, line 43: In IDF5, there should be ADC initialized for keep the ADC handler for other use - e.g. https://github.com/espressif/esp-bsp/blob/a1844b9ace63719a8f8fc3b33a5c187fc165f605/esp32_s3_korvo_2/esp32_s3_korvo_2.c#L48

espzav (changes requested):
@tore-espressif Thank you for this PR. Good job! But it is not consistend with another BSPs now. I left some comments.

esp32_s2_kaluga_kit/include/bsp/esp32_s2_kaluga_kit.h
tore-espressif: IMHO, having different API for v4.4 and v5.x breaks the whole idea of backward compatibility.
But it is probably too late for this discussion... OK, will update

esp32_s2_kaluga_kit/esp32_s2_kaluga_kit.c
tore-espressif: There are some ugly changes required for this.
In Kaluga, we don't use the ADC for any other measurement.
Would it be OK if the user used the second ADC for his measurements?

espzav: We was solving it here: https://github.com/espressif/esp-bsp/pull/201#pullrequestreview-1556913103 Now, the ADC is blocked by buttons and cannot be used other channels of the same ADC unit. After change, user can use it.

tore-espressif:
@espzav thank you for the review, all issues fixed!

espzav (approved):
@tore-espressif Thank you for changes! LGTM

PR #205, opened 2023-08-11
Fix build against esp32_s3_usb_otg and esp32_s2_kaluga_kit comopnent when app does not directly depend on lvgl

tronical:
esp32_s3_usb_otg.h includes lvgl.h and therefore should declare lvgl as a public dependency, like esp-box does.

The same applies to the s2 kaluga kit component, where esp32_s2_kaluga_kit.h includes lvgl.h.


espzav (approved):
@tronical Thank you for this fix! LGTM

PR #216, opened 2023-08-30
lcd_touch_ft5x06: Fixed missing reset.

### LINKED ISSUE #211: no reset signal on reset pin (BSP-370)
no reset signal on reset pin

espzav:

- [x] CI passing

# Change description
Fixed missing reset call in `esp_lcd_touch_ft5x06`. 
Fixed interrupt level in all touch drivers.
Closes #211

### DIFF
--- a/components/lcd_touch/esp_lcd_touch_ft5x06/esp_lcd_touch_ft5x06.c
+++ b/components/lcd_touch/esp_lcd_touch_ft5x06/esp_lcd_touch_ft5x06.c
@@ -242,6 +242,9 @@ static esp_err_t esp_lcd_touch_ft5x06_del(esp_lcd_touch_handle_t tp)
     /* Reset GPIO pin settings */
     if (tp->config.int_gpio_num != GPIO_NUM_NC) {
         gpio_reset_pin(tp->config.int_gpio_num);
+        if (tp->config.interrupt_callback) {
+            gpio_isr_handler_remove(tp->config.int_gpio_num);
+        }
     }
 
     /* Reset GPIO pin settings */
--- a/components/lcd_touch/esp_lcd_touch_ft5x06/esp_lcd_touch_ft5x06.c
+++ b/components/lcd_touch/esp_lcd_touch_ft5x06/esp_lcd_touch_ft5x06.c
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2015-2022 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2015-2023 Espressif Systems (Shanghai) CO LTD
  *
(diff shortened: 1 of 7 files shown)


components/lcd_touch/esp_lcd_touch_ft5x06/esp_lcd_touch_ft5x06.c
Lzw655, line 1: Line 127 also needs to be fixed. It should be like this: `.intr_type = (config->levels.interrupt) ? GPIO_INTR_POSEDGE : GPIO_INTR_NEGEDGE,`

espzav: Thank you for report this issue :-) I fixed it in all touch drivers now.

Lzw655:
@espzav Oh, there is one more thing. We should remove the ISR handler in [del()](https://github.com/espressif/esp-bsp/blob/5a4a565722067decfa057871f8fc7f9975934fab/components/lcd_touch/esp_lcd_touch_ft5x06/esp_lcd_touch_ft5x06.c#L249C1-L249C1) like below:
```c
    if (tp->config.rst_gpio_num != GPIO_NUM_NC) {
        gpio_reset_pin(tp->config.rst_gpio_num);
        if (tp->config.interrupt_callback) {
            gpio_isr_handler_remove(tp->config.int_gpio_num);
        }
    }
```

tore-espressif (commented):
Overall LGTM.

But I agree with @Lzw655 that we could use this PR for fixing the 'delete' behavior.

Moreover, I'd suggest  moving the 'delete' function to the base component `esp_lcd_touch`.
So the `esp_lcd_touch` would deregister the ISR and reset all pins to default. What do you think @espzav ?

espzav:
> Moreover, I'd suggest moving the 'delete' function to the base component `esp_lcd_touch`.
> So the `esp_lcd_touch` would deregister the ISR and reset all pins to default. What do you think @espzav ?

I partially agree, but it wouldn't be consistent, because initialization will be in end driver ( `esp_lcd_touch_xxx`), but removing in `esp_lcd_touch`. Are you sure, that it is good solution?

tore-espressif:
> I partially agree, but it wouldn't be consistent, because initialization will be in end driver ( `esp_lcd_touch_xxx`), but removing in `esp_lcd_touch`. Are you sure, that it is good solution?

True, unless we can move the initialization to the base esp_lcd_touch too, we can leave it as it is

espzav:
@tore-espressif and @Lzw655 Thank you for review. All changes done.

@tore-espressif Could we merge it?

tore-espressif (approved):
LGTM!

PR #217, opened 2023-08-30
esp_lcd: add vendor config

Lzw655:

- [x] CI passing

# Change description

Add a feature for all LCD drivers that uses `vendor_config` to pass external custom initialization commands.

### DIFF
--- a/components/lcd/esp_lcd_gc9503/README.md
+++ b/components/lcd/esp_lcd_gc9503/README.md
@@ -21,61 +21,91 @@ Alternatively, you can create `idf_component.yml`. More is in [Espressif's docum
 
 ## Example use
 
-Create 3-wire SPI panel IO using [esp_lcd_panel_io_additions](https://components.espressif.com/components/espressif/esp_lcd_panel_io_additions) component.
+For most RGB LCDs, they typically use a "[3-Wire SPI + Parallel RGB](https://focuslcds.com/3-wire-spi-parallel-rgb-interface-fan4213/)" interface. The "3-Wire SPI" interface is used for transmitting command data and the "Parallel RGB" interface is used for sending pixel data.
 
-```
+It's recommended to use the [esp_lcd_panel_io_additions](https://components.espressif.com/components/espressif/esp_lcd_panel_io_additions) component to bit-bang the "3-Wire SPI" interface through **GPIO** or an **IO expander** (like [TCA9554](https://components.espressif.com/components/espressif/esp_io_expander_tca9554)). To do this, please first add this component to your project manually. Then, refer to the following code to initialize the GC9503 controller.
+
+```c
+    ESP_LOGI(TAG, "Install 3-wire SPI panel IO");
     spi_line_config_t line_config = {
-        .cs_io_type = IO_TYPE_EXPANDER,
-        .cs_expander_pin = BSP_LCD_SPI_CS,
-        .scl_io_type = IO_TYPE_EXPANDER,
-        .scl_expander_pin = BSP_LCD_SPI_SCK,
-        .sda_io_type = IO_TYPE_EXPANDER,
-        .sda_expander_pin = BSP_LCD_SPI_SDO,
-        .io_expander = expander_handle,
+        .cs_io_type = IO_TYPE_EXPANDER,     // Set to `IO_TYPE_GPIO` if using GPIO, same to below
+        .cs_expander_pin = EXAMPLE_LCD_IO_SPI_CS,
+        .scl_io_type = IO_TYPE_GPIO,
+        .scl_expander_pin = EXAMPLE_LCD_IO_SPI_SCK,
+        .sda_io_type = IO_TYPE_GPIO,
+        .sda_expander_pin = EXAMPLE_LCD_IO_SPI_SDO,
+        .io_expander = expander_handle,     // Set to NULL if not using IO expander
     };
-    esp_lcd_panel_io_3wire_spi_config_t io_config = GC9503_PANEL_IO_3WIRE_SPI_CONFIG(line_config);
+    esp_lcd_panel_io_3wire_spi_config_t io_config = GC9503_PANEL_IO_3WIRE_SPI_CONFIG(line_config, 0);
     esp_lcd_panel_io_handle_t io_handle = NULL;
     ESP_ERROR_CHECK(esp_lcd_new_panel_io_3wire_spi(&io_config, &io_handle));
-```
 
-Initialize GC9503 and create RGB panel.
+/**
+ * Uncomment these line if use custom initialization commands.
+ * The array should be declared as static const and positioned outside the function.
+ */
+// static const gc9503_lcd_init_cmd_t lcd_init_cmds[] = {
+// //   cmd   data        data_size  delay_ms
+//    {0xc1, (uint8_t []){0x3f}, 1, 0},
+//    {0xc2, (uint8_t []){0x0e}, 1, 0},
+//    {0xc6, (uint8_t []){0xf8}, 1, 0},
+//     ...
+// };
 
-```
-    esp_lcd_rgb_panel_config_t panel_conf = {
-        .clk_src = LCD_CLK_SRC_PLL160M,
+    ESP_LOGI(TAG, "Install GC9503 panel driver");
+    esp_lcd_rgb_panel_config_t rgb_config = {
+        .clk_src = LCD_CLK_SRC_DEFAULT,
         .psram_trans_align = 64,
         .data_width = 16,
-        .de_gpio_num = BSP_LCD_DE,
-        .pclk_gpio_num = BSP_LCD_PCLK,
-        .vsync_gpio_num = BSP_LCD_VSYNC,
-        .hsync_gpio_num = BSP_LCD_HSYNC,
+        .bits_per_pixel = 16,
+        .de_gpio_num = EXAMPLE_LCD_IO_RGB_DE,
+        .pclk_gpio_num = EXAMPLE_LCD_IO_RGB_PCLK,
+        .vsync_gpio_num = EXAMPLE_LCD_IO_RGB_VSYNC,
+        .hsync_gpio_num = EXAMPLE_LCD_IO_RGB_HSYNC,
+        .disp_gpio_num = EXAMPLE_LCD_IO_RGB_DISP,
         .data_gpio_nums = {
-            BSP_LCD_DATA0,
-            BSP_LCD_DATA1,
-            BSP_LCD_DATA2,
-            BSP_LCD_DATA3,
-            BSP_LCD_DATA4,
-            BSP_LCD_DATA5,
-            BSP_LCD_DATA6,
-            BSP_LCD_DATA7,
-            BSP_LCD_DATA8,
-            BSP_LCD_DATA9,
-            BSP_LCD_DATA10,
-            BSP_LCD_DATA11,
-            BSP_LCD_DATA12,
-            BSP_LCD_DATA13,
-            BSP_LCD_DATA14,
-            BSP_LCD_DATA15,
+            EXAMPLE_LCD_IO_RGB_DATA0,
+            EXAMPLE_LCD_IO_RGB_DATA1,
+            EXAMPLE_LCD_IO_RGB_DATA2,
+            EXAMPLE_LCD_IO_RGB_DATA3,
+            EXAMPLE_LCD_IO_RGB_DATA4,
+            EXAMPLE_LCD_IO_RGB_DATA5,
+            EXAMPLE_LCD_IO_RGB_DATA6,
+            EXAMPLE_LCD_IO_RGB_DATA7,
+            EXAMPLE_LCD_IO_RGB_DATA8,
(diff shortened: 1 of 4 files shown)


Lzw655:
@espzav PTAL, thank you!

components/lcd/esp_lcd_gc9503/include/esp_lcd_gc9503.h
```diff
@@ -10,26 +10,63 @@
...
+                                                     *   The array should be declared as `static const` and positioned outside the function.
+                                                     *   Please refer to `vendor_specific_init_default` in source file
+                                                     */
+    uint16_t init_cmds_size;    /*<! Number of commands in above array */
+    struct {
+        unsigned int use_external_init_cmds: 1; /*<! Set to 1 if use external initialization commands instead of internal default commands */
+        unsigned int mirror_x: 1;               /*<! Set to 1 if mirror X (by hardware) */
+        unsigned int mirror_y: 1;               /*<! Set to 1 if mirror Y (by hardware)*/
+    } flags;
+} gc9503_vendor_config_t;
```
espzav, line 52: Please, could you update readme with using vendor_config?

components/lcd/esp_lcd_gc9503/include/esp_lcd_gc9503.h
```diff
@@ -10,26 +10,63 @@
...
+typedef struct {
+    const esp_lcd_rgb_panel_config_t *rgb_config;   /*!< RGB panel configuration */
+    const gc9503_lcd_init_cmd_t *init_cmds;         /*!< Pointer to initialization commands array.
+                                                     *   The array should be declared as `static const` and positioned outside the function.
+                                                     *   Please refer to `vendor_specific_init_default` in source file
+                                                     */
+    uint16_t init_cmds_size;    /*<! Number of commands in above array */
+    struct {
+        unsigned int use_external_init_cmds: 1; /*<! Set to 1 if use external initialization commands instead of internal default commands */
+        unsigned int mirror_x: 1;               /*<! Set to 1 if mirror X (by hardware) */
```
espzav, line 49: Please, could you use `mirror()` API instead of this?

components/lcd/esp_lcd_gc9503/include/esp_lcd_gc9503.h
```diff
@@ -10,26 +10,63 @@
...
- * @brief Create LCD panel for GC9503.
+ * @brief LCD panel initialization commands.
+ *
+ */
+typedef struct {
+    int cmd;                /*<! The specific LCD command */
+    const void *data;       /*<! Buffer that holds the command specific data */
+    size_t data_bytes;      /*<! Size of `data` in memory, in bytes */
+    unsigned int delay_ms;  /*<! Delay in milliseconds after this command */
+} gc9503_lcd_init_cmd_t;
```
espzav, line 32: This is good idea, but I am thinking, if there is any solution, how to do it in `esp_lcd` or make any API in `esp_lcd`, which will help with this in all LCD drivers. I would like to see this structure in `esp_lcd`. 
Possible problems:
1. Some LCD drivers needs two byte commands
2. How to implement new API inside `esp_lcd` without breaking changes?

components/lcd/esp_lcd_gc9503/include/esp_lcd_gc9503.h
```diff
@@ -10,26 +10,63 @@
...
+ */
+typedef struct {
+    const esp_lcd_rgb_panel_config_t *rgb_config;   /*!< RGB panel configuration */
+    const gc9503_lcd_init_cmd_t *init_cmds;         /*!< Pointer to initialization commands array.
+                                                     *   The array should be declared as `static const` and positioned outside the function.
+                                                     *   Please refer to `vendor_specific_init_default` in source file
+                                                     */
+    uint16_t init_cmds_size;    /*<! Number of commands in above array */
+    struct {
+        unsigned int use_external_init_cmds: 1; /*<! Set to 1 if use external initialization commands instead of internal default commands */
```
espzav, line 48: Do we need it? I think, we can only check `if(xxx->init_cmds != NULL)`

espzav (commented):
@Lzw655 Thank you for this PR. There is lot of good ideas! I left some comments.

components/lcd/esp_lcd_gc9503/include/esp_lcd_gc9503.h
Lzw655: Yeah, of course. I'll add the example code after we decide to use it.

components/lcd/esp_lcd_gc9503/include/esp_lcd_gc9503.h
Lzw655: In general, RGB screens won't control hardware registers via the SPI interface after initialization. This is because there's a scarcity of chip pins, and we have to free up SPI_SDA and SPI_SCK pins for other functions.

components/lcd/esp_lcd_gc9503/include/esp_lcd_gc9503.h
Lzw655: I have discussed this idea with @suda-morris, and he didn't recommend putting this structure in `esp_lcd`.

1. The `data` has a `void *` type and can be used to pass any bytes commands. Special handling of different LCD commands can be performed internally within the driver.
2. There is no need to add a new API. The usage of this structure is like below, I'll add it into README:
```c
/**
 *  The array should be declared as static const and positioned outside the function.
 */
// static const gc9503_lcd_init_cmd_t lcd_init_cmds[] = {
// //   cmd   data        data_size  delay_ms
//    {0xf0, (uint8_t []){0xc3}, 1, 0},
//    {0xf0, (uint8_t []){0x96}, 1, 0},
//    {0xf0, (uint8_t []){0xa5}, 1, 0},
//     ...
// };

    esp_lcd_rgb_panel_config_t rgb_config= {
        .clk_src = LCD_CLK_SRC_PLL160M,
        .psram_trans_align = 64,
        .data_width = 16,
        .de_gpio_num = BSP_LCD_DE,
        .pclk_gpio_num = BSP_LCD_PCLK,
        .vsync_gpio_num = BSP_LCD_VSYNC,
        .hsync_gpio_num = BSP_LCD_HSYNC,
        .data_gpio_nums = {
            BSP_LCD_DATA0,
            BSP_LCD_DATA1,
            BSP_LCD_DATA2,
            BSP_LCD_DATA3,
            BSP_LCD_DATA4,
            BSP_LCD_DATA5,
            BSP_LCD_DATA6,
            BSP_LCD_DATA7,
            BSP_LCD_DATA8,
            BSP_LCD_DATA9,
            BSP_LCD_DATA10,
            BSP_LCD_DATA11,
            BSP_LCD_DATA12,
            BSP_LCD_DATA13,
            BSP_LCD_DATA14,
            BSP_LCD_DATA15,
        },
        .timings = GC9503_480_480_PANEL_60HZ_RGB_TIMING(),
        .flags.fb_in_psram = 1,
    };
    esp_lcd_panel_handle_t panel_handle = NULL;
    gc9503_vendor_config_t vendor_config = {
        .rgb_config = &rgb_config,
        .init_cmds = lcd_init_cmds,         // Only useful when `use_external_init_cmds` is set to 1
        .init_cmds_size = sizeof(lcd_init_cmds) / sizeof(gc9503_lcd_init_cmd_t),
        .flags = {
            .use_external_init_cmds = 1,    // Set to 1 if use specific initialization commands
        },
    };
    const esp_lcd_panel_dev_config_t panel_config = {
        ...
        .vendor_config = &vendor_config,
       ...
    };
    ESP_ERROR_CHECK(esp_lcd_new_panel_gc9503(io_handle, &panel_config, &panel_handle));
    ESP_ERROR_CHECK(esp_lcd_panel_reset(panel_handle));
    ESP_ERROR_CHECK(esp_lcd_panel_init(panel_handle));
```

components/lcd/esp_lcd_gc9503/include/esp_lcd_gc9503.h
Lzw655: This flag is indeed not that necessary, but it's worth considering a scenario where it might be needed: if a user doesn't want to initialize the screen, similar to the ST7789 driver in IDF.

components/lcd/esp_lcd_gc9503/include/esp_lcd_gc9503.h
espzav: Ok, I understand. It's a pitty, that we cannot use mirror after init.

Lzw655: Yeah, or maybe we can use a new flag like below and overwrite `mirror()` in the driver. I'll try this way later.
```c
 unsigned int mirror_use_hardware: 1;       /*<! Set to 1 if expect `esp_lcd_panel_mirror()` be implemented through hardware */
```

components/lcd/esp_lcd_gc9503/include/esp_lcd_gc9503.h
espzav: If a user doesn't want to use external cmds, then put only NULL to `init_cmds` and it takes defaults.

Lzw655: What I mean is that users don't want to send any initialization commands, including within the driver itself.

components/lcd/esp_lcd_gc9503/include/esp_lcd_gc9503.h
Lzw655: @espzav I finish the function to use `mirror()` instead of this. PTAL, thanks!

components/lcd/esp_lcd_gc9503/include/esp_lcd_gc9503.h
Lzw655: This flag has been removed.

Lzw655:
@tom-borcin PTAL, thank you!

espzav (approved):
@Lzw655 Great work! Thank you. It is a little complicated than usual LCD drivers, but I think, that it cannot be simpler. LGTM
@tore-espressif PTAL

components/lcd/esp_lcd_gc9503/README.md
```diff
@@ -19,63 +19,84 @@ You can add them to your project via `idf.py add-dependancy`, e.g.
 
 Alternatively, you can create `idf_component.yml`. More is in [Espressif's documentation](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/tools/idf-component-manager.html).
 
-## Example use
+## Initialization Code
 
-Create 3-wire SPI panel IO using [esp_lcd_panel_io_additions](https://components.espressif.com/components/espressif/esp_lcd_panel_io_additions) component.
```
tore-espressif, line 24: Could we add a short paragraph that explains 3wire SPI + RGB interface and that we support GPIOs through IO expander?

components/lcd/esp_lcd_gc9503/README.md
```diff
@@ -19,63 +19,84 @@ You can add them to your project via `idf.py add-dependancy`, e.g.
...
-
-```
+```c
+    // Using component esp_lcd_panel_io_additions: https://components.espressif.com/components/espressif/esp_lcd_panel_io_additions
+    ESP_LOGI(TAG, "Install 3-wire SPI panel IO");
     spi_line_config_t line_config = {
```
tore-espressif, line 27: This type is defined in [esp_lcd_panel_io_additions](https://components.espressif.com/components/espressif/esp_lcd_panel_io_additions?language=en), right? Could we add a note that this component must be explicitely added to user's project?

(since it is not liste in `idf_component.yml` of this component?)

tore-espressif (changes requested):
Hi @Lzw655 I left two comments about docs, apart from that LGTM!

components/lcd/esp_lcd_gc9503/README.md
Lzw655: Sure, I added it. PTAL!

components/lcd/esp_lcd_gc9503/README.md
Lzw655: Yeah, you're right. Already done.

Lzw655:
@espzav @tore-espressif If you approve of this use of the vendor_config method, I will proceed to make similar modifications to other LCD drivers.

Lzw655:
@tore-espressif @espzav Should I modify all LCD drivers in this PR or split them individually?

tore-espressif:
> I think, that we can do it in next PR. @tore-espressif can we merge it?

Yep, LGTM

espzav:
@Lzw655 Please squash commits and rebase master, then I will merge it. Thank you

espzav:
@Lzw655 Please, is there any reason, why there are two commits instead of one?

Lzw655:
> @Lzw655 Please, is there any reason, why there are two commits instead of one?

Oh, sorry. 

> @Lzw655 Please, is there any reason, why there are two commits instead of one?

Oh, sorry. Rebased.

PR #219, opened 2023-09-05
esp-box: Prevent I2C bus form getting stuck. (BSP-377)

espressif2022:

- [x] CI passing

# Change description
Iot button call `bsp_get_main_button` too busy, it'll block I2C bus.

### DIFF
--- a/esp-box/esp-box.c
+++ b/esp-box/esp-box.c
@@ -495,7 +495,6 @@ bool bsp_button_get(const bsp_button_t btn)
 static uint8_t bsp_get_main_button(void *param)
 {
     assert(tp);
-    ESP_ERROR_CHECK(esp_lcd_touch_read_data(tp));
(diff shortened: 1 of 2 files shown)


esp-box/esp-box.c
```diff
@@ -495,7 +495,6 @@ bool bsp_button_get(const bsp_button_t btn)
 static uint8_t bsp_get_main_button(void *param)
 {
     assert(tp);
-    ESP_ERROR_CHECK(esp_lcd_touch_read_data(tp));
```
espzav, line 498: Are you sure, that it will be working without this line?

espressif2022: > Are you sure, that it will be working without this line?

Yes, it worked.
Because lvgl will call `esp_lcd_touch_read_data`, it'll get the btn event of the TP.

espzav: Ok, you are right. I think, that it is OK now.

espzav (approved):
LGTM

PR #220, opened 2023-09-05
esp-box-lite: Add initialization and definition of IO button (BSP-376)

espressif2022:

- [x] CI passing

# Change description
Add initialization and definition of IO button in BSP.

I want to provide the same interface as esp-box here.

### DIFF
--- a/esp-box-lite/esp-box-lite.c
+++ b/esp-box-lite/esp-box-lite.c
@@ -23,19 +23,36 @@
 
 static const char *TAG = "ESP-BOX-LITE";
 
+/* ADC Buttons */
+typedef enum {
+    BSP_ADC_BUTTON_PREV,
+    BSP_ADC_BUTTON_ENTER,
+    BSP_ADC_BUTTON_NEXT,
+    BSP_ADC_BUTTON_NUM
+} bsp_adc_button_t;
+
 static lv_indev_t *disp_indev = NULL;   /* Input device (buttons) */
 #if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 0, 0)
 static adc_oneshot_unit_handle_t bsp_adc_handle = NULL;
 #endif
+static bool i2c_initialized = false;
 
 static const button_config_t bsp_button_config[BSP_BUTTON_NUM] = {
+    {
+        .type = BUTTON_TYPE_GPIO,
+        .gpio_button_config.active_level = false,
+        .gpio_button_config.gpio_num = BSP_BUTTON_CONFIG_IO,
+    },
+};
+
+static const button_config_t bsp_adc_button_config[BSP_ADC_BUTTON_NUM] = {
     {
         .type = BUTTON_TYPE_ADC,
 #if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 0, 0)
         .adc_button_config.adc_handle = &bsp_adc_handle,
 #endif
         .adc_button_config.adc_channel = ADC_CHANNEL_0, // ADC1 channel 0 is GPIO1
-        .adc_button_config.button_index = BSP_BUTTON_PREV,
+        .adc_button_config.button_index = BSP_ADC_BUTTON_PREV,
         .adc_button_config.min = 2310, // middle is 2410mV
         .adc_button_config.max = 2510
     },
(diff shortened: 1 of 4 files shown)


esp-box-lite/esp-box-lite.c
```diff
@@ -415,3 +420,30 @@ void bsp_display_unlock(void)
 {
     lvgl_port_unlock();
 }
+
+esp_err_t bsp_iot_button_create(button_handle_t btn_array[], int *btn_cnt, int btn_array_size)
+{
+    esp_err_t ret = ESP_OK;
+    if ((btn_array_size < BSP_BUTTON_NUM) ||
```
espzav, line 427: The array `btn_array_size` can be smaller (without BSP_BUTTON_PREV, BSP_BUTTON_ENTER, BSP_BUTTON_NEXT buttons)

espzav (commented):
@espressif2022 Thank you for add new button API. I didn't add before, because I didn't want to provide navigation buttons to user. This is good idea, how to do it, but user can add smaller array.

esp-box-lite/esp-box-lite.c
espressif2022: > The array `btn_array_size` can be smaller (without BSP_BUTTON_PREV, BSP_BUTTON_ENTER, BSP_BUTTON_NEXT buttons)

Yes，I have the same consideration, this operation wastes three useless handles.
But I want to maintain the consistency of the interface.
I didn't think of a particularly suitable method, do you have any suggestion here?

espzav: If I am thinking about it, we can remove navigation buttons from `bsp_button_t` and from `bsp_button_config`. We can left only public buttons for user usage and this three buttons will be private onsice `.c` file.
What do you think?

espressif2022: > If I am thinking about it, we can remove navigation buttons from `bsp_button_t` and from `bsp_button_config`. We can left only public buttons for user usage and this three buttons will be private onsice `.c` file. What do you think?

Yes,  it needs to update `esp_lv_port`. We update this time, or the next step?

espzav: Why need update `esp_lvgl_port`? I thought only move these three buttons into `esp-box-lite.c` and use it same.

espressif2022: > Why need update `esp_lvgl_port`? I thought only move these three buttons into `esp-box-lite.c` and use it same.


OK，I understand what you mean, just define a private strcut for ADC button here.
I thought you wanted to take out the initialization of the `navigation buttons` in `esp_lv_port` and pass them the handle.

espzav (approved):
@espressif2022 Thank you! Good job! 
Please, what do you think about it @tore-espressif?

tore-espressif:
Thank you for the update @espressif2022 !

Could you also please update the I2C init for esp-box-lite? I have it already implemented here https://github.com/espressif/esp-bsp/pull/196/files#diff-5c9eb54c68657f5d1534daa481e0e1e58aeed2be68b93ad9aa7aada7ff19b32f
But I'd like to avoid releasing of new versions too often for every small fix :)

esp-box-lite/esp-box-lite.c
tore-espressif: So the user won't be able to user the navigation buttons without LVGL, right? Hmmm, seems like a breaking change to me...

Are you sure this is the best option?

espzav: Before user couldn't use it too, because it was used for navigation. It was only in the same structure. I think, this is better.

tore-espressif: OK

espressif2022:
> Thank you for the update @espressif2022 !
> 
> Could you also please update the I2C init for esp-box-lite? I have it already implemented here https://github.com/espressif/esp-bsp/pull/196/files#diff-5c9eb54c68657f5d1534daa481e0e1e58aeed2be68b93ad9aa7aada7ff19b32f But I'd like to avoid releasing of new versions too often for every small fix :)

OK，Got it.
This modification is special, because I want to unify the interface of the three boxes .

espzav:
@espressif2022 Please squash commits and rebase changes from master. Then I will merge it. Thank you

PR #224, opened 2023-09-14
i2c: Fixed i2c init.

espzav:

- [ ] CI passing

# Change description
Fixed I2C initialization.


tore-espressif (approved):
LGTM. For future, we might consider packing multiple fixes into one release
(that is, we don't have to release after every few lines change)

espzav:
> LGTM. For future, we might consider packing multiple fixes into one release (that is, we don't have to release after every few lines change)

Ok, I understand. I only thought, that this fix is not same as other fixes in examples.

PR #226, opened 2023-09-18
lcd_touch_gt911: Add low_power interface and button. (BSP-380)

espressif2022:

- [x] CI passing

# Change description

- Add low-power interfaces.
- Add multiple Addr options.

### DIFF
--- a/components/lcd_touch/esp_lcd_touch_gt911/esp_lcd_touch_gt911.c
+++ b/components/lcd_touch/esp_lcd_touch_gt911/esp_lcd_touch_gt911.c
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2015-2022 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2015-2023 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -20,15 +20,23 @@
 static const char *TAG = "GT911";
 
 /* GT911 registers */
-#define ESP_LCD_TOUCH_GT911_READ_XY_REG (0x814E)
-#define ESP_LCD_TOUCH_GT911_CONFIG_REG  (0x8047)
-#define ESP_LCD_TOUCH_GT911_PRODUCT_ID_REG (0x8140)
+#define ESP_LCD_TOUCH_GT911_READ_KEY_REG    (0x8093)
+#define ESP_LCD_TOUCH_GT911_READ_XY_REG     (0x814E)
+#define ESP_LCD_TOUCH_GT911_CONFIG_REG      (0x8047)
+#define ESP_LCD_TOUCH_GT911_PRODUCT_ID_REG  (0x8140)
+#define ESP_LCD_TOUCH_GT911_ENTER_SLEEP     (0x8040)
+
+/* GT911 support key num */
+#define ESP_GT911_TOUCH_MAX_BUTTONS         (4)
 
 /*******************************************************************************
 * Function definitions
 *******************************************************************************/
 static esp_err_t esp_lcd_touch_gt911_read_data(esp_lcd_touch_handle_t tp);
 static bool esp_lcd_touch_gt911_get_xy(esp_lcd_touch_handle_t tp, uint16_t *x, uint16_t *y, uint16_t *strength, uint8_t *point_num, uint8_t max_point_num);
+#if (CONFIG_ESP_LCD_TOUCH_MAX_BUTTONS > 0)
+static esp_err_t esp_lcd_touch_gt911_get_button_state(esp_lcd_touch_handle_t tp, uint8_t n, uint8_t *state);
+#endif
 static esp_err_t esp_lcd_touch_gt911_del(esp_lcd_touch_handle_t tp);
 
 /* I2C read/write */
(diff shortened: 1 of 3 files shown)


components/lcd_touch/esp_lcd_touch_gt911/esp_lcd_touch_gt911.c
```diff
@@ -132,7 +180,17 @@ static esp_err_t esp_lcd_touch_gt911_read_data(esp_lcd_touch_handle_t tp)
 
     /* Any touch data? */
     if ((buf[0] & 0x80) == 0x00) {
+#if (CONFIG_ESP_LCD_TOUCH_MAX_BUTTONS > 0)
+        tp->data.button[0].status = 0x00;
+#endif
         touch_gt911_i2c_write(tp, ESP_LCD_TOUCH_GT911_READ_XY_REG, clear);
+    } else if ((buf[0] & 0x10) == 0x10) {
```
espzav, line 187: I am sorry, I overlooked something maybe. What does it mean 0x10 ? I checked the GT911 registers and I see this bit is reserved in 0x814E register.
I see this:
image

components/lcd_touch/esp_lcd_touch_gt911/esp_lcd_touch_gt911.c
```diff
@@ -132,7 +180,17 @@ static esp_err_t esp_lcd_touch_gt911_read_data(esp_lcd_touch_handle_t tp)
 
     /* Any touch data? */
     if ((buf[0] & 0x80) == 0x00) {
+#if (CONFIG_ESP_LCD_TOUCH_MAX_BUTTONS > 0)
+        tp->data.button[0].status = 0x00;
```
espzav, line 184: How many buttons are supported in GT911. Should it be cleared all supported buttons?

espzav (commented):
@espressif2022 Thank you for update this driver. I tested button on HW. Sleep mode not tested, I haven't got example for it now and I am really busy. I am sorry for waiting too long. I left some comments.

components/lcd_touch/esp_lcd_touch_gt911/esp_lcd_touch_gt911.c
espressif2022: > I am sorry, I overlooked something maybe. What does it mean 0x10 ? I checked the GT911 registers and I see this bit is reserved in 0x814E register. I see this: image

Yes, you are right. some documents do not explain this bit, it's not very clear about this key event.
image
image

components/lcd_touch/esp_lcd_touch_gt911/esp_lcd_touch_gt911.c
espressif2022: > How many buttons are supported in GT911. Should it be cleared all supported buttons?

According to the screenshots shown above, it seems that I can only get one event.

components/lcd_touch/esp_lcd_touch_gt911/esp_lcd_touch_gt911.c
espzav: I cannot find newer datasheet, but it looks like only information that some key was changed. I have some older version with lot of chinesse letters, but I see, that there are supported four buttons (registers 0x8093 - 0x8096). If the GT11 controller support all four buttons, we should support all of them in this driver.

espressif2022: > I cannot find newer datasheet, but it looks like only information that some key was changed. I have some older version with lot of chinesse letters, but I see, that there are supported four buttons (registers 0x8093 - 0x8096). If the GT11 controller support all four buttons, we should support all of them in this driver.

OK, I'll check this four button event.

espzav (commented):
@espressif2022 Great! Thank you! Only one comment left.
And did you tested the sleep?

components/lcd_touch/esp_lcd_touch_gt911/esp_lcd_touch_gt911.c
```diff
@@ -181,15 +185,32 @@ static esp_err_t esp_lcd_touch_gt911_read_data(esp_lcd_touch_handle_t tp)
     /* Any touch data? */
     if ((buf[0] & 0x80) == 0x00) {
 #if (CONFIG_ESP_LCD_TOUCH_MAX_BUTTONS > 0)
-        tp->data.button[0].status = 0x00;
+        for (i = 0; i < CONFIG_ESP_LCD_TOUCH_MAX_BUTTONS; i++) {
```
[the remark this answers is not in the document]
espressif2022: when all the touch button liftoff, we'll not enter the case `else if ((buf[0] & 0x10) == 0x10) {` 
So I think we need this to clean the button event.

espressif2022:
> @espressif2022 Great! Thank you! Only one comment left. And did you tested the sleep?

Yeah, I've measured the current on the power supply pin of TP before.

components/lcd_touch/esp_lcd_touch_gt911/esp_lcd_touch_gt911.c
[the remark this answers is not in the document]
espzav: Are you sure? I think, that this bit is set, when some button state changed. Or is it only set, when some button is true and thus bit is not set, when button state is false?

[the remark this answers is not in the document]
espressif2022: > Are you sure? I think, that this bit is set, when some button state changed. Or is it only set, when some button is true and thus bit is not set, when button state is false?

I've double check it.
when btn pressed, it'll report event 0x90`(once)` 0x00 0x00 0x00`(Infinite loop)`
when btn release, it'll  report event 0x80`(once)` 0x00 0x00 0x00`(Infinite loop)`

Maybe we shoule clean the event in the 0x80 event ?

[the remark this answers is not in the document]
espzav: The event 0x80 is same for coordinates, then it should be instead of `else` ==> `else if((buf[0] & 0x80) == 0x80)` and there should be clear of buttons, what do you think?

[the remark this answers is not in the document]
espressif2022: > The event 0x80 is same for coordinates, then it should be instead of `else` ==> `else if((buf[0] & 0x80) == 0x80)` and there should be clear of buttons, what do you think?

Yeah, it's better, I've updated it.

espzav (approved):
@espressif2022 Great job! Thank you very much! LGTM
Please, squash commits then I will merge it.

PR #228, opened 2023-09-25
icm42670: Add 6-axis MotionTracking

espzav:
# ESP-BSP Pull Request checklist

- [x] Version of modified component bumped
- [x] CI passing

# Change description
Component for 6-axis MotionTracking icm42670/icm42607.

PR #229, opened 2023-09-26
bsp: Add support of MotionSensor to ESP-BOXes

espzav:

- [x] CI passing

# Change description

The first must be merged this component #228

Add new API for initialization MotionSensor in ESP-BOX and ESP-BOX-3 (It can be added into ESP-BOX-Lite too, but there is not assembled this controller).

### DIFF
--- a/esp-box-3/include/bsp/esp-box-3.h
+++ b/esp-box-3/include/bsp/esp-box-3.h
@@ -27,6 +27,19 @@
 #include "driver/i2s_std.h"
 #endif
 /**************************************************************************************************
+ *  BSP Capabilities
+ **************************************************************************************************/
+
+#define BSP_CAPS_DISPLAY        1
+#define BSP_CAPS_TOUCH          1
+#define BSP_CAPS_BUTTONS        1
+#define BSP_CAPS_AUDIO          1
+#define BSP_CAPS_AUDIO_SPEAKER  1
+#define BSP_CAPS_AUDIO_MIC      1
+#define BSP_CAPS_SDCARD         1
+#define BSP_CAPS_IMU            1
+
+/**************************************************************************************************
  *  ESP-BOX pinout
  **************************************************************************************************/
 /* I2C */
--- a/esp-box-3/idf_component.yml
+++ b/esp-box-3/idf_component.yml
@@ -1,5 +1,5 @@
 
-version: "1.1.0"
+version: "1.1.1"
 description: Board Support Package for ESP32-S3-BOX-3
 url: https://github.com/espressif/esp-bsp/tree/master/esp-box-3
 
@@ -22,6 +22,10 @@ dependencies:
 
   button:
     version: "^2.5"
+    public: true
+
+  icm42670:
+    version: "^1"
     public: true
 
 examples:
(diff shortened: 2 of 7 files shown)


esp-box-3/include/bsp/esp-box-3.h
```diff
@@ -20,12 +20,26 @@
...
 #if ESP_IDF_VERSION < ESP_IDF_VERSION_VAL(5, 0, 0)
 #include "driver/i2s.h"
 #else
 #include "driver/i2s_std.h"
 #endif
+/**************************************************************************************************
+ *  BSP Capabilities
+ **************************************************************************************************/
+
+#define BSP_HAS_DISPLAY        1
```
tore-espressif, line 34: Could we rename to `BSP_CAPS_DISPLAY`? This way it will be similar to esp-idf `SOC_CAPS_...`

esp-box-3/esp-box-3.c
```diff
@@ -626,3 +626,24 @@ esp_err_t bsp_iot_button_create(button_handle_t btn_array[], int *btn_cnt, int b
...
+    /* Configuration of the acceleremeter and gyroscope */
+    const icm42670_cfg_t imu_cfg = {
+        .acce_fs = ACCE_FS_2G,
+        .acce_odr = ACCE_ODR_400HZ,
+        .gyro_fs = GYRO_FS_2000DPS,
+        .gyro_odr = GYRO_ODR_400HZ,
+    };
+    ESP_ERROR_CHECK(icm42670_config(handle, &imu_cfg));
+
+    return handle;
```
tore-espressif, line 648: `handle` is of type `icm42670_handle_t` but return type is `bsp_imu_t`. How do you want to keep this in sync?

tore-espressif (commented):
Few questions for the begining :)

esp-box-3/include/bsp/esp-box-3.h
espzav: Renamed.

esp-box-3/esp-box-3.c
espzav: I added cast to return. Is it ok?

espzav:
@tore-espressif Thank you for review. All issues fixed.

esp-box-3/include/bsp/esp-box-3.h
```diff
@@ -127,6 +141,12 @@ typedef struct {
     lvgl_port_cfg_t lvgl_port_cfg;
 } bsp_display_cfg_t;
 
+/**
+ * @brief BSP IMU type
+ *
+ */
+typedef icm42670_handle_t bsp_imu_t;
```
tore-espressif, line 148: I assume that we plan to create more `typedef some_imu_handle_t bsp_imu_t;` in other BSPs.

This will create very loose dependency, that will be very easy to break in the future.

The example still uses `icm42670_xxx` API. What is the benefit of this type then?

espzav: I thought, that we will have same API for init for all IMU `bsp_imu_t bsp_imu_init(void)`. And I thought, that we can use anz IMU in example, but sure, there is API from specific IMU. 
Ok, it was bad idea. 

I would like see independent API for everything inside BSP. But if we really want it, there must be something like `esp_lcd_touch` or somethung like that :-(

espzav:
Moved IMU initialization into example and removed from BSP. 
@tore-espressif PTAL

tore-espressif (approved):
LGTM!

PR #245, opened 2023-10-30
fix(s3_eye): Make esp_lvgl_port dependency public

tore-espressif:
Quick fix for private dependency `esp_lvgl_port`


espzav (approved):
@tore-espressif Thank you for fix! LGTM

PR #249, opened 2023-11-10
Feature/update box3 lcd driver (BSP-415)

espressif2022:

- [x] CI passing

# Change description

- use vendor_config to update LCD driver of ILI9341.
- update button version to use lastest feature.

### DIFF
--- a/bsp/esp-box-3/esp-box-3.c
+++ b/bsp/esp-box-3/esp-box-3.c
@@ -33,6 +33,26 @@ static const char *TAG = "ESP-BOX-3";
 _Static_assert(CONFIG_ESP_LCD_TOUCH_MAX_BUTTONS > 0, "Touch buttons must be supported for this BSP");
 /** @endcond */
 
+static const ili9341_lcd_init_cmd_t vendor_specific_init[] = {
+    {0xC8, (uint8_t []){0xFF, 0x93, 0x42}, 3, 0},
+    {0xC0, (uint8_t []){0x0E, 0x0E}, 2, 0},
+    {0xC5, (uint8_t []){0xD0}, 1, 0},
+    {0xC1, (uint8_t []){0x02}, 1, 0},
+    {0xB4, (uint8_t []){0x02}, 1, 0},
+    {0xE0, (uint8_t []){0x00, 0x03, 0x08, 0x06, 0x13, 0x09, 0x39, 0x39, 0x48, 0x02, 0x0a, 0x08, 0x17, 0x17, 0x0F}, 15, 0},
+    {0xE1, (uint8_t []){0x00, 0x28, 0x29, 0x01, 0x0d, 0x03, 0x3f, 0x33, 0x52, 0x04, 0x0f, 0x0e, 0x37, 0x38, 0x0F}, 15, 0},
+
+    {0xB1, (uint8_t []){00, 0x1B}, 2, 0},
+    {0x36, (uint8_t []){0x08}, 1, 0},
+    {0x3A, (uint8_t []){0x55}, 1, 0},
+    {0xB7, (uint8_t []){0x06}, 1, 0},
+
+    {0x11, (uint8_t []){0}, 0x80, 0},
+    {0x29, (uint8_t []){0}, 0x80, 0},
+
+    {0, (uint8_t []){0}, 0xff, 0},
+};
+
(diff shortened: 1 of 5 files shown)


bsp/esp-box-lite/idf_component.yml
```diff
@@ -9,7 +9,7 @@ dependencies:
   idf: ">=4.4.5"
 
   button:
-    version: "^2"
+    version: "^3"
```
tore-espressif, line 12: Why do we need to bump this version requirement?

bsp/esp-box-3/esp-box-3.c
```diff
@@ -33,6 +33,26 @@ static const char *TAG = "ESP-BOX-3";
 _Static_assert(CONFIG_ESP_LCD_TOUCH_MAX_BUTTONS > 0, "Touch buttons must be supported for this BSP");
 /** @endcond */
 
+static const ili9341_lcd_init_cmd_t vendor_specific_init[] = {
```
tore-espressif, line 36: Could you please describe this change more? Will we get better display colors with this settings?

bsp/esp-box-lite/idf_component.yml
tore-espressif: IDF-component-manager will automatically select latest compatible version. We shouls still support 2.* versions. We can do something like `>=2`

espressif2022: > IDF-component-manager will automatically select latest compatible version. We shouls still support 2.* versions. We can do something like `>=2`

OK, thanks, I'll check and update it.

bsp/esp-box-3/esp-box-3.c
espressif2022: Yes, the new setting will get better colors.
Previously, the VCOM Register provided by the manufacturer were not correct, causing streaks on the screen, We always thought it was a screen issue.
After communicating and verifying with other manufacturers, we found that the VOM register mentioned above was unreasonable, so we need update it.

bsp/esp-box-lite/idf_component.yml
espressif2022: > Why do we need to bump this version requirement?

Because I'll pull three BSPs together (esp-box, esp-box-lite, esp-box-3).
For esp-box-3, I need the feature in lastest version(>=3), there was a conflict here and I used a local components before. I want to keep the same here, ans  I can delete the local btn components.

espzav:
@espressif2022 Thank you for this PR, I must check it on HW and I am sorry, I haven't got the HW now. (I will do it next week)

espressif2022:
> @espressif2022 Thank you for this PR, I must check it on HW and I am sorry, I haven't got the HW now. (I will do it next week)

OK, I'm updating component esp_lv_port too, I want put the full framebuf in PSRAM.

bsp/esp-box/idf_component.yml
```diff
@@ -19,7 +19,7 @@ dependencies:
     public: true
 
   button:
-    version: "^2.5"
```
espzav, line 22: Minimal version for >=IDF5 is 2.4.1 (there is new ADC driver)
- custom buttons are available from 2.5 (main button)

bsp/esp-box-3/idf_component.yml
```diff
@@ -21,7 +21,7 @@ dependencies:
     public: true
 
   button:
-    version: "^2.5"
+    version: ">=2"
```
espzav, line 24: Minimal version for >=IDF5 is 2.4 (there is new ADC driver)
- custom buttons are available from 2.5 (main button)

bsp/esp-box-lite/idf_component.yml
```diff
@@ -9,7 +9,7 @@ dependencies:
   idf: ">=4.4.5"
 
   button:
-    version: "^2"
```
espzav, line 12: Minimal version for >=IDF5 is 2.4 (there is new ADC driver)

espzav (commented):
@espressif2022 Thank you for this PR. Great work with LCD params. It is really better, but still viewing angles are bad :-( 
Left some comments.

espzav (approved):
@espressif2022 Thank you! Please squash commits and rebase master.

PR #253, opened 2023-11-27
button: Fix button component version in all BSPs.

espzav:

- [ ] CI passing

# Change description
Allow to use latest version of button and led_indicator component for fix compilation errors and warnings after changes in IDF master.

### DIFF
--- a/bsp/esp_wrover_kit/idf_component.yml
+++ b/bsp/esp_wrover_kit/idf_component.yml
@@ -1,4 +1,4 @@
-version: "1.5.0~1"
+version: "1.5.0~2"
 description: Board Support Package for ESP-WROVER-KIT
 url: https://github.com/espressif/esp-bsp/tree/master/bsp/esp_wrover_kit
 
@@ -13,7 +13,7 @@ dependencies:
     public: true
 
   button:
-    version: "^2.5"
+    version: ">=2.5,<4.0"
     public: true
 
 examples:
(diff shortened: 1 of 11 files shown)


bsp/esp_wrover_kit/idf_component.yml
```diff
@@ -13,7 +13,7 @@ dependencies:
     public: true
 
   button:
-    version: "^2.5"
+    version: ">=2.5"
```
igrr, line 16: I'm not sure this is a future-proof way of setting the version. If there is a new release of esp-button, v4.0, and v4.0 is not backwards-compatible to v3.0, then esp_wrover_kit component compilation may start failing.

If you are using `>=`, then you should also specify an upper bound, to avoid pulling a non-compatible version. For example: `>=2.5,<4.0` (assuming that esp_wrover_kit is compatible with `button==3.*`).

Same comment for other changes in the PR.

igrr (changes requested):

bsp/esp_wrover_kit/idf_component.yml
espzav: I understand. I was thinking about it and it is problem of versioning of components. We can put for example version 3: "^3", but when something will change, they change to version 4.x and fix errors only in version 4. Same like this situation. We had version 2.5 and fix is only in version 3.1. 

Ok, I will put  max version `<4.0`

igrr (approved):

PR #254, opened 2023-12-01
gc9503: fix madctl default value

Lzw655:
# ESP-BSP Pull Request checklist

> Note: For new BSPs create a PR with this [link](https://github.com/espressif/esp-bsp/compare/main...my-branch?quick_pull=1&template=pr_template_bsp.md).

- [ ] Version of modified component bumped
- [ ] CI passing

# Change description
* The MADCTL command of GC9503 has a default value `0x10`
* If set to 0, the screen will display abnormally when running LVGL.

PR #256, opened 2023-12-02
bsp: update s3-lcd-ev-board

Lzw655:

- [ ] CI passing

# Change description

For BSP:

* Bugfix
  * Use `on_bounce_frame_finish` instead of `on_vsync` when enabling RGB bounce buffer mode.
  * Fix some issues when enabling anti-tearing and rotation
* Features
  * Configurations:
    * Support to set the pinned core for the LVGL task
  * Implementations:
    * Support ESP32-S3-WROOM-1-N16R16V module
    * Add warning message when compiling and running if using `ESP-IDF` version `<5.1.2`
  * Dependencies
    * Update the version of `ESP-IDF` to `>5.0.1`
    * Use `esp_lcd_gc9503` version `^1` when using `ESP-IDF` version `<5.1.2`
    * Use `esp_lcd_gc9503` version `^3` when using `ESP-IDF` version `>=5.1.2`

For example *display_lvgl_demos*:
* Update sdkconfig and README

For squareline board:
* Update sdkconfig


Lzw655:
It will be ready after #254 is merged.

espzav (commented):
@Lzw655 Thank you for these changes. I didn't test it on HW yet. I will do it soon. LGTM

SquareLine/boards/esp32_s3_lcd_ev_board/main/idf_component.yml
```diff
@@ -2,5 +2,5 @@ description: ESP-BSP SquareLine LVGL Example
 targets:
     - esp32s3
 dependencies:
-  idf: ">=4.4"
+  idf: ">5.0.4,!=5.1.1"
```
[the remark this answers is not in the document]
Lzw655: Sure, deleted.

Lzw655:
@espzav Thank you for reviewing. After increasing the version of IDF to `5.1.2`, I thought maybe it would be better to keep compatible with IDF `>=5.0.1` (v5.0 doesn't support RGB triple-buffer). So I specified the different versions of GC9503 according to the version of IDF and added warning messages when using IDF `<5.1.2`.

Lzw655:
> @Lzw655 Whenever I change `BSP_LCD_RGB_REFRESH_MODE` in menuconfig to 'auto' or 'manual', this image is broken. Could you please have a look at those modes?

Oh, I think you probably enable `CONFIG_ESP32S3_DATA_CACHE_LINE_64B=y` meanwhile. If the `PSRAM DDR 120M`  is not enabled or `BSP_LCD_RGB_REFRESH_MODE` is not `bounce buffer` mode, `CONFIG_ESP32S3_DATA_CACHE_LINE_64B=y` shouldn't be enabled to avoid screen drifting.

examples/display_lvgl_demos/README.md
```diff
@@ -2,32 +2,64 @@
...
-* Flash:
-    * `ESPTOOLPY_FLASHMODE_QIO`
-    * `ESPTOOLPY_FLASHFREQ_120M`
-* PSRAM:
-    * `SPIRAM_MODE_OCT`
-    * `SPIRAM_SPEED_120M` (See [here](https://github.com/espressif/esp-dev-kits/tree/master/esp32-s3-lcd-ev-board#psram-120m-ddr) to enbale this feature of ESP-IDF)
-    * `SPIRAM_FETCH_INSTRUCTIONS`
-    * `SPIRAM_RODATA`
-* Cache:
-    * `ESP32S3_DATA_CACHE_LINE_64B` (It can be enabled only when using bounce buffer or PSRAM with Octal 120M. Otherwise it will cause screen drift.)
```
tore-espressif, line 22: Why is this note deleted?

tore-espressif (commented):
> CONFIG_ESP32S3_DATA_CACHE_LINE_64B=y shouldn't be enabled to avoid screen drift.

@Lzw655 thanks for the explanation. Would it be possible to add a runtime warning in case the user configures `CONFIG_ESP32S3_DATA_CACHE_LINE_64B=y` and uses Auto or Manual mode?

Since the `CONFIG_ESP32S3_DATA_CACHE_LINE_64B=y` is enabled in the example by default, more users can come across this issue.

examples/display_lvgl_demos/README.md
Lzw655: The info here is redundant and already present in the existing [docs](https://docs.espressif.com/projects/esp-faq/en/latest/software-framework/peripherals/lcd.html#how-can-i-improve-the-display-frame-rate-of-lcd-screens). I think we shouldn't maintain two of them. 

But it's better to add info that guides users to find it, so I added a link in README. What do you think?

Lzw655:
> > CONFIG_ESP32S3_DATA_CACHE_LINE_64B=y shouldn't be enabled to avoid screen drift.
> 
> @Lzw655 thanks for the explanation. Would it be possible to add a runtime warning in case the user configures `CONFIG_ESP32S3_DATA_CACHE_LINE_64B=y` and uses Auto or Manual mode?
> 
> Since the `CONFIG_ESP32S3_DATA_CACHE_LINE_64B=y` is enabled in the example by default, more users can come across this issue.

You're right. It was added. PTAL.

bsp/esp32_s3_lcd_ev_board/idf_component.yml
```diff
@@ -1,12 +1,12 @@
-version: "2.0.0~2"
+version: "3.0.0"
```
tore-espressif, line 1: Are there any breaking changes in this change? If not, we can bump to `2.1.0`

tore-espressif (approved):
@Lzw655 thanks for the update, only 1 comment about versioning

bsp/esp32_s3_lcd_ev_board/idf_component.yml
Lzw655: The version of IDF has been updated to `>=5.0.1` since `5.0` doesn't support RGB triple-buffer.

Lzw655:
@espzav @tore-espressif Can we merge it now? Some customers have taken the new board, so it's a little urgent.

PR #263, opened 2023-12-22
doc: add ESP32-S3-USB-OTG instructions to power up USB Host (BSP-426)

georgik:
Documentation update.

Added instructions how to power up USB Host port on ESP32-S3-USB-OTG. Without described change the example does not read values from mouse or keyboard.


tore-espressif (approved):
Thanks @georgik !

PR #266, opened 2024-01-02
Add RGB support for esp_lv_port  (BSP-428)

espressif2022:

- [ ] CI passing

# Change description
1: Delete lv_port from esp32_s3_lcd_ev_board
2: Add RGB flag for esp_lv_port

### DIFF
--- a/bsp/esp32_s3_lcd_ev_board/idf_component.yml
+++ b/bsp/esp32_s3_lcd_ev_board/idf_component.yml
@@ -1,4 +1,4 @@
-version: "2.1.0"
+version: "3.0.0"
 description: Board Support Package for ESP32-S3-LCD-EV-Board
 url: https://github.com/espressif/esp-bsp/tree/master/bsp/esp32_s3_lcd_ev_board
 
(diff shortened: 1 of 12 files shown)


espzav (commented):
@espressif2022 Thank you very much for this PR. I am sorry to late reaction. I read changes lot of times (not too deep), but I still haven't got time for test. I would like to add it to new LVGL port 2.0 (with LVGL9). 

I have only one comment now. I don't like the changes in main header file. I hoped that it every callbacks will be solved inside. I don't know, if is it possible, but I hope so. I will try to understand more later and I will try to use this draft for implementation into new LVGL port :-) 

Thank you very much!

espressif2022:
> @espressif2022 Thank you very much for this PR. I am sorry to late reaction. I read changes lot of times (not too deep), but I still haven't got time for test. I would like to add it to new LVGL port 2.0 (with LVGL9).
> 
> I have only one comment now. I don't like the changes in main header file. I hoped that it every callbacks will be solved inside. I don't know, if is it possible, but I hope so. I will try to understand more later and I will try to use this draft for implementation into new LVGL port :-)
> 
> Thank you very much!

I'll think about it. I haven't dealt with this recently, and maybe after the holiday.
The RGB flush callback was a bit tricky; we didn't integrate it at that time. I'll also take a look at the changes in the LVGL V9 flush callback.

PR #268, opened 2024-01-04
LVGL: Add support of LVGL9 into LVGL port

espzav:

- [x] CI passing

# Change description
Support of LVGL9 in ESP LVGL port (including support LVGL8 too).

# LVGL9 supported parts:
- [x] Automatic selection between LVGL8 and LVGL9
- [x] LCD support
- [x] Screen rotation
- [x] Monochromatic screen
- [x] LCD Touch
- [x] Navigation buttons
- [x] Encoder
- [x] USB HID device

# Other things not supported with LVGL9:
- [x] Images
~~- [ ] lv_obj_set_style_image_opa~~

# TODO (Create Jira tickets)
- [ ] Enable SW rotation in LVGL9
- [ ] Support more color models (now only 16bit color RGB565 is supported)
- [ ] apply rotation from display configuration - https://github.com/espressif/esp-bsp/pull/278


espzav:
@tore-espressif PTAL
All tested examples are here: https://github.com/espressif/esp-bsp/pull/280
I didn't test it more in LVGL8, as we decided not support LVGL8 in BSP examples. Some simple examples are supported in LVGL8.
The main example in component is fully supported in both versions of LVGL.

- [] I think, that I should update IDF examples too.

components/esp_lvgl_port/CMakeLists.txt
```diff
@@ -1,29 +1,61 @@
-file(GLOB_RECURSE IMAGE_SOURCES images/*.c)
 
-idf_component_register(SRCS "esp_lvgl_port.c" ${IMAGE_SOURCES} INCLUDE_DIRS "include" REQUIRES "esp_lcd" PRIV_REQUIRES "esp_timer")
+#Get LVGL version
+idf_component_get_property(lvgl_ver lvgl__lvgl COMPONENT_VERSION)
```
tore-espressif, line 3: To avoid problems in the future: 
We should check for both `lvgl__lvgl` and `lvgl` in case someone wants to use local copy of LVGL

components/esp_lvgl_port/include/esp_lvgl_port.h
```diff
@@ -1,5 +1,5 @@
 /*
```
tore-espressif, line 1: Please consider introducing `esp_lvgl_port_xxx.h` headers. It could make this file a little clearer.
However, initial version can stay like this

components/esp_lvgl_port/src/lvgl9/esp_lvgl_port_button.c
```diff
@@ -0,0 +1,199 @@
+/*
```
tore-espressif, line 1: Are there any siginificant changes in `esp_lvgl_port_xxx.c` files that need a proper review?
Are these files only copy-paster from the original `esp_lvgl_port`?

components/esp_lvgl_port/src/lvgl9/esp_lvgl_port.c
```diff
@@ -0,0 +1,198 @@
+/*
```
tore-espressif, line 1: esp_lvgl_port.c for LVGL8 and 9 is almost the same. Would it make sense to combine some features into common file?

It would help us track bugfixes in the future a lot

tore-espressif (approved):
@espzav great job! I left few comments that might be worth considering

components/esp_lvgl_port/include/esp_lvgl_port.h
```diff
@@ -78,6 +91,7 @@ typedef struct {
         unsigned int buff_dma: 1;    /*!< Allocated LVGL buffer will be DMA capable */
         unsigned int buff_spiram: 1; /*!< Allocated LVGL buffer will be in PSRAM */
         unsigned int sw_rotate: 1;   /*!< Use software rotation (slower) */
+        unsigned int swap_bytes: 1;  /*!< Swap bytes in RGB656 (16-bit) before send to LCD driver */
```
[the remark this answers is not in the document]
tore-espressif: Yep, I cherry-picked your commit 3b41b9f858811466c0257ff2d1734b4a8d5eebb5 and had to set swap_bytes to false

components/esp_lvgl_port/src/lvgl9/esp_lvgl_port.c
espzav: Now, the files are same. But in future, in LVGL9 will be more optimization, task sleeping and similar...

components/esp_lvgl_port/include/esp_lvgl_port.h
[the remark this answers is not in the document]
espzav: This issue is more for PR #280

espzav:
@tore-espressif Thank you for review. Fixed all except mail `esp_lvgl_port.c` file (as a comment). 
- Divided header files too - PTAL

components/esp_lvgl_port/CMakeLists.txt
```diff
@@ -1,29 +1,64 @@
...
+if(lvgl_ver EQUAL "")
+    idf_component_get_property(lvgl_ver lvgl COMPONENT_VERSION)
+endif()
+message(STATUS "LVGL version: ${lvgl_ver}")
+
+#Select folder by LVGL version
+if(lvgl_ver LESS "9.0.0")
```
igrr, line 10: Should be `VERSION_LESS`, I think. Otherwise when the version 10.0.0 comes out, it will evaluate as `LESS` than 9.0.0 (lexicographically).

espzav: @igrr Thank you for the warning :-) I fixed it.

espzav:
My last commit fixing usage of FPS monitor. I don't know, if it is right solution or workaround - I am discussing with Gabor:  https://github.com/lvgl/lvgl/issues/4926#issuecomment-1911631548

Lzw655:
Hi @espzav, I also tried to port LVGL v9, and I've encountered some issues when using multiple CPU cores to render in parallel, and I'd like to share them with you:

- We cannot directly use the FreeRTOS from LVGL because the header file paths it contains are not suitable for ESP-IDF. Moreover, the use of `taskENTER_CRITICAL()` is not compatible. However, with some simple modifications, it can be adapted for use. I recommend getting official support for our system from LVGL or enabling `LV_OS_CUSTOM` and registering it in `esp_lvgl_port`.
- Ensure that the `lv_init()` function and `lv_timer_handler()` are in the same task; otherwise, the program will hang.

BTW, it indeed improves the performance by using multiple CPU cores. By running the v8 benchmark demo, the use of 2-unit increased the weighted FPS from 51 to 73 compared to a 1-unit.

espzav:
> My last commit fixing usage of FPS monitor. I don't know, if it is right solution or workaround - I am discussing with Gabor: [lvgl/lvgl#4926 (comment)](https://github.com/lvgl/lvgl/issues/4926#issuecomment-1911631548)

Here is the fix in LVGL: https://github.com/lvgl/lvgl/pull/5518

espzav:
@Lzw655 Thank you for info, I will try it.

components/esp_lvgl_port/include/esp_lvgl_port.h
```diff
@@ -12,28 +12,15 @@
...
-#define ESP_LVGL_PORT_USB_HOST_HID_COMPONENT 1
+#include "esp_lvgl_port_disp.h"
+#include "esp_lvgl_port_touch.h"
+#include "esp_lvgl_port_knob.h"
+#include "esp_lvgl_port_button.h"
+#include "esp_lvgl_port_usbhid.h"
+
+#if LVGL_VERSION_MAJOR == 8
+#include "esp_lvgl_port_compatibility.h"
 #endif
```
tore-espressif, line 24: This must be added to all `esp_lvgl_port_*.h` header files that use the new `lv_display_t` type. (eg. `esp_lvgl_port_disp.h`...)

tore-espressif (commented):
@espzav few more details :)

components/esp_lvgl_port/include/esp_lvgl_port.h
```diff
@@ -51,95 +38,14 @@ typedef struct {
...
-
 /**
  * @brief LVGL port configuration structure
  *
  */
 #define ESP_LVGL_PORT_INIT_CONFIG() \
     {                               \
         .task_priority = 4,       \
-        .task_stack = 4096,       \
+        .task_stack = 5120,       \
```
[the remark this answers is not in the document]
espzav: Increased.

components/esp_lvgl_port/CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+# Changelog
+
+## 1.5.0
```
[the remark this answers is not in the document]
espzav: Added big changes by GIT log.

espzav:
@tore-espressif Thank you for your review! All fixed. Is it ok, can we merge it now?

tore-espressif:
> @tore-espressif Thank you for your review! All fixed. Is it ok, can we merge it now?

I can see that the CI is still using LVGL 8 https://github.com/espressif/esp-bsp/actions/runs/7707868489/job/21005817735?pr=268#step:4:186

Could you please add `override_path` for `esp_lvgl_port` to all BSPs? Eg here https://github.com/espressif/esp-bsp/blob/master/bsp/esp32_s3_korvo_2/idf_component.yml#L29

This way, the CI and local builds will automatically pick `esp_lvgl_port` from the current state of the repo, instead the latest released version.

After the CI passes with LVGL9 build, we should be good to merge :)

tore-espressif:
@espzav Will we close this PR in favor of #280 ?

PR #272, opened 2024-01-09
Update build-examples-gh-pages-on-push.yml

tore-espressif:

- [x] CI passing

# Change description
With fixed IDF version that we use for publishing example binaries, we are also tight to fixed version of idf-component-manager. Some of our BSPs (esp32_s3_lcd_ev_board) need v1.4 of idf-component-manager here we update it explicitely


espzav (approved):
@tore-espressif Thank you for fix this issue! LGTM

PR #273, opened 2024-01-11
lcd: SSD1306 driver marked as deprecated.

espzav:

- [ ] CI passing


espzav:
@tore-espressif PTAL

tore-espressif:
@espzav Could you also please add `__attribute__((deprecated))` to this function https://github.com/espressif/esp-bsp/blob/1452b261c778453c32a98fe897b571561db95bb5/components/ssd1306/include/ssd1306.h#L53 ?

This way, users that use the latest version of ssd1306 component in their projects, will get deprecation warning during build

What do you think?

espzav:
> This way, users that use the latest version of ssd1306 component in their projects, will get deprecation warning during build
> 
> What do you think?

@tore-espressif Good point, Thank you. Updated.

tore-espressif (approved):
LGTM!

PR #275, opened 2024-01-12
esp_lvgl_port: Fix bad condition in port deinit (BSP-440)

cdollar393:

- [x] CI passing

# Change description
The `lvgl_port_deinit()` method appears to have an incorrect if-condition when checking if the LVGL task is running. Currently the code essentially says `if (!running) running = false;`, but it should be `if (running) running = false;`. The incorrect condition causes the LVGL task to never be stopped.


espzav (commented):
@cdollar393 Thank you for fix this issue. Please increase the component version for merge this PR.

espzav (approved):
LGTM

PR #289, opened 2024-02-01
fix bsp/display.h

tore-espressif:
* Added backlight functions to `bsp/display.h` so users can use them without including the main `bsp/esp-bsp.h`. This is important for future BSP version without LVGL
* Added capabilities of S3_eye board
* Add support for S3-EYE BOOT button

No need to release now

### DIFF
--- a/bsp/esp-box-3/include/bsp/display.h
+++ b/bsp/esp-box-3/include/bsp/display.h
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2023 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2023-2024 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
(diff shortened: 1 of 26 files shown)


bsp/esp32_s2_kaluga_kit/include/bsp/display_backlight.h
```diff
@@ -0,0 +1,28 @@
...
+#pragma once
+#include "esp_err.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* Backlight functions are not implemented - Kaluga board doesn't provide backlight control
+   These functions are here to provide consistent API with other Board Support Packages */
+esp_err_t bsp_display_backlight_on(void);
```
espzav, line 23: Can we add here  `esp_err_t bsp_display_brightness_set(int brightness_percent);` only for compatibility?

espzav (commented):
@tore-espressif Thank you for these changes. It is good idea to move these functions to another header, but why is it inside `display.h`? Do we really need new header file?

bsp/esp32_s2_kaluga_kit/include/bsp/display_backlight.h
tore-espressif: Done, thanks!

tore-espressif:
@espzav Thank you for the review! At first I wanted the `bsp/display.h` to be 'semi-private'. So only users who don't want to use LVGL would include it. But now there are more public definitions so the `bsp/display_backlight.h` does not make sense anymore.

Removed!

espzav (approved):
LGTM

PR #290, opened 2024-02-02
feat: No graphical library in BSPs

tore-espressif:

- [x] Add documentation for this feature
- [ ] Extend to all BSPs: Will be done in a follow up PR
- [ ] Add a CI check that build the `noglib` configurations: Will be done in a follow up PR
- [x] CI passing

# Change description
With this change we can upload No Graphical Library (-noglib) versions of the BSPs. We must remove `esp_lvgl_port` from the list of dependencies and disable the relevant parts of code.

As an example only ESP32-S3-EYE is modified for this change

### DIFF
--- a/BSP_development_guide.md
+++ b/BSP_development_guide.md
@@ -48,28 +48,35 @@ esp_err_t bsp_i2c_deinit(void);
 * See any BSP
 
 #### Display
-* BSPs with display are shipped with [LVGL](https://components.espressif.com/components/lvgl/lvgl) component
-* Display functions are divided into two parts. The first part is only HW initialization (include\bsp\display.h) and second part is initialization of the LVGL (BSP header).
-* Functions, which must be defined in BSP *.c file:
+* BSPs with display are shipped with [LVGL](https://components.espressif.com/components/lvgl/lvgl) component by default. For version without LVGL, you can use BSP version with `noglib` suffix (eg. `esp32_s3_eye_noglib`)
+* Display functions are divided into two parts. The first part is HW initialization only (include/bsp/display.h), second part is initialization of the display and LVGL (bsp/esp-bsp.h)
+* HW initialization functions, which must be defined in bsp/display.h file:
 ``` c
 esp_err_t bsp_display_new(const bsp_display_config_t *config, esp_lcd_panel_handle_t *ret_panel, esp_lcd_panel_io_handle_t *ret_io);
+esp_err_t bsp_display_brightness_set(int brightness_percent);
+esp_err_t bsp_display_backlight_on(void);
+esp_err_t bsp_display_backlight_off(void);
+```
+* LVGL init function, which must be defined in bsp/esp-bsp.h file:
+``` c
 lv_disp_t *bsp_display_start(void);
 lv_disp_t *bsp_display_start_with_config(const bsp_display_cfg_t *cfg);
 bool bsp_display_lock(uint32_t timeout_ms);
 void bsp_display_unlock(void);
-esp_err_t bsp_display_brightness_set(int brightness_percent);
-esp_err_t bsp_display_backlight_on(void);
-esp_err_t bsp_display_backlight_off(void);
 void bsp_display_rotate(lv_disp_t *disp, lv_disp_rot_t rotation);
 ```
 * See esp_wrover_kit, esp32_s2_kaluga_kit ...
 
 #### LCD Touch
 * LCD touch drivers are build around base [esp_lcd_touch](https://components.espressif.com/components/espressif/esp_lcd_touch) component
-* LCD touch functions are divided into two parts. The first part is only HW initialization (include\bsp\touch.h) and second part is initialization of the LVGL (BSP header).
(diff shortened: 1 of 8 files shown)


tore-espressif:
@espzav Please have a brief look. We can discuss tomorrow :)

bsp/esp32_s3_eye/include/bsp/config.h
```diff
@@ -0,0 +1,16 @@
...
+
+#pragma once
+
+/**************************************************************************************************
+ * BSP configuration
+ **************************************************************************************************/
+// By default, this BSP is shipped with LVGL graphical library. Enabling this option will exclude it.
+// If you want to use BSP without LVGL, select BSP version with '-noglib' suffix
+#if !defined(BSP_CONFIG_NO_GRAPHIC_LIB) // Check if the symbol is not coming from compiler definitions (-D...)
+#define BSP_CONFIG_NO_GRAPHIC_LIB (0)
```
espzav, line 15: Can we add into KConfig?

BSP_development_guide.md
```diff
@@ -48,28 +48,35 @@ esp_err_t bsp_i2c_deinit(void);
 * See any BSP
 
 #### Display
-* BSPs with display are shipped with [LVGL](https://components.espressif.com/components/lvgl/lvgl) component
-* Display functions are divided into two parts. The first part is only HW initialization (include\bsp\display.h) and second part is initialization of the LVGL (BSP header).
-* Functions, which must be defined in BSP *.c file:
+* BSPs with display are shipped with [LVGL](https://components.espressif.com/components/lvgl/lvgl) component by default. For version without LVGL, you can use BSP version with `-noglib` suffix (eg. `2.1.0-noglib`)
```
espzav, line 51: How would it be distributed in component manager?

bsp/esp32_s3_eye/include/bsp/config.h
tore-espressif: There are two usage options:
1. The user selects `noglib` BSP version: Enabling graphic library does not make sense, the required `esp_lvgl_port` is not included in this verision of the BSP
2. The user selects standard BSP version: Selecting `BSP_CONFIG_NO_GRAPHICAL_LIB` does not make too much sense. The `esp_lvgl_port` is already present, only the BSP graphical functions are not available

So IMO, having this option hidden is better. Do you see a usecase where the user might want to change this?

BSP_development_guide.md
tore-espressif: As discussed: The distribution will use suffix to BSP _name_.

So there will be derived component (linked to the original) with name eg `esp32_s3_eye_noglib`

bsp/esp32_s3_eye/include/bsp/config.h
espzav: You are right, we don't need it.

tore-espressif:
@espzav This PR is ready for review. The change is only implemented for `esp32_s3_eye` that will be internally tested with many examples. After the testing period I can follow-up with updates to other BSPs

Please let me know which PR you want to merge first :)

bsp/esp32_s3_eye/include/bsp/config.h
```diff
@@ -0,0 +1,16 @@
+/*
+ * SPDX-FileCopyrightText: 2024 Espressif Systems (Shanghai) CO LTD
```
espzav, line 2: Wouldn't be better to move this file to private?

BSP_development_guide.md
```diff
@@ -48,28 +48,35 @@ esp_err_t bsp_i2c_deinit(void);
...
-esp_err_t bsp_display_backlight_off(void);
 void bsp_display_rotate(lv_disp_t *disp, lv_disp_rot_t rotation);
 ```
 * See esp_wrover_kit, esp32_s2_kaluga_kit ...
 
 #### LCD Touch
 * LCD touch drivers are build around base [esp_lcd_touch](https://components.espressif.com/components/espressif/esp_lcd_touch) component
-* LCD touch functions are divided into two parts. The first part is only HW initialization (include\bsp\touch.h) and second part is initialization of the LVGL (BSP header).
-* Functions, which must be defined in BSP *.c file:
+* BSPs with touch display are shipped with [LVGL](https://components.espressif.com/components/lvgl/lvgl) component by default. or version without LVGL, you can use BSP version with `noglib` suffix (eg. `esp32_s3_eye_noglib`)
```
espzav, line 72:
```suggestion
* BSPs with touch display are shipped with [LVGL](https://components.espressif.com/components/lvgl/lvgl) component by default. For version without LVGL, you can use BSP version with `noglib` suffix (eg. `esp32_s3_eye_noglib`)
```

espzav (approved):
@tore-espressif Thank you for this PR, good work! LGTM

bsp/esp32_s3_eye/include/bsp/config.h
tore-espressif: We include this file from public `esp32_s3_eye.h` so it must be public

PR #293, opened 2024-02-08
Bsp/default lvgl version + runtime buffer size set

espzav:

- [ ] CI passing

# Change description
- Set default LVGL version in all BSPs to LVGL8
- Support setting display buffer size and other display settings from runtime initialization in BSP


espzav:
Support setting display buffer size and other display settings from runtime initialization in BSP
- @tore-espressif I made proposal in ESP-BOX, if it will be ok, I will copy to all BSPs - PTAL

tore-espressif:
@espzav The proposal LGTM. Do you want add this change for all BSPs in this PR?

espzav:
> @espzav The proposal LGTM. Do you want add this change for all BSPs in this PR?

Yes, I will do same changes in all BSPs in this PR. Should I allow to make change from Kconfig in all boards too, or only runtime? I leave here, because backward compatibility, but it is not supported in all boards now.

espzav:
@tore-espressif PTAL

tore-espressif (approved):
LGTM!

PR #294, opened 2024-02-13
fix(ci): Separate standard and noglib upload jobs

tore-espressif:


espzav (approved):
@tore-espressif Thank you for this fix! LGTM

PR #296, opened 2024-02-16
fix(ci): Use fixed version of idf-component-manager

tore-espressif:
This PR fixes errors with latest idf-component-manager

@kumekay The latest idf-component-manager v1.5.1 seems to incorrectly resolve conditional dependencies:
Here's the error: https://github.com/espressif/esp-bsp/actions/runs/7917759550/job/21614583839?pr=282#step:4:108
Here's the original idf_component.yml file: https://github.com/espressif/esp-bsp/blob/master/bsp/esp32_s3_lcd_ev_board/idf_component.yml
Could you please have a look too?

cc @espzav @pborcin


espzav (approved):
@tore-espressif Thank you! LGTM

PR #298, opened 2024-02-21
fix(ci): Extend noglib script to all BSPs

tore-espressif:
# Change description
Extend noglib python script so it 
* accepts multiple BSPs
* renames the BSP automatically
* check the dependency table in readme automatically

If you want to add a new noglib BSP, only 2 changes are needed:
1. Add the BSP to this space separated list (eventually, this will include all BSPs) https://github.com/espressif/esp-bsp/blob/c1fd14695e295ce121a734345703dc0ec93e5d42/.github/workflows/upload_component_noglib.yml#L19
2. Add the noglib BSP to upload job https://github.com/espressif/esp-bsp/blob/c1fd14695e295ce121a734345703dc0ec93e5d42/.github/workflows/upload_component_noglib.yml#L25

### DIFF
--- a/.github/ci/bsp_noglib.py
+++ b/.github/ci/bsp_noglib.py
@@ -8,8 +8,10 @@ Remove esp_lvgl_port from BSP's dependencies
 """
 
 import sys
+import argparse
 from pathlib import Path
 from idf_component_tools.manifest import ManifestManager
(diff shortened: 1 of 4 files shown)


espzav (approved):
@tore-espressif Thank you for fast update! LGTM

PR #306, opened 2024-03-06
feat(LVGL port): Support for LVGL9 and support for MIPI-DSI screen.

espzav:

- [x] CI passing

# Change description
- Added LVGL9 support
- Added MIPI-DSI support
- Divided into files per feature


espzav:
@tore-espressif PTAL

components/esp_lvgl_port/examples/i2c_oled/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+cmake_minimum_required(VERSION 3.16)
```
tore-espressif, line 1: I checked the CI and it seems that our `esp_lvgl_port/examples` applications are not built.

Could you please add a CI step to build them? Ideally using idf-build-apps

components/esp_lvgl_port/include/esp_lvgl_port_knob.h
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+#pragma once
+
+#include "esp_err.h"
+#include "lvgl.h"
+
+#if __has_include ("iot_knob.h")
+#include "iot_knob.h"
+#include "iot_button.h"
```
tore-espressif, line 19: When I tried adding `knob` without `button` component the compilation failed here.

Knob does not depend on button in the original component. Could we remove the dependency here? (Or at least not fail in this case?)

tore-espressif (commented):
A few nitpicks :)

components/esp_lvgl_port/include/esp_lvgl_port_knob.h
espzav: We need `button` component for this file: `esp_lvgl_port_knob.c`. When we add knob, there is code for handle click by `button` component.

tore-espressif: I see, could we do something like
``` c
#if __has_include ("iot_knob.h")
#if !__has_include("iot_button.h")
#error LVLG Knob requires button component. Please add it with idf.py add-dependency espressif/button
#endif
#include "iot_knob.h"
#include "iot_button.h"

tore-espressif:
@espzav CI changes LGTM! thank you!

espzav:
@tore-espressif Thank you for review. Is it ok now? Can we merge it?

tore-espressif:
> @tore-espressif Thank you for review. Is it ok now? Can we merge it?
yes, go ahead!

PR #307, opened 2024-03-07
Descriptions and tags updates (BSP-472)

VojtechBartoska:
# Change description
Updating BSPs descriptions to fix searching BSP packages on https://components.espressif.com/

Also adding a tag "bsp".

### DIFF
--- a/bsp/esp_bsp_generic/idf_component.yml
+++ b/bsp/esp_bsp_generic/idf_component.yml
@@ -1,7 +1,10 @@
 
 version: "1.1.2"
-description: Generig Board Support Package
+description: Generig Board Support Package (BSP)
 url: https://github.com/espressif/esp-bsp/tree/master/bsp/esp_bsp_generic
+
+tags:
+  - bsp
 
 dependencies:
   idf: ">=4.4.2"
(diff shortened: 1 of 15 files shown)


espzav (approved):
@VojtechBartoska Thank you for update. LGTM
Only one comment: This not update component registry till we increase component version.

VojtechBartoska:
@espzav I see, so should I update the versions as well?

espzav:
> @espzav I see, so should I update the versions as well?

I think, we can release all BSPs after changes for LVGL9. @tore-espressif what do you think?

tore-espressif:
> I think, we can release all BSPs after changes for LVGL9. @tore-espressif what do you think?

Yep, let's not release now :)

tore-espressif (approved):

bsp/esp_bsp_generic/idf_component.yml
```diff
@@ -1,8 +1,11 @@
 
 version: "1.1.2"
-description: Generig Board Support Package
+description: Generig Board Support Package (BSP)
```
kumekay, line 3:
```suggestion
description: Generic Board Support Package (BSP)
```

kumekay (approved):
1 tiny typo, otherwise LGTM

PR #313, opened 2024-03-22
feat(LVGL port): Support for RGB screen. (BSP-478)

espressif2022:

- [ ] CI passing

# Change description
1: esp_lv_port adds RGB interface support.
2: esp32_s3_lcd_ev_board delete lvgl related content.


components/esp_lvgl_port/include/esp_lvgl_port_disp.h
```diff
@@ -44,20 +44,27 @@ typedef struct {
     uint32_t    buffer_size;    /*!< Size of the buffer for the screen in pixels */
     bool        double_buffer;  /*!< True, if should be allocated two buffers */
     uint32_t    trans_size;     /*!< Allocated buffer will be in SRAM to move framebuf */
+    void        *user_buf1;     /*!< OPTIONAL: A buffer to be used by LVGL to draw the image */
+    void        *user_buf2;     /*!< OPTIONAL: Optionally specify a second buffer to render and flush */
```
espzav, line 48: 1. It seems, that we still add new items in this structure. Here is lot of buffers and I think, that it is not clear for custommers. We should add new union for specific settings for MIPI, RGB, monochromatic, ... 
2. Same for flags. 
3. I would like to add something into esp_lcd, how to determine, what bus is used (MIPI, RGB, ... )

components/esp_lvgl_port/include/esp_lvgl_port.h
```diff
@@ -120,6 +120,8 @@ esp_err_t lvgl_port_stop(void);
  */
 esp_err_t lvgl_port_resume(void);
 
+bool lvgl_port_task_notify(void);
```
espzav, line 123: For what is this function? Any description?

espzav (commented):
@espressif2022 thank you for this PR. I know this is the draft and I didn't make real review yet. I left only some thoughts.

components/esp_lvgl_port/include/esp_lvgl_port.h
espressif2022: This is used to synchronize `RGB` screen refresh and `LVGL tasks`.

If it is an `SPI` screen, the `trans_done` event will not be received if `esp_lcd_panel_draw_bitmap` is not executed. 
This is synchronous, `esp_lcd_panel_draw_bitmap` , and receive `flush_io_ready_callback` event. 
We can handle `lv_disp_flush_ready` in the `trans_done` event.

However, the `esp_lcd_panel_draw_bitmap` and `trans_done` of the `RGB` screen are not generated synchronously. 
The screen is always refreshed and there is always a `trans_done`  event。
So we can only wait for the first event after `draw_bitmap` in `lvgl_port_flush_callback`.

components/esp_lvgl_port/include/esp_lvgl_port_disp.h
espressif2022: > 1. It seems, that we still add new items in this structure. Here is lot of buffers and I think, that it is not clear for custommers. We should add new union for specific settings for MIPI, RGB, monochromatic, ...
> 2. Same for flags.
> 3. I would like to add something into esp_lcd, how to determine, what bus is used (MIPI, RGB, ... )

Yes,  the fbs of SPI, I80 are allocated by esp_lv_port, not in the driver.
But the RGB driver will allocate fb internally, which depends on the user, and the internal buffer is more efficient.

The screen is initialized in BSP, at this time, the fb inside the RGB driver has been allocated. 
It is not appropriate to take it completely from inside the RGB, users may want to use thire own buffer, so it is a problem.

espzav:
It was made by PRs: #322 and #333.

PR #314, opened 2024-03-26
change(mipi_dsi): use public esp_ldo regulator driver

suda-morris:
# Change description

https://docs.espressif.com/projects/esp-idf/en/latest/esp32p4/api-reference/peripherals/ldo_regulator.html

- [x] Use `esp_ldo_regulator.h` API for configuring the DPHY power supplying. The legacy `esp_private/esp_ldo.h` will be removed from esp-idf soon.
- [x] Updated the dependency as well, because mipi dsi driver is only available since esp-idf v5.3
- [x] Disabled the build test for esp32c5, because recently esp-idf master just switch to esp32c5-mp version, where SPI/I2C/UART/GPIO drivers are not ready yet. So let's temporarily removed the esp32-c5 build test.


bsp/esp32_p4_function_ev_board/idf_component.yml
```diff
@@ -1,4 +1,4 @@
-version: "1.0.0"
+version: "1.0.1"
```
suda-morris, line 1: better to use "1.1.0" in this case

espzav:
@suda-morris thank you for this PR. I see, that this changes are public in IDF. Is it still as draft? And one missing header. Do you plan to fix this PR or should I make it in another PR?

suda-morris:
> @suda-morris thank you for this PR. I see, that this changes are public in IDF. Is it still as draft? And one missing header. Do you plan to fix this PR or should I make it in another PR?

Thanks @espzav Yes, it's ready for review now. I've replaced the legacy ldo driver with esp_ldo_regulator.h

tore-espressif (approved):
Thank you @suda-morris LGTM!

espzav (approved):
@suda-morris Thank you for this PR. LGTM

PR #334, opened 2024-05-28
ci: Update IDF version for tests.

espzav:

- [ ] CI passing

# Change description
Updated IDF versions in CI


espzav:
@tore-espressif Fixed BSP-503 here https://github.com/espressif/esp-bsp/pull/334/files#diff-77e3b740a806414c827dd5b3ef77e06f120a2c607e76c59f517654aca79f2dd4 PTAL

.build-test-rules.yml
```diff
@@ -5,10 +5,16 @@ examples:
...
+  disable:
+    - if: IDF_VERSION_MAJOR <= 5.0.4 || IDF_VERSION_MAJOR == 5.1.1
+      reason: Supported from version 5.0.5
+
 components/lcd/esp_lcd_gc9503/test_apps:
   disable:
-    - if: IDF_VERSION_MAJOR < 5
-      reason: Component is supported only for IDF >= 5.0
+    - if: IDF_VERSION_MAJOR <= 5.0.4 || IDF_VERSION_MAJOR == 5.1.1
+      reason: Supported from version 5.0.5
```
pborcin, line 16: Does this work? Shouldn't it be just something like IDF_VERSION or IDF_VERSION_VAL?

.github/ci/update_readme_dependencies.py
```diff
@@ -103,6 +103,13 @@ def get_capabilities_table(header_path, manifest):
         return table_data
 
 
+def index_regexp(yourlist, yourstring):
+    ide = [i for i, item in enumerate(yourlist) if re.search(yourstring, item)]
+    if len(ide) > 0:
+        return ide[0]
+    return -1
+
```
pborcin, line 111: just a suggestion, not necessary to change
```suggestion
def index_regexp(yourlist, yourstring):
    for i, line in enumerate(yourlist):
        if yourstring in line:
            return i
    return -1
```

pborcin (approved):
LGTM

.github/ci/update_readme_dependencies.py
espzav: Thank you for suggestion, it looks good :-) Changed.

.build-test-rules.yml
espzav: Good point, I overlooked it. Fixed, let's wait for CI pass.

espzav:
@pborcin Thank you for review!

Indastri:
LGTM!

espzav:
@tore-espressif PTAL

.github/workflows/build_example.yml
```diff
@@ -27,6 +27,15 @@ jobs:
           - idf_ver: "release-v4.4"
             parallel_count: 1
             parallel_index: 1
+          - idf_ver: "release-v5.1"
+            parallel_count: 1
+            parallel_index: 1
```
tore-espressif, line 32: With parallel_count=1 this job will probably be very slow. We can merge it as it is now, though

PR #353, opened 2024-07-26
icm42670: Support I2C NG driver

espzav:

- [x] CI passing

# Change description
The first PR, how to support I2C NG driver in BSP components.

### DIFF
--- a/components/icm42670/include/icm42670.h
+++ b/components/icm42670/include/icm42670.h
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2023 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2023-2024 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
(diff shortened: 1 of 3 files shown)


espzav:
@tore-espressif PTAL

components/icm42670/include/icm42670.h
```diff
@@ -96,14 +100,18 @@ typedef void *icm42670_handle_t;
...
  *
- * @param port I2C port number
+ * @param i2c_bus I2C bus
  * @param dev_addr I2C device address of sensor
  *
  * @return
  *     - NULL Fail
  *     - Others Success
  */
+#if (ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 3, 0))
```
tore-espressif, line 110: I think this rule might be too limiting. The new I2C Driver-NG is available from IDFv5.2 In practice, there can be other scenarios:

1. IDF v5.2 + Driver-NG
2. IDF v5.3 (or later) + Legacy driver

The only rule is that the driver-ng and the legacy driver cannot be linked in the same application.

So I can see 2 options:
1. Maintain two versions of this component 1.x: legacy driver, 2.x: Driver-NG
2. Make this configurable through Menuconfig

ATM, option 2 seems better to me because it is aligned with linear versioning of our components and option 1 is not scalable to BSP components

pborcin (approved):
LGTM

components/icm42670/include/icm42670.h
tore-espressif: We agreed that we will make 'big leap' and support only Driver-NG in new major versions

PR #356, opened 2024-07-30
bsp(ESP32-P4 EV): Update BSP for HW v1.4 + LCD EK79007 1024x600

espzav:

- [x] CI passing

# Change description
- Update BSP for HW v1.4
- Add support for LCD EK79007 1024x600 (selectable in menuconfig)
- Allow to change color mode in menucinfig (RGB565/RGB888)


bsp/esp32_p4_function_ev_board/Kconfig
```diff
@@ -75,6 +75,31 @@ menu "Board Support Package(ESP32-P4)"
         help
             LEDC channel is used to generate PWM signal that controls display brightness.
             Set LEDC index that should be used.
+            
+        choice BSP_LCD_COLOR_FORMAT
+            prompt "Select LCD color format"
+            default BSP_LCD_COLOR_FORMAT_RGB565
```
tore-espressif, line 81: Q: Do both displays support both RGB565 and RGB888?

tore-espressif (approved):
Thank you very much for the update! I left only 1 question

bsp/esp32_p4_function_ev_board/esp32_p4_function_ev_board.c
```diff
@@ -315,16 +362,21 @@ esp_err_t bsp_touch_new(const bsp_touch_config_t *config, esp_lcd_touch_handle_t
...
         .levels = {
             .reset = 0,
             .interrupt = 0,
         },
         .flags = {
             .swap_xy = 0,
+#if CONFIG_BSP_LCD_TYPE_1024_600
+            .mirror_x = 1,
+            .mirror_y = 1,
+#else
```
Lzw655, line 376: It doesn't have to mirror the LCD because we usually position it vertically with the USB port facing upward to facilitate programming.  

What's more, it does not support mirroring through commands. You can see a warning message on the terminal. Instead, we should use IOs to control the 'UP/DN' and 'SHLR' signals in the LCD sub-board.

image

bsp/esp32_p4_function_ev_board/esp32_p4_function_ev_board.c
```diff
@@ -348,15 +400,20 @@ static lv_display_t *bsp_display_lcd_init(const bsp_display_cfg_t *cfg)
...
         .rotation = {
             .swap_xy = false,
             .mirror_x = true,
             .mirror_y = true,
         },
+#if CONFIG_BSP_LCD_COLOR_FORMAT_RGB888
+        .color_format = LV_COLOR_FORMAT_RGB888,
+#else
+        .color_format = LV_COLOR_FORMAT_RGB565,
+#endif
```
Lzw655, line 416:
```suggestion
#if LVGL_VERSION_MAJOR >= 9
#if CONFIG_BSP_LCD_COLOR_FORMAT_RGB888
        .color_format = LV_COLOR_FORMAT_RGB888,
#else
        .color_format = LV_COLOR_FORMAT_RGB565,
#endif
#endif
```
These cannot be used when using LVGL v8.

bsp/esp32_p4_function_ev_board/esp32_p4_function_ev_board.c
```diff
@@ -393,7 +451,11 @@ lv_display_t *bsp_display_start(void)
         .buffer_size = BSP_LCD_DRAW_BUFF_SIZE,
         .double_buffer = BSP_LCD_DRAW_BUFF_DOUBLE,
         .flags = {
+#if CONFIG_BSP_LCD_COLOR_FORMAT_RGB888
+            .buff_dma = false,
+#else
             .buff_dma = true,
+#endif
             .buff_spiram = false,
```
Lzw655, line 459:
```suggestion
            .buff_dma = false,
            .buff_spiram = true,
```
I tested that using PSARM directly here would also provide good performance. And we can increase `BSP_LCD_DRAW_BUFF_SIZE` to `BSP_LCD_H_RES * BSP_LCD_V_RES`. You can try it with the [esp-ui example](https://gitlab.espressif.cn:6688/ae_group/esp-ui/-/tree/feat/add_example_p4_func_ev_board/examples/esp_idf/esp32_p4_function_ev_board?ref_type=heads).

Lzw655 (approved):
Hi @espzav, thank you for the prompt support. I tested the display function, and it performed well. Could you also integrate support for the audio function?

bsp/esp32_p4_function_ev_board/esp32_p4_function_ev_board.c
espzav: I know, that screen cannot use standard rotation, this is only for right rotation of the touch.

bsp/esp32_p4_function_ev_board/Kconfig
espzav: yes

bsp/esp32_p4_function_ev_board/esp32_p4_function_ev_board.c
espzav: > What's more, it does not support mirroring through commands. You can see a warning message on the terminal. Instead, we should use IOs to control the 'UP/DN' and 'SHLR' signals in the LCD sub-board.

We need any new API for `LVGL port` for rotation this screen type. I will think about it later.

bsp/esp32_p4_function_ev_board/esp32_p4_function_ev_board.c
espzav: I tried benchmark demo:
1. Internal buffer, RGB565, DMA (BSP_LCD_H_RES * 100): 24 FPS (avg)
2. SPIRAM buffer, RGB565 (BSP_LCD_H_RES * 100): 18 FPS (avg)
3. SPIRAM buffer, RGB565 (BSP_LCD_H_RES * BSP_LCD_V_RES): 19 FPS (avg)

I think, we should keep the best performance settings. What do you think @tore-espressif ?

bsp/esp32_p4_function_ev_board/esp32_p4_function_ev_board.c
espzav: Fixed.

espzav:
@tore-espressif and @Lzw655 Thank you for review
About audio, I would like to support audio on this board. We are working on support new I2C NG driver, I think better will be support audio after that,

bsp/esp32_p4_function_ev_board/esp32_p4_function_ev_board.c
tore-espressif: I think that making the best performance default is a good idea. If someone needs to use different configuration, we provide `bsp_display_start_with_config()` function

espzav:
@tore-espressif @Lzw655 We have new discussion about considering initialization of ILI9881C and EK79007. I added new conditions into version for these components in `idf_component.yml`. We can merge it as it is and update it later or we can wait to considering it. 
What do you thing?

PR #357, opened 2024-07-31
feature(lvgl_port): Initial support for SIMD rendering in LVGL

peter-marcisovsky:

- [ ] CI passing

# Change description
This MR adds an initial support for SIMD rendering in LVGL using LVGL's internal Blend API

## This MR Implements:
- Assembly source files integrated into the `lvgl_port` component
- Integration into the LVGL blend API via LVGL's `kconfig`
   - `LV_DRAW_SW_ASM_CUSTOM` must be enabled
   - `LV_DRAW_SW_ASM_CUSTOM_INCLUDE` must be provided with a header file with function header for the assembly implemented functions, in our case `esp_lvgl_port_lv_blend.h`
   - This way, the `esp_lvgl_port_lv_blend.h` is under our control during the development phase, and will be included into the official LVGL repo once finalized, just like the existing SIMD header files from ARM.
- Assembly implementation of following primitive rendering functions:
   - Simple fill for `ARGB8888` color format, supported targets are `esp32s3` and `esp32`
   - Simple fill for `RGB565` color format, supported target is `esp32`
- Test app with functionality and benchmark tests


components/esp_lvgl_port/src/lvgl9/simd/lv_color_blend_to_argb8888_aes3.S
```diff
@@ -0,0 +1,298 @@
...
+#include "sdkconfig.h"
+
+#if CONFIG_IDF_TARGET_ESP32S3
+
+// This is LVGL ARGB8888 simple fill for ESP32S3 processor.
+    .text
+    .align  4
+    .global lv_color_blend_to_argb8888_esp32_aes3
+    .global .lv_color_blend_to_argb8888_esp32_ae32_body
+    .type   lv_color_blend_to_argb8888_esp32_aes3,@function
```
tore-espressif, line 16: What is the difference between `aes3` and `ae32` functions? What do the acronyms mean?

components/esp_lvgl_port/include/esp_lvgl_port_lv_blend.h
```diff
@@ -0,0 +1,131 @@
...
+ **********************/
+
+static const char *TAG_LV_BLEND = "LVGL_PORT_LV_BLEND";
+
+/**********************
+ * GLOBAL PROTOTYPES
+ **********************/
+
+// Extern variable to control switching between Assembly or C implementation, for testing purposes only
+extern bool LV_BLEND_USE_ASM;
```
tore-espressif, line 65: This is very dodgy. We declare a extern variable here but the actual definition is in a test_app.

In case we don't compile the test code, we don't have a definition of this variable!

components/esp_lvgl_port/include/esp_lvgl_port_lv_blend.h
tore-espressif, line 1: I'm not sure I understand how this file is supposed to be used. There aren't any public declarations that users are supposed to call, right? AFAIK, this should be only a plugin to LVGL's render API

components/esp_lvgl_port/include/esp_lvgl_port_lv_blend.h
```diff
@@ -0,0 +1,131 @@
...
+
+/*********************
+ *      INCLUDES
+ *********************/
+#include "esp_log.h"
+#include <sdkconfig.h>
+#include "lvgl.h"
+#include "lv_conf_internal.h"
+
+#if CONFIG_LV_DRAW_SW_ASM_CUSTOM
```
tore-espressif, line 21: In case someone includes this file and the ASM_CUSTOM option is not set, we should at least write a warning that the ASM blender is not actually used

components/esp_lvgl_port/include/esp_lvgl_port_lv_blend.h
```diff
@@ -0,0 +1,131 @@
...
+extern bool LV_BLEND_USE_ASM;
+
+extern int lv_color_blend_to_argb8888_esp32_aes3(asm_dsc_t *asm_dsc);         // ESP32S3 assembly implementation
+extern int lv_color_blend_to_argb8888_esp32_ae32(asm_dsc_t *asm_dsc);         // ESP32 assembly implementation
+
+static inline lv_result_t _lv_color_blend_to_argb8888_esp32(_lv_draw_sw_blend_fill_dsc_t *dsc)
+{
+    // Check if asm variant should be used (Only for testing)
+    if (!LV_BLEND_USE_ASM) {
+        ESP_LOGD(TAG_LV_BLEND, "Calling ANSI impl. of: Simple fill ARGB8888");
```
tore-espressif, line 74: Where is the ANSI implementation called?

components/esp_lvgl_port/test_apps/simd/main/test_lv_fill_functionality.c
```diff
@@ -0,0 +1,340 @@
...
+    - Initialize input parameters (test array length, width, allocate array...) of the benchmark test
+    - Prepare testing matrix, to cover all the possible combinations of destination array widths, lengths, memory alignment...
+    - Run assembly version of the LVGL blending API
+    - Run ANSI C version of the LVGL blending API
+    - Compare the results
+    - Repeat above 3 steps for each test matrix setup
+    - Free structures needed for LVGL blending API
+*/
+
+TEST_CASE_MULTIPLE_STAGES("LV Fill functionality", "[lv_fill][functionality]",
```
tore-espressif, line 87: Could you explain why multi-stage test is used here? From docs, it seems that we might not need this https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/unit-tests.html#multi-stage-test-cases

components/esp_lvgl_port/include/esp_lvgl_port_lv_blend.h
igrr: The way it works is that if the function behind the LV_DRAW_SW_COLOR_BLEND_TO_ARGB8888 macro returns LV_RESULT_INVALID, then the software implementation is going to be used.

https://github.com/lvgl/lvgl/blob/a798963ee89164ba2758c7fdb5fabea33e22a805/src/draw/sw/blend/lv_draw_sw_blend_to_argb8888.c#L235C33-L235C67

components/esp_lvgl_port/include/esp_lvgl_port_lv_blend.h
igrr: Seems like the header file name should be stored in LV_DRAW_SW_ASM_CUSTOM_INCLUDE option: https://github.com/lvgl/lvgl/blob/a798963ee89164ba2758c7fdb5fabea33e22a805/src/draw/sw/blend/lv_draw_sw_blend_to_argb8888.c#L26

https://github.com/lvgl/lvgl/blob/a798963ee89164ba2758c7fdb5fabea33e22a805/Kconfig#L328-L338

So for now we have to do this via sdkconfig.defaults, until https://github.com/lvgl/lvgl/issues/6010 is done.

tore-espressif: I tried that, but LVGL does not see esp_lvgl_port includes. esp_lvgl_port depends on lvgl, not the other way around, so here we have a circular dependency.

We can introduce some workarounds for the time being...

components/esp_lvgl_port/src/lvgl9/simd/lv_color_blend_to_argb8888_aes3.S
peter-marcisovsky: I used acronyms used in [esp-dsp](https://github.com/espressif/esp-dsp/blob/master/modules/matrix/mul/include/dspm_mult.h#L44)
`aes3` is assembly implementation for esp32s3
`ae32` is assembly implementation for esp32
`ansi` is a gneral C code

Do you think it would make more sense if we changed it somehow?

components/esp_lvgl_port/include/esp_lvgl_port_lv_blend.h
peter-marcisovsky: @tore-espressif  Yes it is not ideal, but I could not find any other way of having a control over switching between assembly and C code, without recompiling the whole project. @igrr  I was thinking about something similar, will take a look if we can implement like this.

components/esp_lvgl_port/include/esp_lvgl_port_lv_blend.h
peter-marcisovsky: You can not include this file using Kconfing without enabling the `CONFIG_LV_DRAW_SW_ASM_CUSTOM`, you can include it just manually. But sure will add a warning.

components/esp_lvgl_port/include/esp_lvgl_port_lv_blend.h
```diff
@@ -0,0 +1,131 @@
...
+
+/**********************
+ * GLOBAL PROTOTYPES
+ **********************/
+
+// Extern variable to control switching between Assembly or C implementation, for testing purposes only
+extern bool LV_BLEND_USE_ASM;
+
+extern int lv_color_blend_to_argb8888_esp32_aes3(asm_dsc_t *asm_dsc);         // ESP32S3 assembly implementation
+extern int lv_color_blend_to_argb8888_esp32_ae32(asm_dsc_t *asm_dsc);         // ESP32 assembly implementation
```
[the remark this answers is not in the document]
peter-marcisovsky: I was following the [esp-dsp](https://github.com/espressif/esp-dsp/blob/948115bdad08c110a467cb4f90b48ecefd9fe473/modules/matrix/mul/include/dspm_mult.h#L44) way of having separate name for separate target implementation.

But I think we can have just one function prototype here being `lv_color_blend_to_argb8888_esp32` as you are suggesting. And let the CMake to deal with the different targets.

components/esp_lvgl_port/include/esp_lvgl_port_lv_blend.h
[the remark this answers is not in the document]
peter-marcisovsky: Yes, this is only for development phase and will be removed from the final implementation. I found this debug output quite useful when developing and testing this feature.

components/esp_lvgl_port/src/lvgl9/simd/lv_color_blend_to_argb8888_ae32.S
```diff
@@ -0,0 +1,82 @@
...
+//     uint32_t dst_w;              l32i    8
+//     uint32_t dst_h;              l32i    12
+//     uint32_t dst_stride;         l32i    16
+//     const void * src_buf;        l32i    20
+//     uint32_t src_stride;         l32i    24
+//     const lv_opa_t * mask_buf;   l32i    28
+//     uint32_t mask_stride;        l32i    32
+// } asm_dsc_t;
+
+lv_color_blend_to_argb8888_esp32_ae32:
```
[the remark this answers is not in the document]
peter-marcisovsky: Correct, this function does not blend, just fills. I am just following the LVGL naming convention.

components/esp_lvgl_port/test_apps/simd/main/test_lv_fill_benchmark.c
```diff
@@ -0,0 +1,259 @@
...
+    - Count how many CPU cycles does it take to run a function from the LVGL blending API for each case (ideal and worst case)
+    - Repeat 4 previous steps with Assembly accelerated LVGL blending API
+    - Compare the counted cycles and evaluate the results
+    - Free test array
+    - Free structures needed for LVGL blending API
+*/
+
+TEST_CASE_MULTIPLE_STAGES("LV Fill benchmark", "[lv_fill][benchmark]",
+                          argb8888_benchmark,
+                          rgb565_benchmark)
```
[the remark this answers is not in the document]
peter-marcisovsky: Doesn't have to be multistage. Just wanted to group benchmark and functionality tests

components/esp_lvgl_port/test_apps/simd/main/lv_fill_common.c
```diff
@@ -0,0 +1,202 @@
...
+
+    return ESP_OK;
+}
+
+esp_err_t free_blend_params(void)
+{
+    // Free area
+    if (s_area != NULL) {
+        free(s_area);
+    }
```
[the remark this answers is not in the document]
peter-marcisovsky: Thx for suggesting.

components/esp_lvgl_port/test_apps/simd/main/test_lv_fill_functionality.c
peter-marcisovsky: I was not aware of this usage of multi-stage tests. Just wanted to group functionality and performance tests into 2 separate groups. Will correct it. Thx for the docs reference.

components/esp_lvgl_port/include/esp_lvgl_port_lv_blend.h
igrr: > I tried that, but LVGL does not see esp_lvgl_port includes. esp_lvgl_port depends on lvgl, not the other way around, so here we have a circular dependency.

Yes, we should set this explicitly, like this

```cmake
target_include_directories(idf::lvgl__lvgl PRIVATE path/to/asm/headers)
```

(need to be careful to find the real name of LVGL component, it might be either `lvgl` or `lvgl__lvgl` depending on whether it's overridden in the project, or not.)

components/esp_lvgl_port/src/lvgl9/simd/lv_color_blend_to_argb8888_aes3.S
igrr: I think the target names (`esp32`, `esp32s3`) could be more understandable to other developers looking at this code. We don't have to deal with the ANSI C implementation here, as it is already included into LVGL, so what remains is just to distinguish between the chips.

Another option could be to use some terms to describe the instruction sets, for example `xtensa_base`, `xtensa_pie`, `riscv_pie`, etc. But we might run into the limitations of this approach if we have more RISC-V chips with slightly different sets of PIE (processor instruction extensions).

components/esp_lvgl_port/CMakeLists.txt
```diff
@@ -22,12 +22,21 @@ endif()
...
+
 idf_component_register(
-        SRCS "${PORT_PATH}/esp_lvgl_port.c" "${PORT_PATH}/esp_lvgl_port_disp.c" 
+        SRCS "${PORT_PATH}/esp_lvgl_port.c" "${PORT_PATH}/esp_lvgl_port_disp.c" ${ASM_SOURCES}
         INCLUDE_DIRS "include" 
         PRIV_INCLUDE_DIRS "priv_include"
         REQUIRES "esp_lcd" 
-        PRIV_REQUIRES "esp_timer")
+        PRIV_REQUIRES "esp_timer"
+        WHOLE_ARCHIVE)
```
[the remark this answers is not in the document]
peter-marcisovsky: Without the `WHOLE_ARCHIVE` the project would not link. In order to remove the `WHOLE_ARCHVIE` we can for example use the same approach as in [esp-usb](https://github.com/espressif/esp-usb/blob/c50d3c8511d39bc0e00f92e6477b2cd35cf92215/host/class/uvc/usb_host_uvc/test_app/main/CMakeLists.txt#L24) to force link an .S file, which would require to force link each .S file, since there is only one function present in each .S file. And those functions do no call each other.

@igrr is there any other way of not not using `WHOLE_ARCHIVE` in  this case?

components/esp_lvgl_port/src/lvgl9/simd/lv_color_blend_to_argb8888_aes3.S
peter-marcisovsky: `*_esp32` and `*_esp32s3` would make more sense then. 

Also, `xtensa_base` would not really make sense here, since the esp32s2 does not have zero overhead loops ( `loopnez` opcode in assembly) if I am not mistaken. And  the `loopnez` is used in each assembly function here.

peter-marcisovsky:
Reworked test apps:
- Functionality test: 
   - uses hard copy of necessary lvgl source files, thus creating a local ANSI C version of the LVGL Blend API for functionality test comparison with the Assembly version of the API
   - Does not use neither the LVGL port, nor the LVGL managed component
   
- Benchmark test: 
    - 2 separate tests apps with different sdkconfig settings (Assembly render Enabled/Disabled) must be build for benchmark comparison
    - Uses LVGL port, as it gives a performance comparison of an overall Assembly render integration into the LVGL

components/esp_lvgl_port/include/esp_lvgl_port_lv_blend.h
peter-marcisovsky: Correction of the CMake modification, so the LVGL sees the esp_lvgl_port includes:

```
idf_component_get_property(lvgl_lib lvgl__lvgl COMPONENT_LIB)
target_include_directories(${lvgl_lib} PRIVATE include)
```

espzav (approved):
@peter-marcisovsky thank you for this PR. Good job! LGTM

components/esp_lvgl_port/test_apps/simd/benchmark/main/idf_component.yml
```diff
@@ -0,0 +1,6 @@
+## IDF Component Manager Manifest File
+dependencies:
+  idf: ">=4.4"
```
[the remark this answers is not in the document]
peter-marcisovsky: Removed the whole `idf_component.yml` as you are below, if only an IDF dependency is set.

components/esp_lvgl_port/test_apps/simd/functionality/main/Kconfig.projbuild
```diff
@@ -0,0 +1,5 @@
+# Creating CONFIG_LV_DRAW_SW_ASM_CUSTOM avaliable in lvgl Kconfig to enable assembler source files by deafult
+
+config LV_DRAW_SW_ASM_CUSTOM
```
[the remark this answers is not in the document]
peter-marcisovsky: `LV_DRAW_SW_ASM_CUSTOM` is a config present in LVGL and I am not using neither `lvgl_port` , nor `lvgl` components, since the test app uses a hard copy of LVGL functions as their ANSI implementations. So there is no way to set this config, since it does not exist in the config. (I hope it works like that)

But, I need to set the `LV_DRAW_SW_ASM_CUSTOM` to true, so the assembly function prototypes present in the `esp_lvgl_port_lv_blend.h` are accessible for the test app.

```
#if !CONFIG_LV_DRAW_SW_ASM_CUSTOM
#warning "esp_lvgl_port_lv_blend.h included, but CONFIG_LV_DRAW_SW_ASM_CUSTOM not set. Assembly rendering not used"
#else
```

For this, I created this `projbuild` and set `LV_DRAW_SW_ASM_CUSTOM` to always true.

components/esp_lvgl_port/test_apps/simd/functionality/main/idf_component.yml
```diff
@@ -0,0 +1,3 @@
+## IDF Component Manager Manifest File
+dependencies:
+  idf: ">=4.4"
```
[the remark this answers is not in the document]
peter-marcisovsky: Removed the whole file.

components/esp_lvgl_port/test_apps/simd/benchmark/main/test_app_main.c
```diff
@@ -0,0 +1,64 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <stdio.h>
+#include "unity.h"
+#include "unity_test_utils.h"
+#include "lvgl.h"
+#include "lv_fill_common.h"
+
+#define TEST_MEMORY_LEAK_THRESHOLD (800)
```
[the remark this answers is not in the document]
peter-marcisovsky: What would be rational value for the leak check in this case?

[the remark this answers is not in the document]
tore-espressif: Some parts of newlib allocate memory after fist call to related functions (e.g. `malloc`, or `fopen`), so you will see some memory leak even though your test freed all allocated memory. Typically memory leak is not produced again in second and subsequent test runs (because newlib allocates the memory only once).

300 bytes is reasonable

components/esp_lvgl_port/test_apps/simd/functionality/main/Kconfig.projbuild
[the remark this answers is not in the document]
tore-espressif: Understood, thanks

tore-espressif:
Very nice work @peter-marcisovsky !

The only remaining thing is compatibility with LVGL 9.2, we can discuss next week

components/esp_lvgl_port/CMakeLists.txt
```diff
@@ -76,6 +76,27 @@ if("usb_host_hid" IN_LIST build_components)
     list(APPEND ADD_LIBS idf::usb_host_hid)
 endif()
 
+# Include SIMD assembly source code for rendering, only for LVGL 9.x and only for esp32 and esp32s3
+if(lvgl_ver VERSION_GREATER_EQUAL "9.0.0")
```
tore-espressif, line 80: Please add constrains for `9.0.0 <= LVG_version < 9.2.0`

components/esp_lvgl_port/idf_component.yml
```diff
@@ -4,5 +4,5 @@ url: https://github.com/espressif/esp-bsp/tree/master/components/esp_lvgl_port
 dependencies:
   idf: ">=4.4"
   lvgl/lvgl:
-    version: ">=8,<10"
+    version: "9.1.0"
```
tore-espressif, line 7: This should be reverted

tore-espressif (commented):
@peter-marcisovsky Please check if these comments correctly enable SIMD acceleration only for ESP32 and ESP32s3 and for LVGL versions 9.0 and 9.1

Then we can go ahead with merge :)

peter-marcisovsky:
@tore-espressif  
Resolved all conversations, LVGL version for this feature set in CMake to `9.1.0 <= LVG_version < 9.2.0` 
Ready to merge

espzav:
> LGTM
> 
> @espzav @igrr This is ready for merging. Please let us know if you want to have another look before we merge

I fast reviewed it again, LGTM

igrr:
Latest version LGTM!

Have you done any composite benchmarks (e.g. frames per second on some LVGL demo)? If yes, it would be interesting to see how much we gain from these assembly routines. (Not a blocker for pressing the merge button.)

PR #359, opened 2024-08-07
feat(lcd): Modified the ILI9881 and ST7796 LCD component (BSP-529)

Horion0415:

- [x] CI passing

# Change description


espzav:
@Horion0415 Thank you for changes in LCD drivers. Is it ready for review now? 
Please, rebase master for pass tests.

Horion0415:
Yes, it is ready now. Thank you for your review.

components/lcd/esp_lcd_st7796/esp_lcd_st7796.c
```diff
@@ -1,353 +1,38 @@
...
-    st7796->base.invert_color = panel_st7796_invert_color;
-    st7796->base.set_gap = panel_st7796_set_gap;
-    st7796->base.mirror = panel_st7796_mirror;
-    st7796->base.swap_xy = panel_st7796_swap_xy;
-#if ESP_IDF_VERSION < ESP_IDF_VERSION_VAL(5, 0, 0)
-    st7796->base.disp_off = panel_st7796_disp_on_off;
-#else
-    st7796->base.disp_on_off = panel_st7796_disp_on_off;
+#if SOC_LCD_I80_SUPPORTED
+    ret = esp_lcd_new_panel_st7796_general(io, panel_dev_config, ret_panel);
```
espzav, line 34: When chip supports I80 and MIPI interface and user select `use_mipi_interface`, should be called this function? 

And what about SPI usage?

espzav (commented):
@Horion0415 Thank you for this PR. I tested on HW only ILI9881C, because I haven't got HW for ST7769. 
I left one comment - how to use ST7769 with MIPI, I80 and SPI?

components/lcd/esp_lcd_st7796/esp_lcd_st7796.c
```diff
@@ -1,353 +1,38 @@
...
+esp_err_t esp_lcd_new_panel_st7796(const esp_lcd_panel_io_handle_t io, const esp_lcd_panel_dev_config_t *panel_dev_config,
+                                   esp_lcd_panel_handle_t *ret_panel)
 {
-    esp_err_t ret = ESP_OK;
-    st7796_panel_t *st7796 = NULL;
-    gpio_config_t io_conf = { 0 };
+    ESP_LOGI(TAG, "version: %d.%d.%d", ESP_LCD_ST7796_VER_MAJOR, ESP_LCD_ST7796_VER_MINOR, ESP_LCD_ST7796_VER_PATCH);
+    ESP_RETURN_ON_FALSE(panel_dev_config && ret_panel, ESP_ERR_INVALID_ARG, TAG, "Invalid arguments");
+    st7796_vendor_config_t *vendor_config = (st7796_vendor_config_t *)panel_dev_config->vendor_config;
+    ESP_RETURN_ON_FALSE(vendor_config, ESP_ERR_INVALID_ARG, TAG, "`vendor_config` is necessary");
```
espzav, line 22: This is breaking change. Can we check it only when we need it? For I80/SPI we don't need `vendor_config`, right?

components/lcd/esp_lcd_st7796/esp_lcd_st7796.c
Horion0415: This was an oversight; I forgot that the ST7796 also supports SPI. In fact, my intention was to use the `esp_lcd_new_panel_st7796_mipi` interface when the `use_mipi_interface` bit is enabled. When using `vendor_config`, or when `vendor_config` is used but `use_mipi_interface` is not enabled, the default should be the `esp_lcd_new_panel_st7796_general` interface.

components/lcd/esp_lcd_st7796/esp_lcd_st7796.c
Horion0415: Your point is indeed valid; I overlooked the fact that this is a breaking change. The issue has been fixed.

Horion0415:
Hi, @espzav could you please help resolve the CI issue? It's been giving me a headache. I noticed that the CI for this repository doesn’t always run, which makes it hard for me to know if my changes have passed the CI in time. The `esp_lcd_ili9881c` component didn’t have a test case before, but after I added a new test case, I found that the CI sets the target to a chip other than ESP32-P4, causing the CI to fail. I've added the corresponding rules in `.build-test-rules.yml`, but it still hasn't taken effect.

.build-test-rules.yml
```diff
@@ -28,3 +28,8 @@ components/esp_lvgl_port/examples/rgb_lcd:
   disable:
     - if: IDF_VERSION_MAJOR < 5
       reason: Example for RGB LCD is supported only for IDF >= 5.0
+
+components/lcd/esp_lcd_ili9881c:
+  disable:
+    - if: (IDF_VERSION_MAJOR == 5 and IDF_VERSION_MINOR < 3) or IDF_VERSION_MAJOR < 5 or IDF_TARGET != "esp32p4"
```
espzav, line 34: Please, could you try this 
```suggestion
    - if: ((IDF_VERSION_MAJOR == 5 and IDF_VERSION_MINOR < 3) or IDF_VERSION_MAJOR < 5) and IDF_TARGET != "esp32p4"
```

Or you can try this: 
```
- if: ((IDF_VERSION_MAJOR == 5 and IDF_VERSION_MINOR < 3) or IDF_VERSION_MAJOR < 5) and IDF_TARGET not in ["esp32p4"]
```

Horion0415: Alright, perhaps it's a mistake in the syntax I used. I will make the necessary adjustments and try again.

espzav: I tried on my side and fixed more issues here: https://github.com/espressif/esp-bsp/pull/373
1. File `.build-test-rules.yml`
2. Test app in esp_lcd_7796 Cmake changed and YML too (there is function, which is not supported in IDF 4.4 (I added condition for IDF 5.0 and higher)

Now, it should pass CI.

Horion0415:
Thank you very much for your help! I will sync this modification and submit it again.

.build-test-rules.yml
Horion0415: Thank you very much for your help! I will sync this modification and submit it again.

espzav: You are welcome :-) Please, did you signed CLA (Contributor License Agreement) ?  It seems, that CI is still waiting for it.

Horion0415: Sorry, there were some issues with the CLA earlier, but they have been fixed now. You can merge the PR.

espzav (approved):
@Horion0415 Thank you. LGTM

components/lcd/esp_lcd_st7796/esp_lcd_st7796.c
```diff
@@ -1,353 +1,36 @@
...
-#include "esp_log.h"
+#include "soc/soc_caps.h"
 #include "esp_check.h"
+#include "esp_lcd_types.h"
 
 #include "esp_lcd_st7796.h"
+#include "esp_lcd_st7796_interface.h"
 
-static const char *TAG = "st7796";
+const char *TAG = "st7796";
```
suda-morris, line 14: why changed the visibility?

components/lcd/esp_lcd_ili9881c/include/esp_lcd_ili9881c.h
```diff
@@ -10,25 +10,118 @@
...
+    int cmd;                /*<! The specific LCD command */
+    const void *data;       /*<! Buffer that holds the command specific data */
+    size_t data_bytes;      /*<! Size of `data` in memory, in bytes */
+    unsigned int delay_ms;  /*<! Delay in milliseconds after this command */
+} ili9881c_lcd_init_cmd_t;
+
+/**
+ * @brief LCD panel vendor configuration.
+ *
+ * @note  This structure needs to be passed to the `vendor_config` field in `esp_lcd_panel_dev_config_t`.
```
suda-morris, line 38:
```suggestion
 * @note  This structure needs to be passed to the `esp_lcd_panel_dev_config_t::vendor_config`.
```

components/lcd/esp_lcd_ili9881c/include/esp_lcd_ili9881c.h
```diff
@@ -10,25 +10,118 @@
...
+    const ili9881c_lcd_init_cmd_t *init_cmds;       /*!< Pointer to initialization commands array. Set to NULL if using default commands.
+                                                     *   The array should be declared as `static const` and positioned outside the function.
+                                                     *   Please refer to `vendor_specific_init_default` in source file.
+                                                     */
+    uint16_t init_cmds_size;                        /*<! Number of commands in above array */
+    struct {
+        esp_lcd_dsi_bus_handle_t dsi_bus;               /*!< MIPI-DSI bus configuration */
+        const esp_lcd_dpi_panel_config_t *dpi_config;   /*!< MIPI-DPI panel configuration */
+        uint8_t  lane_num;                              /*!< Number of MIPI-DSI lanes */
+    } mipi_config;
```
suda-morris, line 51: should we wrap it inside a union? so we can extend to support other interface in the future.

components/lcd/esp_lcd_ili9881c/include/esp_lcd_ili9881c.h
```diff
@@ -10,25 +10,118 @@
...
  */
-esp_err_t esp_lcd_new_panel_ili9881c(const esp_lcd_panel_io_handle_t io, const esp_lcd_panel_dev_config_t *panel_dev_config, esp_lcd_panel_handle_t *ret_panel);
+esp_err_t esp_lcd_new_panel_ili9881c(const esp_lcd_panel_io_handle_t io, const esp_lcd_panel_dev_config_t *panel_dev_config,
+                                     esp_lcd_panel_handle_t *ret_panel);
+
+/**
+ * @brief MIPI-DSI bus configuration structure
+ *
+ * @param[in] lane_num Number of data lanes
+ * @param[in] lane_mbps Lane bit rate in Mbps
```
suda-morris, line 74: where are the parameter?

components/lcd/esp_lcd_ili9881c/include/esp_lcd_ili9881c.h
```diff
@@ -10,25 +10,118 @@
...
+            .h_size = 800,                                 \
+            .v_size = 1280,                                \
+            .hsync_back_porch = 140,                       \
+            .hsync_pulse_width = 40,                       \
+            .hsync_front_porch = 40,                       \
+            .vsync_back_porch = 16,                        \
+            .vsync_pulse_width = 4,                        \
+            .vsync_front_porch = 16,                       \
+        },                                                 \
+        .flags.use_dma2d = true,                           \
```
suda-morris, line 122: DMA2D is a limited resource in esp32p4, I hope we can let the user to decide whether or not enable it, in the parameter of ILI9881C_800_1280_PANEL_60HZ_DPI_CONFIG

components/lcd/esp_lcd_ili9881c/include/esp_lcd_ili9881c.h
```diff
@@ -10,25 +10,118 @@
...
+ * @param[in] px_format Pixel format of the panel
+ *
+ */
+#define ILI9881C_800_1280_PANEL_60HZ_DPI_CONFIG(px_format) \
+    {                                                      \
+        .dpi_clk_src = MIPI_DSI_DPI_CLK_SRC_DEFAULT,       \
+        .dpi_clock_freq_mhz = 80,                          \
+        .virtual_channel = 0,                              \
+        .pixel_format = px_format,                         \
+        .num_fbs = 1,                                      \
```
suda-morris, line 111: and this one is also a configuration that we should ask the user to set it

components/lcd/esp_lcd_ili9881c/test_apps/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+# The following lines of boilerplate have to be in your project's CMakeLists
+# in this exact order for cmake to work correctly
+cmake_minimum_required(VERSION 3.5)
```
suda-morris, line 3: nitpick, 3.16

components/lcd/esp_lcd_ili9881c/esp_lcd_ili9881c.c
```diff
@@ -461,17 +554,18 @@ static esp_err_t panel_ili9881c_disp_on_off(esp_lcd_panel_t *panel, bool on_off)
...
         command = LCD_CMD_SLPOUT;
     }
-    ESP_RETURN_ON_ERROR(esp_lcd_panel_io_tx_param(io, command, NULL, 0), TAG,
-                        "io tx param failed");
+    ESP_RETURN_ON_ERROR(esp_lcd_panel_io_tx_param(io, command, NULL, 0), TAG, "send command failed");
     vTaskDelay(pdMS_TO_TICKS(100));
 
     return ESP_OK;
 }
+#endif
```
suda-morris, line 571:
```suggestion
#endif // SOC_MIPI_DSI_SUPPORTED
```

components/lcd/esp_lcd_ili9881c/esp_lcd_ili9881c.c
```diff
@@ -1,267 +1,75 @@
...
  * SPDX-License-Identifier: Apache-2.0
  */
 
-#include <stdlib.h>
-#include <sys/cdefs.h>
-#include "freertos/FreeRTOS.h"
-#include "freertos/task.h"
+#include "soc/soc_caps.h"
+
+#if SOC_MIPI_DSI_SUPPORTED
```
suda-morris, line 9: if we want to use a macro to control the build of a source file, we can also use the CMake file, this macro is also available in the CMake.

components/lcd/esp_lcd_ili9881c/esp_lcd_ili9881c.c
```diff
@@ -273,15 +81,15 @@ esp_err_t esp_lcd_new_panel_ili9881c(const esp_lcd_panel_io_handle_t io, const e
         ESP_GOTO_ON_ERROR(gpio_config(&io_conf), err, TAG, "configure GPIO for RST line failed");
     }
 
-    switch (panel_dev_config->rgb_ele_order) {
+    switch (panel_dev_config->color_space) {
```
suda-morris, line 84: color_space is a deprecated name, please use `rgb_ele_order`

components/lcd/esp_lcd_ili9881c/esp_lcd_ili9881c.c
```diff
@@ -302,151 +110,436 @@ esp_err_t esp_lcd_new_panel_ili9881c(const esp_lcd_panel_io_handle_t io, const e
...
 }
 
 static esp_err_t panel_ili9881c_init(esp_lcd_panel_t *panel)
 {
-    ili9881c_panel_t *ili9881c = __containerof(panel, ili9881c_panel_t, base);
+    ili9881c_panel_t *ili9881c = (ili9881c_panel_t *)panel->user_data;
     esp_lcd_panel_io_handle_t io = ili9881c->io;
+    const ili9881c_lcd_init_cmd_t *init_cmds = NULL;
+    uint16_t init_cmds_size = 0;
+    uint8_t lane_command = ILI9881C_DSI_2_LANE;
```
suda-morris, line 382: the lane number is configured by user, let's don't hard code it

components/lcd/esp_lcd_ili9881c/esp_lcd_ili9881c.c
```diff
@@ -302,151 +110,436 @@ esp_err_t esp_lcd_new_panel_ili9881c(const esp_lcd_panel_io_handle_t io, const e
...
+        lane_command = ILI9881C_DSI_2_LANE;
+        break;
+    case 3:
+    case 4:
+        lane_command = ILI9881C_DSI_3_4_LANE;
+        break;
+    default:
+        ESP_LOGE(TAG, "Invalid lane number %d", ili9881c->lane_num);
+        return ESP_ERR_INVALID_ARG;
+    }
```
suda-morris, line 399: question, should we always set the lane number first, then other parameters?

components/lcd/esp_lcd_ili9881c/esp_lcd_ili9881c.c
```diff
@@ -302,151 +110,436 @@ esp_err_t esp_lcd_new_panel_ili9881c(const esp_lcd_panel_io_handle_t io, const e
...
+        init_cmds_size = sizeof(vendor_specific_init_default) / sizeof(ili9881c_lcd_init_cmd_t);
+    }
+
+    for (int i = 0; i < init_cmds_size; i++) {
+        // Check if the command has been used or conflicts with the internal
+        if (is_command0_enable && init_cmds[i].data_bytes > 0) {
+            switch (init_cmds[i].cmd) {
+            case LCD_CMD_MADCTL:
+                is_cmd_overwritten = true;
+                ili9881c->madctl_val = ((uint8_t *)init_cmds[i].data)[0];
```
suda-morris, line 440: if `ili9881c->madctl_val` already equals to `((uint8_t *)init_cmds[i].data)[0]`, should we still call this a *conflict*?

components/lcd/esp_lcd_ili9881c/esp_lcd_ili9881c.c
```diff
@@ -302,151 +110,436 @@ esp_err_t esp_lcd_new_panel_ili9881c(const esp_lcd_panel_io_handle_t io, const e
...
+                         init_cmds[i].cmd);
+            }
+        }
+
+        // Send command
+        ESP_RETURN_ON_ERROR(esp_lcd_panel_io_tx_param(io, init_cmds[i].cmd, init_cmds[i].data, init_cmds[i].data_bytes), TAG, "send command failed");
+        vTaskDelay(pdMS_TO_TICKS(init_cmds[i].delay_ms));
+
+        if ((init_cmds[i].cmd == ILI9881C_CMD_CNDBKxSEL) && (((uint8_t *)init_cmds[i].data)[2] == ILI9881C_CMD_BKxSEL_BYTE2_PAGE0)) {
+            is_command0_enable = true;
```
suda-morris, line 463: what's this used for?

components/lcd/esp_lcd_st7796/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
-idf_component_register(SRCS "esp_lcd_st7796.c"
+idf_component_register(SRCS "esp_lcd_st7796.c" "esp_lcd_st7796_general.c" "esp_lcd_st7796_mipi.c"
```
suda-morris, line 1: should only add the `"esp_lcd_st7796_mipi.c"` when `CONFIG_SOC_MIPI_DSI_SUPPORTED` is true.

components/lcd/esp_lcd_st7796/include/esp_lcd_st7796.h
```diff
@@ -40,6 +44,15 @@ typedef struct {
                                                  *   Please refer to `vendor_specific_init_default` in source file.
                                                  */
     uint16_t init_cmds_size;                    /*<! Number of commands in above array */
+#if SOC_MIPI_DSI_SUPPORTED
+    struct {
+        esp_lcd_dsi_bus_handle_t dsi_bus;               /*!< MIPI-DSI bus configuration */
+        const esp_lcd_dpi_panel_config_t *dpi_config;   /*!< MIPI-DPI panel configuration */
+    } mipi_config;
```
suda-morris, line 51: question, should we use a union here in case there will a SPI/I80 specific configuration.

components/lcd/esp_lcd_st7796/esp_lcd_st7796.c
```diff
@@ -1,353 +1,36 @@
...
-
-    const st7796_lcd_init_cmd_t *init_cmds = NULL;
-    uint16_t init_cmds_size = 0;
-    if (st7796->init_cmds) {
-        init_cmds = st7796->init_cmds;
-        init_cmds_size = st7796->init_cmds_size;
     } else {
-        init_cmds = vendor_specific_init_default;
-        init_cmds_size = sizeof(vendor_specific_init_default) / sizeof(st7796_lcd_init_cmd_t);
+        ret = esp_lcd_new_panel_st7796_general(io, panel_dev_config, ret_panel);
```
suda-morris, line 32: this is the so-called "runtime dispatch", where we will always link in the SPI and I80 driver, even user only uses the MIPI interface. Leading to a significant increment to the binary size.

suda-morris (changes requested):
LGMT, thanks for the refactor @Horion0415

components/lcd/esp_lcd_st7796/esp_lcd_st7796.c
suda-morris: but it's up to you, not a blocker of this PR

components/lcd/esp_lcd_st7796/esp_lcd_st7796.c
Horion0415: Sorry, I accidentally deleted this. I will fix it later.

components/lcd/esp_lcd_ili9881c/include/esp_lcd_ili9881c.h
Horion0415: There's no need to create a union for the ILI9881, as it only supports the MIPI-DSI interface.

components/lcd/esp_lcd_ili9881c/include/esp_lcd_ili9881c.h
Horion0415: Yes, that was an oversight.

components/lcd/esp_lcd_ili9881c/include/esp_lcd_ili9881c.h
Horion0415: Similarly, ‘ILI9881C_800_1280_PANEL_60HZ_DPI_CONFIG’ is our recommended configuration. It has been tested and proven to deliver good results. If the user has specific configuration needs, we believe they should customize their own structure instead of using this configuration.

components/lcd/esp_lcd_ili9881c/esp_lcd_ili9881c.c
Horion0415: Got it. However, considering that the Arduino LCD Panel library will also migrate this component in the future, I think it's more appropriate to use macro checks in the C file.

components/lcd/esp_lcd_ili9881c/esp_lcd_ili9881c.c
Horion0415: In fact, the order of setting the lane number doesn't matter as long as it's set during initialization. However, sometimes different commands require page flipping, so handling the lane number first is more convenient.

components/lcd/esp_lcd_ili9881c/esp_lcd_ili9881c.c
Horion0415: The purpose of this check is that we do not want users to manipulate the MADCTL and COLMOD registers in the initialization parameters; these two registers should be managed by the driver. If they are manipulated during initialization, we want to provide a corresponding warning.

components/lcd/esp_lcd_ili9881c/esp_lcd_ili9881c.c
Horion0415: Similarly, this is a command page check because the same command has different effects on different pages. For example, on command page 0, 0x36 controls the MADCTL register. On another page, it controls other registers. However, we prefer to manage registers like MADCTL and COLMOD in the driver and do not want users to manipulate them in the initialization commands, as this may cause unnecessary effects. Therefore, this part mainly serves to check the command page to ensure that we can determine whether the user has manipulated registers like MADCTL and COLMOD.

components/lcd/esp_lcd_ili9881c/esp_lcd_ili9881c.c
```diff
@@ -302,151 +110,436 @@ esp_err_t esp_lcd_new_panel_ili9881c(const esp_lcd_panel_io_handle_t io, const e
...
+    ili9881c_panel_t *ili9881c = (ili9881c_panel_t *)panel->user_data;
     esp_lcd_panel_io_handle_t io = ili9881c->io;
+    uint8_t madctl_val = ili9881c->madctl_val;
+
+    ESP_RETURN_ON_FALSE(io, ESP_ERR_INVALID_STATE, TAG, "invalid panel IO");
+
+    // Control mirror through LCD command
     if (mirror_x) {
-        ili9881c->madctl_val |= ILI9881C_SS_PANEL;
+        madctl_val |= ILI9881C_CMD_GS_BIT;
```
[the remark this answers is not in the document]
Horion0415: The name of this parameter is defined in the manual and is not chosen arbitrarily.

components/lcd/esp_lcd_st7796/CMakeLists.txt
Horion0415: Similarly, this is also written this way to facilitate the migration of the component to Arduino.

components/lcd/esp_lcd_st7796/include/esp_lcd_st7796.h
Horion0415: I don't think a union is needed for now. It seems there are no additional configurations at the moment. If extra configurations are needed in the future, we can add them later.

components/lcd/esp_lcd_st7796/esp_lcd_st7796.c
Horion0415: Yes, I understand this situation. This will indeed add extra binary size, but it is done for compatibility with previous versions. I don't want this change to become a breaking change. I will consider separating them in the future.

suda-morris (approved):

PR #360, opened 2024-08-07
chore: Add obsolescence notices

tore-espressif:
ESP-BOX and ESP-BOX-Lite are End-of-Life, we provide the BSPs as-is.

ES7210 and ES8311 components are deprecated and superseded by esp_codec_dev component.

The deprecated components are excluded from CI

We limit esp_codec_dev to 1.1.x because of https://github.com/espressif/esp-adf/issues/1244


tore-espressif:
cc @Indastri

Indastri:
LGTM!

bsp/esp-box-lite/idf_component.yml
```diff
@@ -9,10 +9,10 @@ tags:
   - bsp
 
 dependencies:
-  idf: ">=4.4.5"
+  idf: ">=4.4.5,<6.0"
 
   button:
-    version: ">=2.4"
+    version: "^2.4"
```
espzav, line 15: Is it good to use this in `button` component? I think, better (safer) will be `~2.4`

bsp/esp-box/idf_component.yml
```diff
@@ -19,11 +19,11 @@ dependencies:
     override_path: "../../components/esp_lvgl_port"
 
   esp_codec_dev:
-    version: "^1.0.3"
+    version: "~1.1"
     public: true
 
   button:
-    version: ">=2.5"
+    version: "^2.5"
```
espzav, line 26: Is it good to use this in `button` component? I think, better (safer) will be `~2.4`

espzav (commented):
@tore-espressif Thank you for this PR. I left two comments. LGTM

bsp/esp-box-lite/idf_component.yml
tore-espressif: Button is already at v3.3.0 
image

So I think that fixing to v2.* release is safe enough, because I don't expect any new 2.* release.

With `^2.4` we also support v2.5 which some users might need...

espzav: Oh, ok.

espzav (approved):

PR #362, opened 2024-08-08
fix: Rename function for rotation area in LVGL port.

espzav:

- [ ] CI passing

# Change description
- Renamed function for rotation area in LVGL port (linking error with master LVGL)

### DIFF
--- a/components/esp_lvgl_port/src/lvgl9/esp_lvgl_port_disp.c
+++ b/components/esp_lvgl_port/src/lvgl9/esp_lvgl_port_disp.c
@@ -428,7 +428,7 @@ static void _lvgl_port_transform_monochrome(lv_display_t *display, const lv_area
     }
 }
 
-void lv_display_rotate_area(lv_display_t *disp, lv_area_t *area)
+void lvgl_port_rotate_area(lv_display_t *disp, lv_area_t *area)
 {
     lv_display_rotation_t rotation = lv_display_get_rotation(disp);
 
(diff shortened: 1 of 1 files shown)


components/esp_lvgl_port/src/lvgl9/esp_lvgl_port_disp.c
```diff
@@ -428,7 +428,7 @@ static void _lvgl_port_transform_monochrome(lv_display_t *display, const lv_area
     }
 }
 
-void lv_display_rotate_area(lv_display_t *disp, lv_area_t *area)
+void lvgl_port_rotate_area(lv_display_t *disp, lv_area_t *area)
```
tore-espressif, line 431: Other option for consideration:
* We can declare this function as `weak` so the linker can use either this or LVGL's implementation, if it is present

I'm not sure though, whether this is a good idea for this case...

components/esp_lvgl_port/idf_component.yml
```diff
@@ -1,4 +1,4 @@
-version: "2.3.0"
+version: "2.3.0~1"
```
tore-espressif, line 1: Could we keep the version for now? I expect more changes to esp_lvgl_port soon

tore-espressif (commented):
Thanks for the quick fix @espzav !

components/esp_lvgl_port/src/lvgl9/esp_lvgl_port_disp.c
espzav: I suggest, we can keep it like this now.

components/esp_lvgl_port/idf_component.yml
espzav: ok

tore-espressif (approved):

PR #363, opened 2024-08-08
fix(esp_lvgl_port): Fix esp_lvgl_port with local LVGL component

tore-espressif:
esp_lvgl_port component configuration is entirely dependent on LVGL's version. Moreover LVGL can be used as local or managed component. This PR aims to support all possible combinations

* Related to https://github.com/lvgl/lvgl/pull/6654
* esp_lvgl_port cannot be linked with LVGL's master branch, we must merge this first #362 

From esp_lvgl_port/CMakeLists.txt
``` cmake
# This component uses a CMake workaround, so we can compile esp_lvgl_port for both LVGL8.x and LVGL9.x
# At the time of idf_component_register() we don't know which LVGL version is used, so we only register an INTERFACE component (with no sources)
# Later, when we know the LVGL version, we create another CMake library called 'lvgl_port_lib' and link it to the 'esp_lvgl_port' INTERFACE component
```

### DIFF
--- a/components/esp_lvgl_port/CMakeLists.txt
+++ b/components/esp_lvgl_port/CMakeLists.txt
@@ -1,17 +1,34 @@
 include($ENV{IDF_PATH}/tools/cmake/version.cmake) # $ENV{IDF_VERSION} was added after v4.3...
 
 if("${IDF_VERSION_MAJOR}.${IDF_VERSION_MINOR}" VERSION_LESS "4.4")
-    return() 
+    return()
 endif()
 
-#Get LVGL version
-idf_component_get_property(lvgl_ver lvgl__lvgl COMPONENT_VERSION)
-if(lvgl_ver EQUAL "")
-    idf_component_get_property(lvgl_ver lvgl COMPONENT_VERSION)
+# This component uses a CMake workaround, so we can compile esp_lvgl_port for both LVGL8.x and LVGL9.x
+# At the time of idf_component_register() we don't know which LVGL version is used, so we only register an INTERFACE component (with no sources)
+# Later, when we know the LVGL version, we create another CMake library called 'lvgl_port_lib' and link it to the 'esp_lvgl_port' INTERFACE component
+idf_component_register(
+        INCLUDE_DIRS "include"
+        PRIV_INCLUDE_DIRS "priv_include"
+        REQUIRES "esp_lcd")
+
+# Get LVGL version
+idf_build_get_property(build_components BUILD_COMPONENTS)
+if(lvgl IN_LIST build_components)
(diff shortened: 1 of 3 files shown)


espzav (approved):
@tore-espressif Good work! Thank you, it is nice solution :-) LGTM

components/esp_lvgl_port/CMakeLists.txt
```diff
@@ -1,17 +1,34 @@
...
-    return() 
+    return()
 endif()
 
-#Get LVGL version
-idf_component_get_property(lvgl_ver lvgl__lvgl COMPONENT_VERSION)
-if(lvgl_ver EQUAL "")
-    idf_component_get_property(lvgl_ver lvgl COMPONENT_VERSION)
+# This component uses a CMake workaround, so we can compile esp_lvgl_port for both LVGL8.x and LVGL9.x
+# At the time of idf_component_register() we don't know which LVGL version is used, so we only register an INTERFACE component (with no sources)
```
igrr, line 8: FWIW, you can add a dummy source file here, and then call `target_sources(${COMPONENT_LIB} PRIVATE ${PORT_PATH}/esp_lvgl_port.c` once you know LVGL version, later.

components/esp_lvgl_port/CMakeLists.txt
```diff
@@ -66,9 +76,22 @@ if("usb_host_hid" IN_LIST build_components)
...
+add_library(lvgl_port_lib STATIC
+    ${PORT_PATH}/esp_lvgl_port.c
+    ${PORT_PATH}/esp_lvgl_port_disp.c
+    ${ADD_SRCS}
+    )
+target_include_directories(lvgl_port_lib PUBLIC "include")
+target_include_directories(lvgl_port_lib PRIVATE "priv_include")
+target_link_libraries(lvgl_port_lib PUBLIC
+    idf::esp_lcd
+    idf::${lvgl_name}
```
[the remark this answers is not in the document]
tore-espressif: I tested with `set(COMPONENTS main)` and did not encounter any problem.

I think the reason is that we require lvgl even from the idf_component.yml https://github.com/espressif/esp-bsp/blob/master/components/esp_lvgl_port/idf_component.yml#L6

So it can't be eliminated

components/esp_lvgl_port/CMakeLists.txt
tore-espressif: This is a backup plan :D I like this solution better, although I'm not 100% sure about its future-proofness or maintainability...

tore-espressif:
> LGTM, just two minor notes.

@igrr thank you very much for having a look!

PR #364, opened 2024-08-08
docs(esp_lvgl_port): remove LVGL9 unstable notice

tore-espressif:
LVGL9 seems to be stable now. Can we remove the warning?


espzav (approved):
LGTM

PR #368, opened 2024-08-12
fix(ci): Use fixed version of idf-component-manager

tore-espressif:
Fix idf-component-version to 1.*. 

Version 2.* was released today and it is not compatible


espzav (approved):
@tore-espressif Thank you for this fix! LGTM

Issue #358, opened 2024-08-05
esp_lcd_touch: Add a track_id for esp_lcd_touch_get_coordinates


lijunru-hub:
### Related area

esp_lcd_touch

### Hardware specification

ALL Chip

### Is your feature request related to a problem?

In multi-touch systems, the track ID is a crucial piece of information. Some chips, such as the GT911, inherently support track ID information. We should ensure that the driver exposes this information.

### Describe the solution you'd like

like **bool esp_lcd_touch_get_coordinates(esp_lcd_touch_handle_t tp, uint16_t *x, uint16_t *y, uint16_t *strength, uint8_t *point_num, uint8_t max_point_num)**

### I have checked existing list of Feature requests and the Contribution Guide

- [X] I confirm I have checked existing list of Feature requests and Contribution Guide.

lijunru-hub:
cc@espzav

lijunru-hub:
In GT911's datasheet, it can be supported easily.
![image](https://github.com/user-attachments/assets/7d3710e6-5068-4a5d-b74d-cf0262eabaaa)

lijunru-hub:
FT5x06
![image](https://github.com/user-attachments/assets/dac53a38-7d72-4534-8d28-1c3ab310bb4e)

tore-espressif:
@lijunru-hub sounds like a nice feature. We can discuss during next BSP meeting :) 
cc @espzav

PetrESP:
This feature has been implemented in [this](https://github.com/espressif/esp-bsp/pull/691) and [this](https://github.com/espressif/esp-bsp/pull/682) pull request

PR #369, opened 2024-08-13
feat(lcd_touch): support trank id

lijunru-hub:

- [ ] CI passing

# Change description

link https://github.com/espressif/esp-bsp/issues/358


components/lcd_touch/esp_lcd_touch/esp_lcd_touch.c
```diff
@@ -58,7 +58,7 @@ esp_err_t esp_lcd_touch_read_data(esp_lcd_touch_handle_t tp)
     return tp->read_data(tp);
 }
 
-bool esp_lcd_touch_get_coordinates(esp_lcd_touch_handle_t tp, uint16_t *x, uint16_t *y, uint16_t *strength, uint8_t *point_num, uint8_t max_point_num)
+bool esp_lcd_touch_get_coordinates(esp_lcd_touch_handle_t tp, uint16_t *x, uint16_t *y, uint16_t *strength, uint8_t *track_id, uint8_t *point_num, uint8_t max_point_num)
```
espzav, line 61: We cannot change API. We can create another function with similar API.

components/lcd_touch/esp_lcd_touch/include/esp_lcd_touch.h
```diff
@@ -59,7 +59,7 @@ typedef struct {
     } flags;
 
     /*!< User callback called after get coordinates from touch controller for apply user adjusting */
-    void (*process_coordinates)(esp_lcd_touch_handle_t tp, uint16_t *x, uint16_t *y, uint16_t *strength, uint8_t *point_num, uint8_t max_point_num);
+    void (*process_coordinates)(esp_lcd_touch_handle_t tp, uint16_t *x, uint16_t *y, uint16_t *strength, uint8_t *track_id, uint8_t *point_num, uint8_t max_point_num);
```
espzav, line 62: We cannot change API. We can create another function with similar API.

lijunru-hub: What would be a good name for the new function?

The current function name is static bool esp_lcd_touch_ft5x06_get_xy.

espzav: We have two options for it:
1. `bool esp_lcd_touch_get_coordinates_track_id(esp_lcd_touch_handle_t tp, uint8_t *track_id);`
2. `bool esp_lcd_touch_get_coordinates_track_id(esp_lcd_touch_handle_t tp, uint16_t *x, uint16_t *y, uint16_t *strength, uint8_t *track_id, uint8_t *point_num, uint8_t max_point_num)`

Same for `bool (*get_xy)(...)` and callback `void (*process_coordinates)(...)`

lijunru-hub: However, the touch chip actually supports some additional special features. For example, the FT5x06 can indicate whether a swipe is to the left or right, and whether each touch point is currently being pressed or swiped, among other things. It would be best to address these issues all at once and lay the groundwork for future requirements.

```c
bool esp_lcd_touch_get_coordinates_data(esp_lcd_touch_handle_t tp, void *data)
{
   ft5x06_data_t *data = (ft5x06_data_t *)data；
   // data->x[i]
   // data->trank_id
   //...
}
```

espzav: It is good idea, but it looks little dangerous use void type and we don't know const count of touches.

 We can do something like this:
 ```
 typedef struct {
    uint16_t x;
    uint16_t y;
    uint16_t strength;
    uint16_t track_id;
    void * user_data;
 } esp_lcd_touch_coordinates_data_t;

bool esp_lcd_touch_get_coordinates_data(esp_lcd_touch_handle_t tp, esp_lcd_touch_coordinates_data_t * data, uint8_t *point_num, uint8_t max_point_num);

 ```
Usage:
```
esp_lcd_touch_coordinates_data_t data[4];
uint8_t points = 0;
esp_lcd_touch_get_coordinates_data(tp, data, &points, sizeof(data)/sizeof(esp_lcd_touch_coordinates_data_t));
```

Note: This structure can be very easily extended in future. And any driver can use `user_data` for specific strusture. There should be any data type check (I don't know, if it is possible in C - something like compare `typeof()`)

@igrr @tore-espressif What do you think about this solution?

lijunru-hub: I think this idea is great, but some information might pertain to the overall system. For instance, certain touch controllers, like the FT5x06, can provide data on the current gesture, such as zooming in/out or moving up/down/left/right. This type of information doesn't belong to any single touch point.

image

lijunru-hub: How about handling everything within a struct? 

typedef struct {
   uint16_t x;
   uint16_t y;
   uint16_t strength;
   uint16_t track_id;
} point_data_t;

typedef struct {
   esp_lcd_touch_gesture_t gesture_id;
   point_data_t *data;
   uint32_t num_points;
} esp_lcd_touch_coordinates_data_t;


However, it's important to note that not every chip has all of this information

espzav: As you said, some features are not supported in all chips, I think the gestures should be in own function and we can return NOT_SUPPORTED error.

tore-espressif: > And any driver can use user_data for specific structure. There should be any data type check (I don't know, if it is possible in C - something like compare typeof())

If we want to return device specific data, this should be handled in the specific touch driver eg. `esp_lcd_touch_ft5x06` and NOT in `esp_lcd_touch` base component. C language does not have runtime type information

So there are 2 types of returned data
1. somewhat common across different touch controllers (eg. track_id). These could be added to base `esp_lcd_touch` with a function similar to `esp_lcd_touch_get_coordinates_data`
2. Specific to the touch driver e.g. gestures. These function should be added to the specific touch driver eg. `esp_lcd_touch_ft5x06_gesture()`

What do you think?

espzav: > 2\. Specific to the touch driver e.g. gestures. These function should be added to the specific touch driver eg. `esp_lcd_touch_ft5x06_gesture()`

Are `gestures` specific for touch driver? Aren't it more common over lot of drivers?

tore-espressif: > Are gestures specific for touch driver? Aren't it more common over lot of drivers?

🤷‍♂️ :D

If yes, let's make it all common

lijunru-hub:
hi all @tore-espressif @espzav ，what is the plan about this feature

espzav:
> hi all @tore-espressif @espzav ，what is the plan about this feature

Hi @lijunru-hub, we can continue with this PR. Do you agree with proposed changes of the new API instead of breaking changes?

lijunru-hub:
> > hi all @tore-espressif @espzav ，what is the plan about this feature
> 
> Hi @lijunru-hub, we can continue with this PR. Do you agree with proposed changes of the new API instead of breaking changes?

Sure，i agree

espzav:
> > > hi all @tore-espressif @espzav ，what is the plan about this feature
> > 
> > 
> > Hi @lijunru-hub, we can continue with this PR. Do you agree with proposed changes of the new API instead of breaking changes?
> 
> Sure，i agree

Thank you. It seems that I made a mistake in my question and you may bad understand it. I thought, if you have any time to continue please? If not, we try to add into our plan, but it will be in Q2 first.

lijunru-hub:
@espzav 
It seems that I haven't had much time to focus on this feature. Could you please provide an update on its current progress?

PetrESP:
In the end this feature was implemented in [this](https://github.com/espressif/esp-bsp/pull/691) and [this](https://github.com/espressif/esp-bsp/pull/682) pull request

PR #383, opened 2024-08-27
fix: Do not access LVGL 9.2 private members

tore-espressif:
LVGL 9.2 makes `user_data` private. Access it with provided getter functions


pborcin (approved):
LGTM

PR #384, opened 2024-08-28
Draft: Release process for noglib BSPs

tore-espressif:
## Release process for noglib BSPs

* noglib BSPs will be released from [release/noglib](https://github.com/espressif/esp-bsp/tree/release/noglib) branch
* release/noglib branch is updated with 'merge' strategy (not rebasing) to preserve git history
* There are 2 stages in the process
1. Check
1.1. Create noglib BSPs
1.2. Build noglib example from #372 
1.3. Check correctness of idf_component.yml files
2. Release (only on master branch)
2.1. Create noglib BSPs
2.2. Push changes to release/noglib
2.3. Push noglib BSPs to ESP-Registry

## Related
* Passing release job https://github.com/espressif/esp-bsp/actions/runs/10610982648/job/29409614982
* noglib example in #372 

## TODO
- [x] Update documentation!
- [x] Write PR description


tore-espressif:
@georgik @igrr This is a proposal how to relase noglib BSP versions. PTAL at the PR description and READMEs to find out more!

georgik:
Current solution looks okeyish.

If I understand the process. Default BSP is with LVGL and then NOGLIB is reduction of existing BSP.

Wouldn't it be possible to reverse the architecture, that NOGLIB is base dependency for BSP with LVGL? That would make more sense from software point of view.

Instead of reduction, it will be natural extension and also scripts could be reduced.

Please, note that names of BSP's are hardcoded several times in the scripts. This will lead to errors, when someone will add BSP in one place and forget to add it to another.

tore-espressif:
> Wouldn't it be possible to reverse the architecture, that NOGLIB is base dependency for BSP with LVGL? That would make more sense from software point of view.

I agree that this way it would be cleaner solution. The only reason why we create NOGLIB from LVGL version is historical (the original BSP contains LVGL) and so we don't create breaking change for existing BSP users

> Please, note that names of BSP's are hardcoded several times in the scripts. This will lead to errors, when someone will add BSP in one place and forget to add it to another.

thanks, I'll have a look

PR #401, opened 2024-10-05
feat(p4): Added avoid-tear feature and audio-related code. (BSP-560)

Horion0415:

- [ ] CI passing

# Change description
- Added tear-proofing feature in the LVGL context
- Added audio-related code
- Optimized some configurations and initialization methods


tore-espressif (approved):
@Horion0415 thank you very much for the PR! I left few nitpicks/questions

@espzav PTAL and the esp_lvgl_port part :)

bsp/esp32_p4_function_ev_board/include/bsp/esp32_p4_function_ev_board.h
```diff
@@ -58,12 +68,17 @@
...
-#define BSP_SD_D3             (GPIO_NUM_42)
-#define BSP_SD_CMD            (GPIO_NUM_44)
-#define BSP_SD_CLK            (GPIO_NUM_43)
+#define BSP_SD_SLOT_0_DEFAULT_INIT     \
+    .clk = 0,                          \
+    .cmd = 0,                          \
+    .d0  = 0,                          \
+    .d1  = 0,                          \
+    .d2  = 0,                          \
+    .d3  = 0,                          \
```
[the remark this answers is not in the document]
Horion0415: Yes, this part has been tested. Since slot_0 was selected in the configuration and the pin is fixed, you can simply fill in 0. In fact, only this initialization method can avoid conflicts with esp-hosted.

bsp/esp32_p4_function_ev_board/include/bsp/esp32_p4_function_ev_board.h
```diff
@@ -58,12 +68,17 @@
 #endif
 
 /* uSD card */
-#define BSP_SD_D0             (GPIO_NUM_39)
-#define BSP_SD_D1             (GPIO_NUM_40)
-#define BSP_SD_D2             (GPIO_NUM_41)
-#define BSP_SD_D3             (GPIO_NUM_42)
-#define BSP_SD_CMD            (GPIO_NUM_44)
-#define BSP_SD_CLK            (GPIO_NUM_43)
```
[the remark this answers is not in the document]
Horion0415: Okay, it should indeed be kept as part of the content.

igrr:
@Horion0415 SD card changes have been merged in #405, please rebase on top of it. Sorry for the trouble!

components/esp_lvgl_port/include/esp_lvgl_port_disp.h
```diff
@@ -80,7 +80,9 @@ typedef struct {
  * @brief Configuration MIPI-DSI display structure
  */
 typedef struct {
-    int dummy;
+    struct {
+        unsigned int avoid_tearing: 1;  /*!< 1: Use internal RGB buffers as a LVGL draw buffers to avoid tearing effect */
```
igrr, line 84: Is there any case when users would need to set this to 0? From the description it seems like it should always be set to 1. In which case, does it need to be configurable?

bsp/esp32_p4_function_ev_board/include/bsp/esp32_p4_function_ev_board.h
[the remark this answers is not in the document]
Horion0415: Oh, I see. Thank you very much for fixing this issue. I will resume using the correct pin numbers moving forward.

components/esp_lvgl_port/include/esp_lvgl_port_disp.h
Horion0415: In fact, enabling tear prevention means frame synchronization is required, which can impact the frame rate. Additionally, enabling tear prevention will consume more memory resources, and for some users with relatively simple UI scenarios, this feature may not be necessary. Therefore, I believe it's important to keep this option available.

igrr: Great, thanks for explaining. Can you mention some of these cons in the description of this field? E.g. "enabling this option increases RAM usage and reduces frame rate."

Horion0415: Screen tearing occurs when reading from and writing to the same frame buffer simultaneously, resulting in old and new frames appearing on the same screen. This creates a noticeable tear for the human eye. To address this issue for RGB screens and MIPI-DSI interface screens, we typically use multiple frame buffers to implement anti-tearing operations. This requires allocating an additional 1 to 2 full-screen buffers. For example, with a screen resolution of 1024 x 600 on the P4 development board and using the RGB565 color format, this results in an additional PSRAM memory usage of 1M to 2M.

As mentioned earlier, we use a multi-buffer approach to avoid tearing, which requires buffer switching and frame synchronization (for instance, we need to wait until the screen has completely read a frame of data before switching frame buffers). This is inherently more time-consuming than not using anti-tearing features (where frame buffers are continuously rendered without synchronization), leading to a certain degree of frame rate drop. However, for some users with simple UI scenarios (e.g., screens that change infrequently), it is likely that they can meet their needs without utilizing this anti-tearing feature.

Horion0415: I will describe these drawbacks in the comments later. Thank you very much for reviewing this part!

components/esp_lvgl_port/src/lvgl8/esp_lvgl_port_disp.c
```diff
@@ -415,19 +451,19 @@ static void lvgl_port_flush_callback(lv_disp_drv_t *drv, const lv_area_t *area,
...
     if (disp_ctx->trans_size == 0) {
-        if (disp_ctx->disp_type == LVGL_PORT_DISP_TYPE_RGB && (drv->direct_mode || drv->full_refresh)) {
+        if ((disp_ctx->disp_type == LVGL_PORT_DISP_TYPE_RGB || disp_ctx->disp_type == LVGL_PORT_DISP_TYPE_DSI) && (drv->direct_mode || drv->full_refresh)) {
             if (lv_disp_flush_is_last(drv)) {
                 /* If the interface is I80 or SPI, this step cannot be used for drawing. */
                 esp_lcd_panel_draw_bitmap(disp_ctx->panel_handle, x_start, y_start, x_end + 1, y_end + 1, color_map);
                 /* Waiting for the last frame buffer to complete transmission */
-                ulTaskNotifyValueClear(NULL, ULONG_MAX);
-                ulTaskNotifyTake(pdTRUE, portMAX_DELAY);
+                xSemaphoreTake(disp_ctx->trans_sem, 0);
```
espzav, line 459: Please, could you update this with wait callback like this: https://github.com/espressif/esp-bsp/blob/61ab9cb60d5206962a563cb1caaf4a291d40bb80/components/esp_lvgl_port/src/lvgl9/esp_lvgl_port_disp.c#L476 (latest master)

Horion0415: Sure, I can add an interface. Additionally, I want to explain that the change to a semaphore is due to certain scenarios. For example, when using a camera with LVGL, to prevent screen tearing, the camera needs to call lv_refr_now(). With the previous approach, it would cause the LVGL task to hang.

espzav: Ok, I understand. But with wait callback it can be faster. LVGL can continue work during waiting and task is not hanged.

Horion0415: Yes, the original method would indeed be faster, but from my testing, I found that the impact isn't significant. In the future, the ESP32-P4 development board may release some AI features integrated with LVGL, such as facial recognition. For now, this feature is still needed for support.

Horion0415: In fact, I believe the lvgl_port_flush_wait_callback interface should not be used. Before calling ulTaskNotifyValueClear(NULL, ULONG_MAX) and ulTaskNotifyTake(pdTRUE, portMAX_DELAY), esp_lcd_panel_draw_bitmap should be called first to ensure buffer switching can occur. (Calling esp_lcd_panel_draw_bitmap will not immediately switch the buffer, but will wait until the current frame is completely flushed before switching.) Otherwise, rendering and reading might occur on the same buffer.

bsp/esp32_p4_function_ev_board/esp32_p4_function_ev_board.c
```diff
@@ -434,10 +577,25 @@ static lv_display_t *bsp_display_lcd_init(const bsp_display_cfg_t *cfg)
             .swap_bytes = (BSP_LCD_BIGENDIAN ? true : false),
 #endif
             .sw_rotate = cfg->flags.sw_rotate, /* Only SW rotation is supported for 90° and 270° */
+#if CONFIG_BSP_DISPLAY_LVGL_FULL_REFRESH
```
espzav, line 580: I think, we should set buff_dma = false, when full refresh is enabled, because we haven't got enough memory.

components/esp_lvgl_port/src/lvgl8/esp_lvgl_port.c
```diff
@@ -198,10 +198,10 @@ static void lvgl_port_task(void *arg)
             task_delay_ms = lv_timer_handler();
             lvgl_port_unlock();
         }
-        if ((task_delay_ms > lvgl_port_ctx.task_max_sleep_ms) || (1 == task_delay_ms)) {
+        if (task_delay_ms > lvgl_port_ctx.task_max_sleep_ms) {
             task_delay_ms = lvgl_port_ctx.task_max_sleep_ms;
-        } else if (task_delay_ms < 1) {
-            task_delay_ms = 1;
+        } else if (task_delay_ms < 5) {
```
espzav, line 203: Why did you changed to 5?

components/esp_lvgl_port/src/lvgl8/esp_lvgl_port.c
```diff
@@ -198,10 +198,10 @@ static void lvgl_port_task(void *arg)
             task_delay_ms = lv_timer_handler();
             lvgl_port_unlock();
         }
-        if ((task_delay_ms > lvgl_port_ctx.task_max_sleep_ms) || (1 == task_delay_ms)) {
```
espzav, line 201: ` || (1 == task_delay_ms)` this part of code was recommendation from LVGL for right handling sleep. Why did you remove it?

components/esp_lvgl_port/src/lvgl8/esp_lvgl_port_disp.c
```diff
@@ -239,7 +247,14 @@ static lv_disp_t *lvgl_port_add_disp_priv(const lvgl_port_display_cfg_t *disp_cf
 #if CONFIG_IDF_TARGET_ESP32S3 && ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 0, 0)
         buffer_size = disp_cfg->hres * disp_cfg->vres;
         ESP_GOTO_ON_ERROR(esp_lcd_rgb_panel_get_frame_buffer(disp_cfg->panel_handle, 2, (void *)&buf1, (void *)&buf2), err, TAG, "Get RGB buffers failed");
+#elif CONFIG_IDF_TARGET_ESP32P4 && ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 3, 0)
+        buffer_size = disp_cfg->hres * disp_cfg->vres;
+        ESP_GOTO_ON_ERROR(esp_lcd_dpi_panel_get_frame_buffer(disp_cfg->panel_handle, 2, (void *)&buf1, (void *)&buf2), err, TAG, "Get RGB buffers failed");
 #endif
+
+        trans_sem = xSemaphoreCreateCounting(1, 0);
```
espzav, line 255: It seems, we forgot delete semaphore in `lvgl_port_remove_disp`. Same for LVGL8 and LVGL9.

bsp/esp32_p4_function_ev_board/esp32_p4_function_ev_board.c
Horion0415: Yes, that was my oversight.

components/esp_lvgl_port/src/lvgl8/esp_lvgl_port.c
Horion0415: In the initial version of the BSP I implemented, the test results for the LVGL Music Demo were higher compared to the current version. After testing and comparison, it was found that adjusting this part to 5 leads to an improvement.

components/esp_lvgl_port/src/lvgl8/esp_lvgl_port_disp.c
Horion0415: Yes, that was an oversight.

components/esp_lvgl_port/src/lvgl8/esp_lvgl_port.c
espzav: ok, we can keep it as it is, when there is better performance :-)

bsp/esp32_p4_function_ev_board/esp32_p4_function_ev_board.c
Horion0415: I have reviewed this part, and I think it doesn’t need to be modified because, after enabling the tear prevention feature, no additional buffer will be requested. Instead, it will use the buffer allocated by the LCD, so the allocation method is not effective.

components/esp_lvgl_port/src/lvgl8/esp_lvgl_port_disp.c
espzav: Please, could you update it in `components/esp_lvgl_port/src/lvgl9/esp_lvgl_port_disp.c` too?

Horion0415: Oh, I forgot. Alright, I will make changes later.

espzav: Thanks.

bsp/esp32_p4_function_ev_board/esp32_p4_function_ev_board.c
espzav: I tried it again and I found, where is the problem. When SW rotation is enabled and `full_refresh` is enabled too, the buffer for SW rotation must be full too and there is not enough memory in DMA (capabilities of rotation buffer are same as selected for draw buffer).

Horion0415: image
This was an oversight. In fact, in both modes (direct-mode and full-refresh), it is not possible to use the lv_display_rotate_area(display, &area) interface for rotation. Users will have to handle full-frame rotation themselves.

espzav: ok, then we should disable `.sw_rotate = false,` in LVGL9 in direct and full_refresh mode.

Horion0415: Okay, I have made changes according to the situation. Please check if this is suitable. In fact, this limitation has existed since v8.

Horion0415: Great, the issue has been resolved! Thank you very much.

espzav:
@Horion0415 Sorry for my mistake, I overlooked - there is missing update version of `esp_lvgl_port` + update changelog.

Horion0415:
Alright, that was my oversight. Considering the significant modifications to the esp_lvgl_port this time, along with the addition of some new features, I have bumped the minor version. I'm not sure if this is appropriate?

espzav:
> Alright, that was my oversight. Considering the significant modifications to the esp_lvgl_port this time, along with the addition of some new features, I have bumped the minor version. I'm not sure if this is appropriate?

Thank you very much! It is ok, the major version is usually the big change (breaking change)

espzav (approved):
LGTM

igrr:
@espzav seems that this PR has introduced a build warning:

```
/__w/idf-extra-components/idf-extra-components/thorvg/examples/thorvg-example/managed_components/espressif__esp32_p4_function_ev_board_noglib/esp32_p4_function_ev_board.c: In function 'bsp_audio_init':
/__w/idf-extra-components/idf-extra-components/thorvg/examples/thorvg-example/managed_components/espressif__esp32_p4_function_ev_board_noglib/esp32_p4_function_ev_board.c:184:15: error: unused variable 'ret' [-Werror=unused-variable]
  184 |     esp_err_t ret = ESP_FAIL;
      |               ^~~
```
https://github.com/espressif/idf-extra-components/actions/runs/11603657651/job/32311078844#step:5:30731

tore-espressif:
Will fix soon, sorry for the inconvenience...

tore-espressif:
Fixed here https://github.com/espressif/esp-bsp/pull/427

Horion0415:
I am truly sorry for this oversight. (╥﹏╥)

PR #405, opened 2024-10-11
fix(esp32_p4_function_ev_board): use slot 0 for SD card

igrr:
SD card on P4 EV board is connected to slot 0. Previously the code used slot 1 and routed it through GPIO Matrix to slot 0 pins. This resulted in conflicts when trying to use slot 1 for connecting to ESP32-C6 over SDIO.

See https://github.com/espressif/esp-idf/issues/14699#issuecomment-2407264707 for background.

### DIFF
--- a/bsp/esp32_p4_function_ev_board/esp32_p4_function_ev_board.c
+++ b/bsp/esp32_p4_function_ev_board/esp32_p4_function_ev_board.c
@@ -89,6 +89,7 @@ esp_err_t bsp_sdcard_mount(void)
     };
 
     sdmmc_host_t host = SDMMC_HOST_DEFAULT();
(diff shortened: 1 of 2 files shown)


tore-espressif:
@igrr Thanks for the PR, similar implementation here https://github.com/espressif/esp-bsp/pull/401

If you need it now, we can rebase this PR and the rebase&update https://github.com/espressif/esp-bsp/pull/401

igrr:
Ooops, I've seen that one but didn't realize from the title that it also fixed SD card.

If you plan on merging #401 today then I can close this one. If not, maybe merge this one first and then rebase #401.

tore-espressif:
Will merge after CI pass

tore-espressif (approved):

PR #407, opened 2024-10-11
feat(esp32_p4_function_ev_board): add example of wifi connection

igrr:
This PR adds a simple example of connecting to Wi-Fi on ESP32-P4 EV board.

This example adds esp_wifi_remote as a dependency and sets the required sdkconfig options.

The goal is to allow users to quickly get started with Wi-Fi connectivity on this board by running 
```
idf.py create-project-from-example espressif/esp32_p4_function_ev_board:wifi_connection
```


igrr:
@tore-espressif WDYT about adding such an example to bsp/esp32_p4_function_ev_board/examples directory?

tore-espressif:
If ESP32-P4 + ESP32-C6 hardware setup is planned for all our P4 devkits, I (as a user) would expect this to be in esp-idf examples.

If the ESP32-C6 modem extension is not going to be typical usecase, we can leave the example here

igrr:
I think we will keep ESP32-C6 on ESP32-P4 EV board, but we can't be sure that makers of other P4 boards will also add ESP32-C6. If they do, still they might opt to use a different set of pins for SDIO.

On IDF side, so far support for P4 is being added only in `iperf` example. Still, it will take some time to merge this and this will only be available in IDF master branch for some time.

So I thought that making this example easily available via the component manager would be a reasonable solution in the meantime.

tore-espressif (approved):

igrr:
@tore-espressif Might need to ask for your help with the pre-commit hook. I'm guessing that the script is considering example README.md file as a BSP readme file... Could you suggest what I can change to make this check succeed?

tore-espressif:
@igrr Please ignore the pre-commit fail, probably some issue on our side

tore-espressif:
> Might need to ask for your help with the pre-commit hook. I'm guessing that the script is considering example README.md file as a BSP readme file... Could you suggest what I can change to make this check succeed?

We can force merge, if you need it now. I'll take a look on Monday

bsp/esp32_p4_function_ev_board/examples/wifi_connection/main/idf_component.yml
```diff
@@ -0,0 +1,8 @@
+dependencies:
+  espressif/esp32_p4_function_ev_board:
+    version: "*"
+    override_path: "../../../"
+  espressif/esp_wifi_remote:
+    version: "*"
```
igrr, line 6: Probably makes sense to pin at least some version of esp_wifi_remote (and maybe esp_hosted?) here.

tore-espressif:
@igrr I rebased your PR, the pre-commit check is passing now

bsp/esp32_p4_function_ev_board/examples/wifi_connection/main/idf_component.yml
espzav: I agree, do you plan update and set some major version here?

espzav (approved):
@igrr Thank you for this example, good work! LGTM

bsp/esp32_p4_function_ev_board/examples/wifi_connection/main/idf_component.yml
```diff
@@ -0,0 +1,8 @@
+dependencies:
+  espressif/esp32_p4_function_ev_board:
+    version: "*"
```
igrr, line 3: Seems like the BSP dependency isn't actually needed...

espzav:
As we discussed, I am closing this PR, we have WiFi example for ESP32-P4 in esp-idf. 
If we need it in future, we can reopen it.

PR #414, opened 2024-10-22
feat(esp-box-3): Support I2C Driver-NG

tore-espressif:
Update esp-box-3 to driver-NG

In the end I had to add `i2c_master_bus_handle_t bsp_i2c_get_handle(void);` because IDF function `i2c_master_get_bus_handle` is only present in IDF v5.4 and was not backported

Anyway, we depend on v5.3 because of esp_codec_dev which does not support Driver-NG in v5.2


espzav (approved):
@tore-espressif Thank you for this PR and for the fix my accelerometter issues :-) 
LGTM

Issue #400, opened 2024-09-27
Software rotation with esp_lvgl_port on ESP32-P4 with ILI9881C display


dmartauz:
### Board

ESP32-P4-Function-EV-Board

### Hardware Description

ILI9881C touchscreen display (8" 800 x 1280 pixels)

### IDE Name

esp-idf/master

### Operating System

Windows 10

### Description

The display natively works in portrait orientation and it is displaying simple tabview. In this orientation it works well.

![IMG_20240927_201910708](https://github.com/user-attachments/assets/abf3dcee-3684-4e58-81b6-07b106a442a6)

Because I want to use it in landscape orientation and it seems that ILI9881C does not support hardware rotation I am attempting to use `.sw_rotate = true` and after uncommenting last line of the code it seems that first rotated frame is drawn correctly but following frames have some offset (shifted LVGL FPS tracker).

![IMG_20240927_202104171](https://github.com/user-attachments/assets/ffc7e15e-8e14-4eb5-8b62-850632000a05)

After clicking on the tabview also tabview buttons get shifted.

![IMG_20240927_202144160](https://github.com/user-attachments/assets/09b18fd0-ff23-4599-b746-7633c6c54b3e)

Is master branch of esp_lvgl_port already using PPA in case of ESP32-P4?

### Sketch

```cpp
    const lvgl_port_display_cfg_t disp_cfg = {
        .io_handle = mipi_dbi_io,
        .panel_handle = mipi_dpi_panel,
        .buffer_size = EXAMPLE_MIPI_DSI_LCD_V_RES * EXAMPLE_MIPI_DSI_LCD_H_RES, //EXAMPLE_LVGL_DRAW_BUF_LINES
        .double_buffer = true,
        .hres = EXAMPLE_MIPI_DSI_LCD_H_RES,
        .vres = EXAMPLE_MIPI_DSI_LCD_V_RES,
        .monochrome = false,
        .color_format = LV_COLOR_FORMAT_RGB565,
        .rotation = {
            .swap_xy = false,
            .mirror_x = false,
            .mirror_y = false,
        },
        .flags = {
            .buff_dma = false,
            .buff_spiram = true,
            .sw_rotate = true,
            .swap_bytes = false,
            .full_refresh = false,
            .direct_mode = false,
        }
    };

    const lvgl_port_display_dsi_cfg_t disp_dsi_cfg;

[...]

PR #418, opened 2024-10-24
fix(lvgl_port): Fixed SW rotation in LVGL9.2 and fixed free buffers.

espzav:

- [ ] CI passing

# Change description
- Fixed SW rotation in LVGL9.2 (https://github.com/espressif/esp-bsp/issues/400)
- Fixed freeing right buffers when error


tore-espressif (approved):
@espzav Thanks! It looks like straight forward fix, should I check on HW too?

Issue #420, opened 2024-10-26
esp_bsp_generic fails to build after the the the led_strip change in upstream


rishi9699:
### Board

All

### Hardware Description

Irrelevant

### IDE Name

Irrelevant

### Operating System

Ubuntu 24

### Description

There has been a [change](https://github.com/espressif/idf-extra-components/issues/438) in the led_strip source code which has caused esp_bsp_generic builds to fail. I tried running the 'light' example in esp-matter and the build is failing. Giving me the following log,

/home/rishi/light/managed_components/espressif__esp_bsp_generic/src/esp_bsp_generic.c:197:6: error: 'led_strip_config_t' has no member named 'led_pixel_format'
197 | .led_pixel_format = LED_PIXEL_FORMAT_GRB, // Pixel format of your LED strip
| ^~~~~~~~~~~~~~~~
/home/rishi/light/managed_components/espressif__esp_bsp_generic/src/esp_bsp_generic.c:197:25: error: 'LED_PIXEL_FORMAT_GRB' undeclared here (not in a function)
197 | .led_pixel_format = LED_PIXEL_FORMAT_GRB, // Pixel format of your LED strip
| ^~~~~~~~~~~~~~~~~~~~
/home/rishi/light/managed_components/espressif__esp_bsp_generic/src/esp_bsp_generic.c:222:22: error: initializer element is not constant
222 | .led_strip_cfg = bsp_leds_rgb_strip_config,

### Sketch

```cpp
None
```

### I have checked existing issues, README.md and ESP32 Forum

- [X] I confirm I have checked existing issues, online documentation and Troubleshooting guide.

rishi9699:
I have created a [pull request](https://github.com/espressif/esp-bsp/pull/423) for the fixing the same.

PR #423, opened 2024-10-27
incorporating relevant changes from led_strip v3.0.0 (BSP-574)

rishi9699:

- [ ] CI passing

# Change description
Recently a new version (v3.0.0) of led_strip was pushed. This changed the `led_strip_config_t` struct in the 

> led_strip/include/led_strip_types.h

 file.

Please check the latest change of led_strip repository [here](https://github.com/espressif/idf-extra-components/commit/31585c8c97b5a403e1550270fb9fad58879cc2dd)

Closes https://github.com/espressif/esp-bsp/issues/420

### DIFF
--- a/bsp/esp_bsp_generic/src/esp_bsp_generic.c
+++ b/bsp/esp_bsp_generic/src/esp_bsp_generic.c
@@ -194,7 +194,6 @@ static led_indicator_gpio_config_t bsp_leds_gpio_config[] = {
 static const led_strip_config_t bsp_leds_rgb_strip_config = {
     .strip_gpio_num = CONFIG_BSP_LED_RGB_GPIO,   // The GPIO that connected to the LED strip's data line
     .max_leds = BSP_LED_NUM,                  // The number of LEDs in the strip,
-    .led_pixel_format = LED_PIXEL_FORMAT_GRB, // Pixel format of your LED strip
     .led_model = LED_MODEL_WS2812,            // LED strip model
     .flags.invert_out = false,                // whether to invert the output signal
 };
--- a/bsp/esp32_c3_lcdkit/esp32_c3_lcdkit.c
+++ b/bsp/esp32_c3_lcdkit/esp32_c3_lcdkit.c
@@ -68,7 +68,6 @@ static i2s_chan_handle_t i2s_tx_chan;
 static const led_strip_config_t bsp_strip_config = {
     .strip_gpio_num = BSP_RGB_CTRL,
     .max_leds = 1,
-    .led_pixel_format = LED_PIXEL_FORMAT_GRB,
     .led_model = LED_MODEL_WS2812,
     .flags.invert_out = false,
 };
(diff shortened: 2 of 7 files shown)


igrr:
cc @espzav @tore-espressif

I think we also need to add a version constraint on led_strip somewhere, or possibly have ifdefs for led_strip v2 and v3.

bsp/esp_bsp_generic/src/esp_bsp_generic.c
```diff
@@ -194,8 +194,8 @@ static led_indicator_gpio_config_t bsp_leds_gpio_config[] = {
 static const led_strip_config_t bsp_leds_rgb_strip_config = {
     .strip_gpio_num = CONFIG_BSP_LED_RGB_GPIO,   // The GPIO that connected to the LED strip's data line
     .max_leds = BSP_LED_NUM,                  // The number of LEDs in the strip,
-    .led_pixel_format = LED_PIXEL_FORMAT_GRB, // Pixel format of your LED strip
     .led_model = LED_MODEL_WS2812,            // LED strip model
+    .color_component_format = LED_STRIP_COLOR_COMPONENT_FMT_GRB,  // LED color component format
```
suda-morris, line 198:
```suggestion
```

suda-morris: leave this field empty means, i.e. init the field to `0`, the led_strip will fall-back to use the default format "GRB" as well.

Then this esp_bsp_generic package can work both with led_strip version2 and led_strip version3.

suda-morris:
> cc @espzav @tore-espressif
> 
> I think we also need to add a version constraint on led_strip somewhere, or possibly have ifdefs for led_strip v2 and v3.

We'd better to set a proper dependency in [here](https://github.com/espressif/esp-iot-solution/blob/master/components/led/led_indicator/idf_component.yml#L7)

igrr:
True, led_indicator component should also change the way the dependency is set.

However, since esp_bsp_generic also uses led_strip API directly, it should specify the dependency and not rely on the version set in led_indicator.

Besides, I think it's better to allow both versions of led_strip, if possible. If we force led_strip>=3, we have to make a new major release of the BSP component. If we force led_strip<3, users won't be able to use esp_bsp_generic with led_strip>=3.

suda-morris:
> Besides, I think it's better to allow both versions of led_strip, if possible.

Yes, we can. Just remove the configuration of `led_pixel_format` and `color_component_format` when initialize the led_strip, let the driver fall-back to choose the default one. The esp_bsp_generic can compile with led_strip version 2 and 3.

suda-morris:
I take the liberty to fix the build errors that are introduced by led_strip v3. Now the bsp package can work with both led_strip v2 and led_strip v3. @igrr @espzav PTAL.

suda-morris (approved):

Lzw655 (approved):
Hi @suda-morris, LGTM. Please also update the version of the `esp32-c3-lcdkit` BSP. I tested the LED functionality using the compatible development board, and everything works fine.

suda-morris:
> Hi @suda-morris, LGTM. Please also update the version of the `esp32-c3-lcdkit` BSP. I tested the LED functionality using the compatible development board, and everything works fine.

c3-lcdkit already has a contraint to the led_strip:

```yml
  led_strip:
    version: "^2"
    public: true
```

So this PR actually didn't do any bugfix for that BSP, it's just a refactor.

PR #439, opened 2024-11-22
fix(lvgl_port_disp): rgb panel callback bad ctx param (BSP-582)

liamHowatt:

- [ ] CI passing

# Change description

Fix a crash when using https://github.com/lvgl/lv_port_espressif_esp32-s3-lcd-ev-board with ESP IDF >= v5.1.2. Versions of ESP IDF older than v5.1.2 are not effected by the bug because the dependencies are solved in such a way that the file `components/esp_lvgl_port/src/lvgl9/esp_lvgl_port_disp.c` (where the issue is) is not compiled.

The issue occurs when the callback `lvgl_port_flush_rgb_vsync_ready_callback` receives a `user_ctx` which is `&disp_ctx->disp_drv` which is `lv_display_t **` when the function expects it to be `lv_display_t *`. A hardfault happens at the dereference here: https://github.com/espressif/esp-bsp/blob/4bbe669a6a8f48ebe0909d355b5581ede2973878/components/esp_lvgl_port/src/lvgl9/esp_lvgl_port_disp.c#L424

The hardfault could also happen 3 lines earlier inside `lv_display_get_user_data` if the struct layout aliasing lottery plays out that way.

Reproduce by:
1. clone https://github.com/lvgl/lv_port_espressif_esp32-s3-lcd-ev-board
2. `idf.py flash monitor`

A boot loop will occur with a crash at the line I linked.

Specific board:
Mainboard "ESP32-S3-LCD_Ev_Board_MB_V1.0" with the sub-board "ESP32-S3-LCD_Ev_Board_SUB3_V1.0"

cc @espzav


liamHowatt:
Thanks for the Info! Updated!

components/esp_lvgl_port/src/lvgl8/esp_lvgl_port_disp.c
```diff
@@ -152,9 +152,9 @@ lv_display_t *lvgl_port_add_disp_rgb(const lvgl_port_display_cfg_t *disp_cfg, co
         };
 
         if (rgb_cfg->flags.bb_mode && (ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 1, 2))) {
-            ESP_ERROR_CHECK(esp_lcd_rgb_panel_register_event_callbacks(disp_ctx->panel_handle, &bb_cbs, &disp_ctx->disp_drv));
+            ESP_ERROR_CHECK(esp_lcd_rgb_panel_register_event_callbacks(disp_ctx->panel_handle, &bb_cbs, disp_ctx->disp_drv));
```
espzav, line 155: In LVGL8 it was right, because `disp_drv`  is not pointer but struct.

components/esp_lvgl_port/src/lvgl8/esp_lvgl_port_disp.c
```diff
@@ -152,9 +152,9 @@ lv_display_t *lvgl_port_add_disp_rgb(const lvgl_port_display_cfg_t *disp_cfg, co
...
         } else {
-            ESP_ERROR_CHECK(esp_lcd_rgb_panel_register_event_callbacks(disp_ctx->panel_handle, &vsync_cbs, &disp_ctx->disp_drv));
+            ESP_ERROR_CHECK(esp_lcd_rgb_panel_register_event_callbacks(disp_ctx->panel_handle, &vsync_cbs, disp_ctx->disp_drv));
```
espzav, line 157: In LVGL8 it was right, because `disp_drv`  is not pointer but struct.

espzav (commented):
@liamHowatt Thank you for the fix in LVGL9. I am not sure, why it was working before :-( In LVGL8 it was correct, because there is not pointer but structure.

liamHowatt:
> @liamHowatt Thank you for the fix in LVGL9. I am not sure, why it was working before :-(

I'm not too sure exactly why either. I had some ideas in the description of this PR but I really don't know. I think I recall running into this crash some months ago. It's an amazing coincidence DamienEspitallier decided to fix it at the same time.

> In LVGL8 it was correct, because there is not pointer but structure.

Ah, thanks.

> Sorry for misleading you about lvgl8... :(

It's ok :slightly_smiling_face: I should've double checked it.

espzav (approved):
@liamHowatt Thank you, LGTM

PR #447, opened 2024-12-01
Wrong dependency version (BSP-590)

devprofile98:
# ESP-BSP Pull Request checklist

> Note: For new BSPs create a PR with this [link](https://github.com/espressif/esp-bsp/compare/main...my-branch?quick_pull=1&template=pr_template_bsp.md).
- [x] Version of modified component bumped
- [ ] CI passing

# Change description
The **add-dependency** for **esp_lcd_touch_gt911** installs a version that is incompatible with the provided usage example.

**esp_lcd_touch_io_gt911_config_t** type is defined in version **1.1.1** [here](https://github.com/espressif/esp-bsp/commit/d118a01b29764b32c5ad2ec54fe0bf72f3769699#diff-9d9605942dc95af8aed100641fc40e7fe6ab8fdde90092e32197f8db3cd1bb2dR49) , so the **1.0.0** version is not compatible with the example.

PR #448, opened 2024-12-02
Update our good-old test-app

tore-espressif:
Most of components in this repository now have their dedicated test_app or example.

So now, the old test_app is build only for the explicitely added components.

Fixes CI in #447


espzav (approved):
@tore-espressif Thank you for this PR. LGTM

PR #450, opened 2024-12-04
Fix: knob loses diff event

espressif2022:

- [ ] CI passing

# Change description
1: The new component of the `knob`  will clear the flag of  [knob->event](https://github.com/espressif/esp-iot-solution/blob/c683b16aa9f7c3202952fb814d19ed126847e75d/components/knob/iot_knob.c#L195), so we add `count_value`  for conditional checks.

### DIFF
--- a/components/esp_lvgl_port/src/lvgl8/esp_lvgl_port_knob.c
+++ b/components/esp_lvgl_port/src/lvgl8/esp_lvgl_port_knob.c
@@ -128,8 +128,10 @@ static void lvgl_port_encoder_read(lv_indev_drv_t *indev_drv, lv_indev_data_t *d
     knob_event_t event = iot_knob_get_event(ctx->knob_handle);
 
     if (last_v ^ invd) {
+        data->enc_diff = (event == KNOB_LEFT) ? -1 :
+                         (event == KNOB_RIGHT) ? 1 :
+                         (int32_t)((uint32_t)invd - (uint32_t)last_v);
         last_v = invd;
-        data->enc_diff = (KNOB_LEFT == event) ? (-1) : ((KNOB_RIGHT == event) ? (1) : (0));
     } else {
         data->enc_diff = 0;
     }
--- a/components/esp_lvgl_port/src/lvgl9/esp_lvgl_port_knob.c
+++ b/components/esp_lvgl_port/src/lvgl9/esp_lvgl_port_knob.c
@@ -139,8 +139,10 @@ static void lvgl_port_encoder_read(lv_indev_t *indev_drv, lv_indev_data_t *data)
     knob_event_t event = iot_knob_get_event(ctx->knob_handle);
 
     if (last_v ^ invd) {
+        data->enc_diff = (event == KNOB_LEFT) ? -1 :
+                         (event == KNOB_RIGHT) ? 1 :
+                         (int32_t)((uint32_t)invd - (uint32_t)last_v);
         last_v = invd;
-        data->enc_diff = (KNOB_LEFT == event) ? (-1) : ((KNOB_RIGHT == event) ? (1) : (0));
     } else {
         data->enc_diff = 0;
     }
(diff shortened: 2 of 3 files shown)


components/esp_lvgl_port/src/lvgl9/esp_lvgl_port_knob.c
```diff
@@ -139,8 +139,10 @@ static void lvgl_port_encoder_read(lv_indev_t *indev_drv, lv_indev_data_t *data)
     knob_event_t event = iot_knob_get_event(ctx->knob_handle);
 
     if (last_v ^ invd) {
+        data->enc_diff = (event == KNOB_LEFT) ? -1 :
```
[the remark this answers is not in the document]
espressif2022: Yeah, maybe we can register a new knob event . For the new knob, we can hardly read the knob events, we can delete the event handle in the `lvgl_port_encoder_read`.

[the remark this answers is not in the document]
espressif2022: @espzav I was a bit busy recently and haven't updated yet. Please check again.

components/esp_lvgl_port/src/lvgl8/esp_lvgl_port_knob.c
```diff
@@ -125,11 +131,17 @@ static void lvgl_port_encoder_read(lv_indev_drv_t *indev_drv, lv_indev_data_t *d
...
 
     int32_t invd = iot_knob_get_count_value(ctx->knob_handle);
-    knob_event_t event = iot_knob_get_event(ctx->knob_handle);
+    knob_event_t event = ctx->event;
 
     if (last_v ^ invd) {
+
+        int32_t diff = (int32_t)((uint32_t)invd - (uint32_t)last_v);
+
+        diff += (event == KNOB_RIGHT && invd < last_v) ? CONFIG_KNOB_HIGH_LIMIT :
```
espzav, line 140: Why we need this? 
`event` is only last event and when knob rotate to right by some steps then to left, the `event` will left, but more steps can be still right.

espzav (commented):
@espressif2022 Thank you for update. I will test it today or tomorrow with HW :-)

components/esp_lvgl_port/src/lvgl8/esp_lvgl_port_knob.c
espressif2022: Yeah, you're right. In some case, as follows:
Starting at 10, rotating right by 10 [20] (not captured by LVGL this time), and then rotating left by 5 [15].

This would result in a `rotate left` event, but with a positive diff, leading to an abnormal diff for the left rotation.

Should we ignore the diff?

espzav: In your example case, there is still direction (diff as you naimed). I think, the direction is only in sign minus or plus value. That's all.

espressif2022: Use interrupts to record the current diff value in real time. 
Accumulate the value when rotating in the same direction; reset to zero and start recalculating when the direction reverses.

espzav: Why do you reset value when changed direction? I thought, it is enough to add or sub from latest value.

espressif2022: If the encoder is rotated `10 steps to the right`, but `lvgl_port_encoder_read `does not capture it (assuming the UI is experiencing severe lag at this moment), and then it is rotated `5 steps to the left`, I think a result of `5 steps to the left` is `better` than `5 steps to the right`. 
Therefore, the count for the different direction should be reset.

espzav (approved):
@espressif2022 Thank you for changes. LGTM
Please, squash commits and update branch before merging.

tore-espressif:
@espressif2022 Could you please add a note to CHANGELOG.md too?

PR #457, opened 2024-12-16
Use driver_data instead of user_data (BSP-600)

KenVanHoeylandt:
# Change description

`esp_lvgl_port` was using user data to store data. When I started wrapping the drivers in my project and using the user data, this broke `esp_lvgl_port`.
Since LVGL9, there's a "driver data" attribute, so I refactored the existing drivers to use the driver data functions instead, which frees up the user data, which can then be left to the user.


KenVanHoeylandt:
I had to revert some parts: This was new code from `master` which wasn't present in LVGL 9.0 for my project.
I've reverted the relevant bits and tested it in my project (using this branch). It compiles and LVGL works.
I used a Lilygo T-Deck Plus.

tore-espressif (approved):
Thank you for the fix!

espzav (approved):
@KenVanHoeylandt Thank you for this update. LGTM

Issue #455, opened 2024-12-10
Potential Exception Due to Task Priority in `esp_lvgl_port` with LVGL9 (BSP-596)


nicechao:
### Board

ESP32S3

### Hardware Description

ESP32S3+ST7789

### IDE Name

ESP-IDF5.3 + esp_lvgl_port 2.4.3 + lvgl 9.2.2

### Operating System

UBUNTU24.10

### Description

During code debugging, the following exception occurred:

```
Guru Meditation Error: Core  1 panic'ed (LoadProhibited). Exception was unhandled.

Core  1 register dump:
PC      : 0x42028d0e  PS      : 0x00060330  A0      : 0x82028d9c  A1      : 0x3fcb1aa0  
0x42028d0e: search_suitable_block at /projpath/managed_components/lvgl__lvgl/src/stdlib/builtin/lv_tlsf.c:563

A2      : 0x00000000  A3      : 0x3fcb1ac0  A4      : 0x3fcb1ac4  A5      : 0x3fcaed94  
A6      : 0x00000000  A7      : 0x00000000  A8      : 0x00000010  A9      : 0x3fcb1a80  
A10     : 0x00000003  A11     : 0x3fcb1ac0  A12     : 0x3fcb1ac4  A13     : 0x00000000  
A14     : 0x3c07d097  A15     : 0x3fcb163c  SAR     : 0x00000004  EXCCAUSE: 0x0000001c  
EXCVADDR: 0x00000014  LBEG    : 0x400556d5  LEND    : 0x400556e5  LCOUNT  : 0xfffffffd  
0x400556d5: strlen in ROM
0x400556e5: strlen in ROM

Backtrace: 0x42028d0b:0x3fcb1aa0 0x42028d99:0x3fcb1ac0 0x42028f3d:0x3fcb1af0 0x420279c4:0x3fcb1b10 0x42029100:0x3fcb1b30 0x420264b2:0x3fcb1b80 0x42019c31:0x3fcb1ba0 0x4201114a:0x3fcb1be0 0x42011680:0x3fcb1c30 0x4200adda:0x3fcb1c60 0x4200ae4b:0x3fcb1cd0 0x4037ecf9:0x3fcb1d10
0x42028d0b: search_suitable_block at managed_components/lvgl__lvgl/src/stdlib/builtin/lv_tlsf.c:563
0x42028d99: block_locate_free at managed_components/lvgl__lvgl/src/stdlib/builtin/lv_tlsf.c:769
[...]

PR #468, opened 2025-01-03
fix(esp_lvgl_port): Make sure LVGL is initialize at the end of lvgl_port_init() call

tore-espressif:
Previously, we initialized LVGL in the new LVGL handling task. If lvgl_port_init() was called from high priority task, it could return with uninitialized LVGL.

Closes https://github.com/espressif/esp-bsp/issues/455
Closes https://github.com/espressif/esp-bsp/issues/475

### DIFF
--- a/components/esp_lvgl_port/CHANGELOG.md
+++ b/components/esp_lvgl_port/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [Unreleased]
+
+### Fixes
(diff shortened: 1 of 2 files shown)


espzav:
@tore-espressif This is not good fix. There is reason, why init is called from LVGL port task. It should be in the same task like `lv_timer_handler`.

tore-espressif:
> @tore-espressif This is not good fix. There is reason, why init is called from LVGL port task. It should be in the same task like `lv_timer_handler`.

Fixed by FreeRTOS task notifications

espzav (approved):
@tore-espressif Thank you for this fix. LGTM

PR #470, opened 2025-01-03
Fix(esp_lcd_touch_cst816s): Missing field initializers under C++ (BSP-609)

tajnymag:
# ESP-BSP Pull Request checklist

> Note: For new BSPs create a PR with this [link](https://github.com/espressif/esp-bsp/compare/main...my-branch?quick_pull=1&template=pr_template_bsp.md).
- [x] Version of modified component bumped
- [ ] CI passing

# Change description
ESP_LCD_TOUCH_IO_I2C_CST816S_CONFIG macro was causing warning under C++. I've added initialization to all fields.

PR #474, opened 2025-01-09
fix(esp_lcd_touch_cst816s): Fixed compilation error for IDF5.1 and lower

espzav:

- [x] CI passing
## Related
Caused by #470
# Change description
- Fixed IFD5.1 compilation issue in CST816S touch controller driver
- We need to CI tests for all components, which aren't used in BSPs. Because we need test it for all IDF versions before merge PR!


espzav:
@tore-espressif PTAL

tore-espressif (approved):

PR #478, opened 2025-01-11
BSP optional I2C pull up (BSP-620)

adamkaliszan:
There is no option in KCONFIG to enable I2C pull up. This option is supported by ESP-IDF.

This option is extremely helpful for newbies.


espzav (approved):
@adamkaliszan thank you for this change, LGTM

PR #482, opened 2025-01-14
fix(ci): Build i2c_example only for IDF >= 5.2

tore-espressif:
Our CI was broken for some time and did not catch this issue.
Could close #481

- [ ] CI passing

# Change description
Build i2c_example only for IDF >= 5.2

### DIFF
--- a/components/.build-test-rules.yml
+++ b/components/.build-test-rules.yml
@@ -4,7 +4,21 @@ components/esp_lvgl_port:
     - "components/esp_lvgl_port/**"
     - "components/lcd_touch/esp_lcd_touch_tt21100/**"
     - "components/lcd_touch/esp_lcd_touch_gt1151/**"
+
+components/esp_lvgl_port/examples/i2c_oled:
+  depends_filepatterns:
+    - "components/esp_lvgl_port/**"
     - "components/lcd/sh1107/**"
+  disable:
+    - if: (IDF_VERSION_MAJOR == 5 and IDF_VERSION_MINOR < 2) or IDF_VERSION_MAJOR < 5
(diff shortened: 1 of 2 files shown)


components/.build-test-rules.yml
```diff
@@ -4,7 +4,21 @@ components/esp_lvgl_port:
...
+  disable:
+    - if: (IDF_VERSION_MAJOR == 5 and IDF_VERSION_MINOR < 2) or IDF_VERSION_MAJOR < 5
+      reason: Requires I2C Driver-NG which was introduced in v5.2
+
+components/esp_lvgl_port/test_apps/simd:
+  depends_filepatterns:
+    - "components/esp_lvgl_port/**"
+  enable:
+    - if: IDF_TARGET in ["esp32", "esp32s3"]
+      reason: Supports only xtensa targets
```
tore-espressif, line 21: @peter-marcisovsky This test_apps was not build in CI before. 

Could you please check if the 'reason' makes sense here?

tore-espressif:
@espzav I'm sorry for the CI issues, this should fix all esp_lvgl_port apps, so we can close https://github.com/espressif/esp-bsp/pull/481

components/.build-test-rules.yml
peter-marcisovsky: Sorry I missed this notification. 
The reason is fine, we can also add support for esp32s2 (it requires only minor changes into the esp32 assembly), to have complete Xtensa targets support.

tore-espressif: Thanks! Sorry, we merged it already, it was a CI blocker... will update the rest later

PR #500, opened 2025-02-03
bsp: Update BSPs for using button v4

espzav:

- [x] CI passing

# Change description
- Updated BSPs for using it with button component v4

# Updated BSPs to button v4
- [x] esp_bsp_devkit (done in another PR)
- [x] esp_bsp_generic (done in another PR)
- [x] esp_wrover_kit
- [x] esp32_azure_iot_kit
- [x] esp32_c3_lcdkit
- [x] esp32_lyrat
- [x] esp32_p4_function_ev_board (no buttons)
- [x] esp32_s2_kaluga_kit
- [x] esp32_s3_eye
- [x] esp32_s3_korvo_1 - NOT TESTED ON HW
- [x] esp32_s3_korvo_2
- [x] esp32_s3_lcd_ev_board
- [x] esp32_s3_usb_otg
- [x] esp-box-3
- [x] m5dial
- [x] m5stack_core
- [x] m5stack_core_2 (no buttons)
- [x] m5stack_core_s3 (no buttons)

## No SD card = RELEASE
- esp32_c3_lcdkit
- esp32_s2_kaluga_kit 
- esp32_s3_lcd_ev_board
- m5dial


components/esp_lvgl_port/src/lvgl9/esp_lvgl_port_button.c
```diff
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2024 Espressif Systems (Shanghai) CO LTD
```
atanisoft, line 2: does the lvgl8 port need similar modifications?

espzav: Yes, it does. I will do it. This PR is not finished yet.

espzav:
@tore-espressif PTAL

components/esp_lvgl_port/src/lvgl9/esp_lvgl_port_button.c
espzav: Now, the PR is finished and ready for review :-)

components/esp_lvgl_port/CMakeLists.txt
```diff
@@ -119,5 +123,15 @@ target_link_libraries(lvgl_port_lib PRIVATE
...
 
+if (NOT button_ver)
+    message("IOT button version is not set or empty!")
+    set(button_ver "4.0.0")
+endif()
+
+#Get component button major version
+string(REPLACE "." ";" BUTTON_VERSION_LIST ${button_ver})
+list(GET BUTTON_VERSION_LIST 0 button_ver_major)
+target_compile_definitions(lvgl_port_lib PUBLIC "-DCOMPONENT_BUTTON_VERSION_MAJOR=${button_ver_major}")
```
tore-espressif, line 134: The version define should come from the button component 

https://github.com/espressif/esp-iot-solution/blob/4730d91db70df7e6e0a3191d725ab1c5f98ff9ce/tools/cmake_utilities/package_manager.cmake#L44

So it should be `BUTTON_VER_MAJOR`. Could you please try it? It would help us avoid any convoluted CMake logic

espzav: @tore-espressif What do you think about backward compatibility of BSP? Now, there is no breaking change of API, but it is not compatible with old button component. Should it be compatible with both? 
Other BSPs will be updated in next PR.

tore-espressif (approved):
@espzav esp_lvgl_port LGTM.

I don't have any of these boards now to test...

Strictly speaking, bump of major version of public component is a breaking change -> we should bump major versions of the BSPs too... 
However, looking at the button v4 changes, there are only 2:
* Button creation -> this is handled by BSP
* Register/unregistering of button callback -> this will require users to change their code, if they use these functions

espzav:
> However, looking at the button v4 changes, there are only 2:

I am not sure, if there are real breaking changes:
1. When you are using button v3 with LVGL port --> No breaking changes
2. When you change from v3 to v4 button with LVGL port --> You must make changes
3. In BSPs, there is everything handled inside or in examples --> No breaking changes

espzav:
And when I am thinking about it, there can be solution for use both version of button in BSPs too. I am not suer, if we want it (add more ifdefs)

tore-espressif:
> And when I am thinking about it, there can be solution for use both version of button in BSPs too. I am not suer, if we want it (add more ifdefs)

I'd prefer supporting only one (the latest) version

espzav:
> I'd prefer supporting only one (the latest) version

OK, should I increase major version of affected BSPs?

lijunru-hub:
Sorry for the late reply. The custom button can be created using the factory mode of the new button component. You can refer to this file:  

[[custom_button_test.c](https://github.com/espressif/esp-iot-solution/blob/master/components/button/test_apps/main/custom_button_test.c)](https://github.com/espressif/esp-iot-solution/blob/master/components/button/test_apps/main/custom_button_test.c)

espzav:
> Sorry for the late reply. The custom button can be created using the factory mode of the new button component. You can refer to this file:
> 
> [[custom_button_test.c](https://github.com/espressif/esp-iot-solution/blob/master/components/button/test_apps/main/custom_button_test.c)](https://github.com/espressif/esp-iot-solution/blob/master/components/button/test_apps/main/custom_button_test.c)

@lijunru-hub Thank you!

espzav:
@tore-espressif I updated this PR with changes for all BSPs with button v4. PTAL

lijunru-hub:
LGTM!

PR #503, opened 2025-02-05
Fix I2C NG includes

tore-espressif:

- [x] CI passing

# Change description
* For BSP that exposed public function `i2c_master_bus_handle_t bsp_i2c_get_handle(void);` (korvo-2 and s3-eye), the `esp_driver_i2c` component must be listed in public dependencies
* On CoreS3 the camera is now enabled during `bsp_i2c_init()` call. Previously it was enabled during `bsp_display_new()`, but this caused problems if the users wanted to initialize the camera before LCD initialization

### DIFF
--- a/bsp/esp-box-3/include/bsp/esp-box-3.h
+++ b/bsp/esp-box-3/include/bsp/esp-box-3.h
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2022-2024 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2022-2025 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -16,7 +16,6 @@
 #include "driver/i2s_std.h"
 #include "driver/i2c_master.h"
 #include "driver/sdmmc_host.h"
-#include "soc/usb_pins.h"
 #include "lvgl.h"
 #include "esp_lvgl_port.h"
 #include "esp_codec_dev.h"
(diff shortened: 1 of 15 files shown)


bsp/m5stack_core_s3/m5stack_core_s3.c
```diff
@@ -88,6 +89,13 @@ esp_err_t bsp_i2c_init(void)
     BSP_ERROR_CHECK_RETURN_ERR(i2c_master_bus_add_device(i2c_handle, &aw9523_config, &aw9523_h));
 
     i2c_initialized = true;
+
+    /*
+     * Here we explicitly enable camera, because there is no other API that would enable it.
+     * All other peripherals (touch, audio, SD card...) are enabled in their init call (eg. bsp_sdcard_mount()),
+     * but we do not have a BSP init call for camera.
+     */
+    BSP_ERROR_CHECK_RETURN_ERR(bsp_enable_feature(BSP_FEATURE_CAMERA));
```
espzav, line 98: We can make it as public API like here: https://github.com/espressif/esp-bsp/blob/6c402325d9e2624608143ce39ead9c65bb9a7b34/bsp/m5stack_core_2/include/bsp/m5stack_core_2.h#L357C11-L357C29 
And then enable it in example or application.

tore-espressif: mm, but then I cannot write 1 example for multiple BSPs... we did not consider this before... Let's discuss tomorrow

espzav: I understand. I am thinking, that there will be nice to add some define to each BSP. Something like this:
```
#define BSP_ESP_BOX_3
```

And then we can use `#ifdef BSP_ESP_BOX_3` for special API in examples. I have similar problem in benchmark example [here](https://github.com/espressif/esp-bsp/blob/244b10bb4bd598178d27e1f31a5e10b190eacf9d/examples/display_lvgl_benchmark/main/main.c#L21)

tore-espressif:
@espzav 
* I removed the commit that changed `bsp_enable_feature` for now, so the other changes are not delayed
* I removed include of `soc/usb_pins.h` as this file will be removed in IDF6.0

PR #505, opened 2025-02-10
fix(esp_lvgl_port): Add support for LVGL 9.2 and above

tore-espressif:

- [ ] CI passing

# Change description
LVGL has changed their blending API here https://github.com/lvgl/lvgl/pull/6688 , now it seems that it is considered private.
Moving forward, we plan to move the blending code to LVGL repo


components/esp_lvgl_port/include/esp_lvgl_port_lv_blend.h
```diff
@@ -18,6 +18,19 @@ extern "C" {
...
+/**
+ * LVGL 9.2 changed the blending API compared to 9.1
+ * The types were moved to lv_draw_sw_blend_private.h
+ * and the leading underscore from type names was removed
+ */
+#if __has_include("draw/sw/blend/lv_draw_sw_blend_private.h")
+#include "draw/sw/blend/lv_draw_sw_blend_private.h"
+#define _lv_draw_sw_blend_fill_dsc_t lv_draw_sw_blend_fill_dsc_t
+#define _lv_draw_sw_blend_image_dsc_t lv_draw_sw_blend_image_dsc_t
+#endif
```
tore-espressif, line 32: Hi @kisvegabor,

We’re including this compatibility code so our blending acceleration can compile correctly for both LVGL 9.1 and 9.2. When this feature was first implemented, the blending interface was intended to be public, but it now appears to be treated as a private API. Could you please share your plans for the blending interface moving forward? Are there any intentions to expose it publicly, or should we plan for further adjustments if it remains private?

Thanks!

components/esp_lvgl_port/CMakeLists.txt
```diff
@@ -77,9 +77,8 @@ if("usb_host_hid" IN_LIST build_components)
 endif()
 
 # Include SIMD assembly source code for rendering, only for (9.1.0 <= LVG_version < 9.2.0) and only for esp32 and esp32s3
```
peter-marcisovsky, line 79:
```suggestion
# Include SIMD assembly source code for rendering, only for LVG_version >= 9.1.0 and only for esp32 and esp32s3
```

peter-marcisovsky (commented):
I can confirm that on this branch, the SIMD source code present in `esp-bsp` compiles successfully on `LVGL >= 9.1.x`

This solution solves both problems with SIMD src code support on `LVGL 9.1.x` and `LVGL 9.2.x`
1. changing structure names (`_lv_draw_sw_blend_fill_dsc_t` vs `lv_draw_sw_blend_fill_dsc_t`)

2. `LV_DRAW_SW_ASM_CUSTOM_INCLUDE` (Macro which we use to include our header file `esp_lvgl_port_lv_blend.h` with assembly function prototypes) being present in multiple places around LVGL (`lv_draw_sw.c` for example) and this .c file is missing include of `lv_draw_sw_blend_private.h`, causing compilation to fail

components/esp_lvgl_port/include/esp_lvgl_port_lv_blend.h
tore-espressif: Fair enough, thank you!

tore-espressif:
@espzav  PTAL :)

examples/display_lvgl_demos/main/CMakeLists.txt
```diff
@@ -1,17 +1,3 @@
-set(LV_DEMO_DIR "")
-set(LV_DEMOS_SOURCES "")
-if(CONFIG_LV_USE_DEMO_BENCHMARK)
-    list(APPEND LV_DEMO_DIR ../managed_components/lvgl__lvgl/demos)
-    file(GLOB_RECURSE LV_DEMOS_SOURCES ${LV_DEMO_DIR}/*.c)
-endif()
```
espzav, line 6: Why these lines were removed? Now, the `lv_demo_benchmark` cannot be compiled.

tore-espressif: will check, thanks

tore-espressif: @espzav I had to increse size of `factory` partition so the benchmark demo would fit. But apart from that it works OK. I'd like to use the files from LVGL's CMake. https://github.com/lvgl/lvgl/blob/master/env_support/cmake/esp.cmake#L32

What issues you see in the lv_demo_benchmark demo?

espzav: Thank you :-)

kisvegabor:
cc @ulipe

Looks great and very interesting! Two questions:
1. The benchmark was committed by @github-action. Do you run the benchmark with CI somehow? 
2. What does the subscript numbers (e.g. `(+17)`) mean?

espzav:
> cc @uLipe
> 
> Looks great and very interesting! Two questions:
> 
> 1. The benchmark was committed by @GitHub-Action. Do you run the benchmark with CI somehow?
> 2. What does the subscript numbers (e.g. `(+17)`) mean?

Hi @kisvegabor,
1. We have a runner for all boards (called BSP Wall Runner). It is for testing and we can run the benchmark example there (when PR is labeled - Run benchmark). Here is pytest for it: https://github.com/espressif/esp-bsp/blob/master/examples/display_lvgl_benchmark/pytest_display_lvgl_benchmark.py
2. The numbers are difference between master. In every push to master is benchmark run again and saved latest values to releases (https://github.com/espressif/esp-bsp/releases/tag/benchmark-latest). Each PR can compare the benchmark values with master.

PR #509, opened 2025-02-12
fix(lcd_touch): avoid assigning a value to handle after creation failure

Lzw655:

- [ ] CI passing

# Change description

Taking the `esp_lcd_touch_ft5x06` component as an example, when `esp_lcd_touch_new_i2c_ft5x06()` fails, `out_touch` is still assigned a freed memory address. 

This PR fixes the issue by assigning the value before handling the `err:` label.

### DIFF
--- a/components/lcd_touch/esp_lcd_touch_ft5x06/esp_lcd_touch_ft5x06.c
+++ b/components/lcd_touch/esp_lcd_touch_ft5x06/esp_lcd_touch_ft5x06.c
@@ -153,6 +153,8 @@ esp_err_t esp_lcd_touch_new_i2c_ft5x06(const esp_lcd_panel_io_handle_t io, const
     ret = touch_ft5x06_init(esp_lcd_touch_ft5x06);
     ESP_GOTO_ON_ERROR(ret, err, TAG, "FT5x06 init failed");
 
+    *out_touch = esp_lcd_touch_ft5x06;
+
 err:
     if (ret != ESP_OK) {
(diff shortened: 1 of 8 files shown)


components/lcd_touch/esp_lcd_touch_ft5x06/esp_lcd_touch_ft5x06.c
```diff
@@ -161,8 +163,6 @@ esp_err_t esp_lcd_touch_new_i2c_ft5x06(const esp_lcd_panel_io_handle_t io, const
         }
```
espzav, line 163: I think, the best solution is, assign `NULL` when failed:
```
*out_touch = NULL;
```

Lzw655: I believe we should not modify the user-provided address when an error occurs. The return value allows the user to fully determine success or failure.

tore-espressif (approved):
@Lzw655 LGTM. if you want the change to be released, please bump the versions :)

Lzw655:
> @Lzw655 LGTM. if you want the change to be released, please bump the versions :)

Thanks for the reminder! And I just found another issue, please check the modifications again. https://github.com/espressif/esp-bsp/pull/509/commits/5882eb0b7143c03da8eaaf54bc98e34076d0e1b3 @tore-espressif @espzav

espzav (approved):
@Lzw655 Thank you for fix these issues. LGTM

PR #513, opened 2025-02-21
feature(ili9341): Add alternative driver configuration for LCD ILI9341

vvranjek:
This is an alternative driver option that works with some LCD displays. The same as in TFT_eSPI library here: https://github.com/Bodmer/TFT_eSPI/blob/5793878d24161c1ed23ccb136f8564f332506d53/TFT_Drivers/ILI9341_Init.h#L127

### DIFF
--- a/components/lcd/esp_lcd_ili9341/README.md
+++ b/components/lcd/esp_lcd_ili9341/README.md
@@ -4,19 +4,22 @@
 
 Implementation of the ILI9341 LCD controller with esp_lcd component.
 
-| LCD controller | Communication interface | Component name | Link to datasheet |
-| :------------: | :---------------------: | :------------: | :---------------: |
-| ILI9341        | SPI                     | esp_lcd_ili9341     | [Specification](https://cdn-shop.adafruit.com/datasheets/ILI9341.pdf) |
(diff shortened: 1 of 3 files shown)


espzav:
I think, this cannot be part of driver. User can use `vendor_config` for change initialization commands. It can by part of readme file or something like that.

espzav:
@vvranjek please, did you see my latest comment?

vvranjek:
@espzav  It's been a long time since I was working with this, I just wanted to contribute my findings. Here is my view on this: users probably can use vendor_config to use custom drivers config, but this is nothing custom, the LCD that I got works out-of-the-box with the code that I provided in PR, not with original driver. This is the driver for ili9341, but ili9341 apparently is being manufactured with different specs/config, thus the need for different/alternative drivers. Since we don't know any manufacturing codes or part numbers that would tell us which specific hardware it is, there is no other option but to have alternative versions of the driver and the user must use trial and error to find out which one works for them. I spent days figuring this out, that is why I shared the code. Arduino libs use the same approach with alternative driver.

espzav:
Hi @vvranjek, we decided make it a little different - this PR: #687

PR #514, opened 2025-02-24
ci(runner): Fixed build and run CI when push to master

espzav:

- [x] CI passing

# Change description
- Sorry for issue CI when pushed to master

### DIFF
--- a/.github/workflows/build-run-applications.yml
+++ b/.github/workflows/build-run-applications.yml
@@ -75,23 +75,26 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Get changed files
+        if: github.event_name == 'pull_request' # This action is working only in pull-request (it fails in push to master)
         id: changed-files
         uses: tj-actions/changed-files@v45
         with:
           separator: ';' # idf-build-apps expects files seprated with semicolon
 
       - name: Build ESP-BSP applications
+        if: always()
         shell: bash
         env:
           IDF_EXTRA_ACTIONS_PATH: "${{ github.workspace }}/examples"
-          ALL_CHANGED_FILES: ${{ steps.changed-files.outputs.all_modified_files }}
+          # When event not a pull-request, the action before must be skipped and the steps.changed-files.outputs.all_modified_files variable is not exists and must be removed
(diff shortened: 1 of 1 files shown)


espzav:
@tore-espressif I hope that it will be working on master now :-) PTAL

.github/workflows/build-run-applications.yml
```diff
@@ -75,23 +75,25 @@ jobs:
...
         run: |
           . ${IDF_PATH}/export.sh
           pip install idf-component-manager==1.* ruamel.yaml idf-build-apps==2.4.3 --upgrade
 
-          echo "Files changed: ${ALL_CHANGED_FILES}"
-          idf-build-apps find --modified-files "${ALL_CHANGED_FILES}"
-          idf-build-apps build --parallel-count ${{ matrix.parallel_count }} --parallel-index ${{ matrix.parallel_index }} --modified-files "${ALL_CHANGED_FILES}" --collect-app-info build_info_${{ matrix.idf_ver }}_${{ matrix.parallel_index }}.json
+          echo "Files changed: "${{env.ALL_CHANGED_FILES}}
+          idf-build-apps find ${{env.ALL_CHANGED_FILES}}
+          idf-build-apps build --parallel-count ${{ matrix.parallel_count }} --parallel-index ${{ matrix.parallel_index }} ${{env.ALL_CHANGED_FILES}} --collect-app-info build_info_${{ matrix.idf_ver }}_${{ matrix.parallel_index }}.json
```
tore-espressif, line 96: @espzav Thank you very much for looking into this!

There might be way how to make this easier:

There is a `--include-skipped-apps` option that will essentially 'ignore' the `--modified-files` option. This way, for merges to `master` we would only add the `--include-skipped-apps` option.

https://docs.espressif.com/projects/idf-build-apps/en/latest/references/cli.html#named-arguments

What do you think?

espzav: I am fixing this error: https://github.com/espressif/esp-bsp/actions/runs/13495261465/job/37701002108 and it seems I need to skip "Get changed files" action. Then the variable `steps.changed-files.outputs.all_modified_files` not exists and I had to remove the `--modified-files=`. 
I am not sure if I can do it easier.

tore-espressif: Ok, I see now. Could you then please add more comments to the file? In case someone wants to improve/modify the workflow it might be difficult to understand why this code is there...

espzav: added - is it ok?

tore-espressif (approved):
Thanks!

PR #524, opened 2025-02-27
fix(esp_lvgl_port): Fix draw buffer alignment

tore-espressif:

# Change description
- Honor draw buffer alignment from LVGL config
- Speed up benchmark. WIth refresh period = 33ms (default) we cannot get better than 30 FPS in the tests!


espzav (approved):
@tore-espressif Thank you for this change, it was good idea to use function `heap_caps_aligned_alloc`. It might help in RGB888 align for DMA (I hope).
LGTM

PR #532, opened 2025-03-04
ci: Update runners image to ubuntu-24.04

tore-espressif:
Previously used ubuntu-20.04 will be deprecated by 1.4.2025.

https://github.com/actions/runner-images/issues/11101


espzav (approved):
@tore-espressif Thank you, LGTM

PR #541, opened 2025-03-14
feat(bsp): New SD card API (proposal) (BSP-189)

### LINKED ISSUE #149: Allow mounting the SD card with SDMMC_FREQ_HIGHSPEED (BSP-286)
I would like to be able to mount the SD card on ESP32-S3-EYE with `host.max_freq_khz = SDMMC_FREQ_HIGHSPEED`.

Currently I have to copy the SD card mounting code from the BSP component into the app just to modify this setting.

Could we consider either or both solutions:

1. Add a way to change the SD card frequency setting, either via Kconfig or a run-time parameter.
2. Always set `max_freq_khz = SDMMC_FREQ_HIGHSPEED` on boards where we know this works. (But still provide a way to fall back to a lower frequency, same as 1.)

espzav:

- [x] CI passing

# Change description
- This is the proposal of the new API for SD card. There can be used MMC mode or SPI.
- Initialization is same and there is not breaking change.
- Only one breaking change: 
- - Removed `extern sdmmc_card_t *bsp_card;` --> New function `bsp_sdcard_get_handle()`

Closes #149

### DIFF
--- a/bsp/esp32_p4_function_ev_board/esp32_p4_function_ev_board.c
+++ b/bsp/esp32_p4_function_ev_board/esp32_p4_function_ev_board.c
@@ -41,8 +41,10 @@ static const char *TAG = "ESP32_P4_EV";
 static lv_indev_t *disp_indev = NULL;
 #endif // (BSP_CONFIG_NO_GRAPHIC_LIB == 0)
 
-sdmmc_card_t *bsp_sdcard = NULL;    // Global uSD card handler
+static sdmmc_card_t *bsp_sdcard = NULL;    // uSD card handle
 static bool i2c_initialized = false;
+static bool spi_sd_initialized = false;
+static sd_pwr_ctrl_handle_t pwr_ctrl_handle = NULL; //SD LDO handle
 static TaskHandle_t usb_host_task;  // USB Host Library task
 #if (ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 3, 0))
 static i2c_master_bus_handle_t i2c_handle = NULL;  // I2C Handle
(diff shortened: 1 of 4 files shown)


bsp/esp32_p4_function_ev_board/include/bsp/esp32_p4_function_ev_board.h
```diff
@@ -218,7 +226,39 @@ esp_err_t bsp_spiffs_unmount(void);
...
 #define BSP_SD_MOUNT_POINT      CONFIG_BSP_SD_MOUNT_POINT
-extern sdmmc_card_t *bsp_sdcard;
+
+/* SD card is connected to Slot 0 pins. Slot 0 uses IO MUX, so not specifying the pins here */
+#define BSP_SDCARD_MMCSLOT_DEFAULT() {\
+    .cd = SDMMC_SLOT_NO_CD,\
+    .wp = SDMMC_SLOT_NO_WP,\
+    .width = 4,\
+    .flags = 0,\
+}
```
igrr, line 236: What about defining a function like `bsp_get_sdmmc_slot_config(slot, &config)`, which would fill the `sdmmc_slot_config_t` structure? This would be future-proof in case a board has two SD card slots.

igrr (commented):
I think the most valuable part of SD card support in BSPs is not related to the helper functions for mounting the card, but with the functions/macros to get the correct slot and host configs.

When the developer first starts the project, the requirements for SD card could be very simple: mount the card, return an error on any failure. An all-in-one helper function is okay for this.

Later, the requirements for mounting the card will likely become more complex. The application needs to handle the failure to mount the card depending on what exactly has failed. (Unreadable card? Tell the user to replace the card. Corrupted filesystem? Ask the user if the card should be formatted. etc...)

In this scenario, it doesn't make sense for the BSP to support all these possible cases. What we need from the BSP is to let the application know the correct host and slot configs for the given board, and the application can use these configs to implement the more complex card mounting logic.

I think defining _functions_ to return host and slot configs is better than defining macros (like BSP_SDCARD_SPISLOT_DEFAULT). First, if a function returns `sdspi_device_config_t`, this automatically explains which structure it is supposed to fill. With a macro, there is no indication to which structure the macro is related to.

bsp/esp32_p4_function_ev_board/include/bsp/esp32_p4_function_ev_board.h
espzav: Thank you.
OK, I defined macros due to memory. We can use static variables inside these functions, which will be in memory only after call the function. I will change it.

espzav: @igrr I updated it with your ideas.

bsp/esp32_p4_function_ev_board/esp32_p4_function_ev_board.c
```diff
@@ -112,21 +112,73 @@ i2c_master_bus_handle_t bsp_i2c_get_handle(void)
...
+    return bsp_sdcard;
+}
+
+sdmmc_host_t *bsp_sdcard_get_host(void)
+{
+    static sdmmc_host_t host_config = SDMMC_HOST_DEFAULT();
+    host_config.slot = SDMMC_HOST_SLOT_0;
+    host_config.max_freq_khz = SDMMC_FREQ_HIGHSPEED;
+
+    return &host_config;
```
igrr, line 126: Not sure this is a good idea to return the address of the static variable. The caller may want to change the flags, and this will modify the static variable.

Can you instead let the caller declare the config and pass the pointer into this function? This function can then fill the structure. After that the caller can still modify some fields. There will be no static RAM usage this way.

espzav: Thank you, you are right. Other functions can stay as they are, or should they be same?

igrr: I think bsp_sdcard_sdspi_get_slot should be changed the same way, because you are modifying the structure depending on the spi_host argument.

espzav: Then I'll change everything so that the API is the same :-)

igrr:
LGTM as far as the API goes!

bsp/esp32_p4_function_ev_board/esp32_p4_function_ev_board.c
```diff
@@ -112,21 +112,75 @@ i2c_master_bus_handle_t bsp_i2c_get_handle(void)
...
+    assert(config);
+
+    sdmmc_host_t host_config = SDMMC_HOST_DEFAULT();
+    host_config.slot = slot;
+    host_config.max_freq_khz = SDMMC_FREQ_HIGHSPEED;
+
+    memcpy(config, &host_config, sizeof(sdmmc_host_t));
+}
+
+void bsp_sdcard_sdmmc_get_slot(const int slot, sdmmc_slot_config_t *config)
```
tore-espressif, line 131: `slot` parameter is not used. Can we remove it? Or is this for future extensions where we do not want to use Slot0?

igrr: That was my suggestion (https://github.com/espressif/esp-bsp/pull/541#discussion_r1995665076) — just to keep the API the same for the boards which may have two SD card slots.

(I'm open to suggestions on how else to address this.)

tore-espressif: Fair enough, thanks!

bsp/esp32_p4_function_ev_board/esp32_p4_function_ev_board.c
```diff
@@ -112,21 +112,75 @@ i2c_master_bus_handle_t bsp_i2c_get_handle(void)
...
+{
+    assert(config);
+    memset(config, 0, sizeof(sdspi_device_config_t));
```
tore-espressif, line 134:
```suggestion
    memset(config, 0, sizeof(sdmmc_slot_config_t));
```

tore-espressif (approved):

PR #551, opened 2025-03-26
ci(runner): Fix benchmark results releases

espzav:

- [ ] CI passing

# Change description
- I found my issue in CI - new benchmarks was not saved into releases

Reverted (I am not sure with this change):
- Changed `on pull_request` to `on pull_request_target` (https://github.com/thollander/actions-comment-pull-request?tab=readme-ov-file#permissions)


tore-espressif (approved):

espzav:
I removed change on pull_request to on pull_request_target. Because there can be security risk from forks. I must read about it more and make decision in future.

PR #571, opened 2025-05-09
fix(esp_lvgl_port): Register encoder button callbacks only if encoder_enter handle is not null

mpmarfisi:
# Change description
When registering an encoder an encoder using `lvgl_port_add_encoder(const lvgl_port_encoder_cfg_t * encoder_cfg`, if `encoder_cfg->encoder_enter` is `NULL`, button callbacks are still registered. This can lead to unintended behavior if no "enter" handler is expected.

### DIFF
--- a/components/esp_lvgl_port/idf_component.yml
+++ b/components/esp_lvgl_port/idf_component.yml
@@ -1,4 +1,4 @@
-version: "2.6.0"
+version: "2.6.1"
 description: ESP LVGL port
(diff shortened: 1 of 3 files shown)


espzav (commented):
@mpmarfisi Thank you for this fix. LGTM
Please remove version change (we will release `esp_lvgl_port` later with updated CHANGELOG file)

mpmarfisi:
@espzav done!

PR #579, opened 2025-06-04
ci(build): Build and run all on master every midnight + fixed benchmarks on master

espzav:
- [x] CI passing

# Change description
- I decided to run build+run CI every day at midnight. We need to check if any other changes in components or IDF don't break our code.
- Fixed benchmarks - removed diff on master


espzav:
@tore-espressif PTAL, it is very simple update :-)

PR #585, opened 2025-06-18
feat(icm42670): Expose functions for reading/writing registers

johnboiles:
This PR exposes methods for reading/writing registers with the ICM42670 accelerometer. This enables end-users to use more advanced functionality like configuring interrupts.

### DIFF
--- a/components/icm42670/icm42670.c
+++ b/components/icm42670/icm42670.c
@@ -11,6 +11,7 @@
 #include <sys/time.h>
 #include "esp_system.h"
 #include "esp_check.h"
+#include "esp_timer.h"
 #include "icm42670.h"
 
 #define I2C_CLK_SPEED 400000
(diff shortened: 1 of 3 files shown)


components/icm42670/icm42670.c
```diff
@@ -414,3 +415,71 @@ esp_err_t icm42670_complimentory_filter(icm42670_handle_t sensor, const icm42670
...
+        ESP_LOGE(TAG, "Invalid MREG value %d", mreg);
+        return ESP_ERR_INVALID_ARG;
+    }
+    ESP_RETURN_ON_ERROR(icm42670_write(sensor, ICM42670_BLK_SEL_R, &blk_sel_r, 1), TAG,
+                        "Failed to set BLK_SEL_R");
+    ESP_RETURN_ON_ERROR(icm42670_write(sensor, ICM42670_MADDR_R, &reg, 1), TAG,
+                        "Failed to set MADDR_R");
+
+    int64_t start = esp_timer_get_time();
+    while (esp_timer_get_time() - start < 10000) {
```
espzav: Could you use this `vTaskDelay(pdMS_TO_TICKS(10));` instead of the while loop?

components/icm42670/icm42670.c
```diff
@@ -414,3 +415,71 @@ esp_err_t icm42670_complimentory_filter(icm42670_handle_t sensor, const icm42670
...
+    }
+
+    ESP_RETURN_ON_ERROR(icm42670_read(sensor, ICM42670_M_R, val, 1), TAG, "Failed to read M_R");
+
+    start = esp_timer_get_time();
+    while (esp_timer_get_time() - start < 10000) {
```
espzav: Could you use this vTaskDelay(pdMS_TO_TICKS(10)); instead of the while loop?

components/icm42670/icm42670.c
```diff
@@ -414,3 +415,71 @@ esp_err_t icm42670_complimentory_filter(icm42670_handle_t sensor, const icm42670
...
+        return ESP_ERR_INVALID_ARG;
+    }
+    ESP_RETURN_ON_ERROR(icm42670_write(sensor, ICM42670_BLK_SEL_W, &blk_sel_w, 1), TAG,
+                        "Failed to set BLK_SEL_W");
+    ESP_RETURN_ON_ERROR(icm42670_write(sensor, ICM42670_MADDR_W, &reg, 1), TAG,
+                        "Failed to set MADDR_W");
+    ESP_RETURN_ON_ERROR(icm42670_write(sensor, ICM42670_M_W, &val, 1), TAG, "Failed to set M_W");
+
+    int64_t start = esp_timer_get_time();
+    while (esp_timer_get_time() - start < 10000) {
```
espzav: Could you use this vTaskDelay(pdMS_TO_TICKS(10)); instead of the while loop?

espzav (commented):
@johnboiles thank you for this PR. I left some notes.
I am not sure, but when user change `mreg`, then will not work other API functions, because they are using default `mreg`. Am I right?

components/icm42670/icm42670.c
johnboiles: My mistake, this is supposed to be >10µs according to the datasheet. The spin loop is probably more correct for something that short right? 
image

johnboiles:
> I am not sure, but when user change mreg, then will not work other API functions, because they are using default mreg.

From the relevant datasheet page, you can interact with other registers after changing `MREG*` registers, but you must wait at least 10µs. In my code, I set `USER BANK 0` registers after configuring `MREG1` and it seems to work fine.
image

components/icm42670/icm42670.c
johnboiles: see above

components/icm42670/icm42670.c
johnboiles: see above

components/icm42670/icm42670.c
espzav: For 10 us, it is ok. Thank you for clarification.

components/icm42670/icm42670.c
```diff
@@ -414,3 +415,71 @@ esp_err_t icm42670_complimentory_filter(icm42670_handle_t sensor, const icm42670
...
+        return ESP_ERR_INVALID_ARG;
+    }
+    ESP_RETURN_ON_ERROR(icm42670_write(sensor, ICM42670_BLK_SEL_R, &blk_sel_r, 1), TAG,
+                        "Failed to set BLK_SEL_R");
+    ESP_RETURN_ON_ERROR(icm42670_write(sensor, ICM42670_MADDR_R, &reg, 1), TAG,
+                        "Failed to set MADDR_R");
+
+    int64_t start = esp_timer_get_time();
+    while (esp_timer_get_time() - start < 10) {
+    }
```
tore-espressif: Could you consider esp_rom_delay_us()? This will be even more readable, plus we will not have to add the `esp_timer` dependency to CMakeLists.txt
```c
    #include "esp_rom_sys.h"
    ...
    esp_rom_delay_us(10);
```

johnboiles: That's a great idea, I didn't know about that function. I've made the change.

espzav:
@johnboiles The issue from failed tests was fixed on master, please rebase master to your branch, bump the version of a vomponent and squash all commits.

johnboiles:
> please rebase master to your branch, bump the version of a component and squash all commits.

Done!

PR #595, opened 2025-07-01
feat(ds18b20): support to trigger multi sensors conversion (BSP-693)

### LINKED ISSUE #200: [DS18B20] Feature Request: Trigger temperature conversion on all devices (BSP-348)
When working with more than one DS18B20 temperature sensors, three in my case, it would be nice to call `ds18b20_trigger_temperature_conversion()` on all three and then only delay for the required amount before reading the temperatures from all three. 

The current implementation forces me to wait `x ms` for each sensor where ideally I could only wait once.

I would imagine an function defined as follow:
```
esp_err_t ds18b20_trigger_temperature_conversions(const ds18b20_device_handle_t *ds18b20, uint8_t ds18b20_size );
```

Kainarx:
Closes #200


Kainarx:
@suda-morris  PTAL

arc12:
This would be a useful change.
Please resolve the conflicts and merge.
Thanks

suda-morris (approved):

PR #598, opened 2025-07-04
ci(component_upload): Upload all examples to each BSP with right YML files.

espzav:

- [x] CI passing

# Change description
- Allow to add one example to multiple BSPs 
  - it will change YML for specific BSP 
  - it will copy specific `sdkconfig.bsp.`* to `sdkconfig.defaults`

### DIFF
--- a/.github/ci/bsp_example_update.py
+++ b/.github/ci/bsp_example_update.py
@@ -1,0 +1,97 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: 2025 Espressif Systems (Shanghai) CO LTD
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+import sys
+import yaml
+import argparse
+import shutil
+
+bsp_dir = "bsp"
+bsps = {name for name in os.listdir(bsp_dir) if os.path.isdir(os.path.join(bsp_dir, name))}
+
+
+def load_yaml_file(filepath):
+    with open(filepath, "r") as f:
+        return yaml.safe_load(f)
+
+
+def save_yaml_file(filepath, data):
+    with open(filepath, "w") as f:
+        yaml.dump(data, f, default_flow_style=False, sort_keys=False)
+
+
+def bsp_short_name(bsp):
+    return bsp.split('/')[-1]
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Update examples for specific BSP.")
+    parser.add_argument("--bsp", required=True, help="Name of the BSP component (directory under bsp/)")
+    args = parser.parse_args()
+    target_bsp = args.bsp
+    if not target_bsp:
+        print("Please set --bsp argument")
+        sys.exit(1)
+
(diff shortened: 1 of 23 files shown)


.github/workflows/upload_component.yml
```diff
@@ -10,18 +10,56 @@ on:
...
+      - name: Upload BSP components
+        uses: espressif/upload-components-ci-action@v1
+        with:
+          directories: >
+            bsp/${{ matrix.bsp }}
+          namespace: "espressif"
+          api_token: ${{ secrets.IDF_COMPONENT_API_TOKEN }}
+          dry_run: ${{ github.ref_name != 'master' || github.repository_owner != 'espressif' }}
+
+      - name: Upload Other components
```
XDanielPaul, line 59: Suggestion:

I believe the `Upload Other components` step is going to be triggered for each bsp matrix option (e.g. 21x).

A solution would be to extract it into a separate job

.github/workflows/upload_component.yml
```diff
@@ -10,18 +10,56 @@ on:
...
+          m5_atom_s3,
+          m5dial,
+        ]
     steps:
       - uses: actions/checkout@master
 
-      - name: Upload components to component service
+      - name: Prepare BSP examples
+        run: |
+          pip install idf-component-manager==1.* pyyaml
```
XDanielPaul: Question:

Is `idf-component-manager` dependency needed here?

.github/ci/bsp_example_update.py
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: 2025 Espressif Systems (Shanghai) CO LTD
+# SPDX-License-Identifier: Apache-2.0
+
+import os
```
XDanielPaul, line 5: Nit:

I recommend using Pathlib, but as this is just a CI script I believe it's okay

.github/workflows/upload_component.yml
```diff
@@ -10,18 +10,56 @@ on:
...
+          python .github/ci/bsp_example_update.py --bsp=${{ matrix.bsp }}
+
+      - name: Upload BSP components
+        uses: espressif/upload-components-ci-action@v1
```
XDanielPaul, line 47: Suggestion:

You may consider updating to `upload-components-ci-action` to v2

XDanielPaul:
@espzav I had a few comments, but otherwise looks good to me! 🚀 :)

.github/workflows/upload_component.yml
espzav: You are right, thanks. I changed.

.github/workflows/upload_component.yml
espzav: I am not sure, removed :-)

.github/ci/bsp_example_update.py
espzav: ok, thank you, I keep it as it is now.

PR #600, opened 2025-07-08
fix(component_upload): Fix params for v2 upload-component

espzav:

- [ ] CI passing

# Change description


espzav:
> @espzav Thanks! I thought that this was already fixed on `upload-components-ci-action` ...

There was another issue and this is only warning, that these variables are deprecated and should be used new ones. It was recommended to change it.

PR #614, opened 2025-07-30
Refactor/set rgb element order

suda-morris:
IDF recommend to use rgb_ele_order to specify the RGB element order. 

In v4.x, we use "color_space" to distinguish RGB and BGR, but in fact they're the same color space. Thus we deprecate that name.
In v5.x, we use "rgb_endian" to distinguish RGB and BGR, but in practice, users are confused by the "rgb_endian" and the "data_endian" (a new config). Thus we start to recommend to use "rgb_ele_order".


espzav (approved):
@suda-morris Thank you for this PR! LGTM

PR #637, opened 2025-09-14
refactor(mipi_lcd): set phy_clk_src to 0 as default clock source

suda-morris:
The dphy pll reference clock got changed in esp32p4 version 3.0. The real default clock source are different between major versions.

But using `0` as default can work for all P4 version, the driver will select a default clock source automatically.


espzav (approved):
@suda-morris Thank you! LGTM

PR #667, opened 2025-10-27
ci(build): Update idf-build-apps to 2.10.1

espzav:

- [x] CI passing

# Change description
- Update idf-build-apps version to same with Launchpad CI
- Synchronize IDF versions Launchpad and build CI
- Fixed issues


.github/workflows/build-run-applications.yml
```diff
@@ -98,7 +98,7 @@ jobs:
           ALL_CHANGED_FILES: ${{ (github.event_name != 'pull_request') && format(' ') || format('--modified-files=') }}${{ (github.event_name != 'pull_request') && format(' ') || steps.changed-files.outputs.all_modified_files }}
         run: |
           . ${IDF_PATH}/export.sh
-          pip install idf-component-manager==2.2.* ruamel.yaml idf-build-apps==2.4.3 --upgrade
+          pip install idf-component-manager==2.2.* ruamel.yaml idf-build-apps==2.10.1 --upgrade
```
PetrESP: Suggestion:
Maybe we could specify  version of `idf-component-manager` and `idf-build_apps` as an environment variable next to the ESP-IDF matrix? That way we could have a single place which specifies tool versions.

espzav: Thank you, good idea. I changed it :-)

.github/workflows/build-run-applications.yml
```diff
@@ -29,6 +29,9 @@ concurrency:
 
 jobs:
   build:
+    env:
+      component_manager_ver: "2.2.*"
+      idf_build_apps_ver: "2.10.1"
```
tore-espressif: can we do latest stable 2.13.3? https://pypi.org/project/idf-build-apps/

espzav: I want to have same version as using Launchpad: https://github.com/espressif/idf-examples-launchpad-ci-action/blob/5f94d3fb3550db06bbb0506c2b219359b1f60816/action.yml#L37
Because sometimes PR CI pass and it crashes on master, when building Launchpad.

PR #674, opened 2025-11-10
fix(ci): Fix jira issue sync and set checkout action to a specific ve…

PetrESP:

- [ ] CI passing

# Change description
Fixed jira sync action version from master (which doesn't exist) to `v1`
Specified version of checkout action which is safer than using master


PetrESP:
Link to a failing jira sync job: https://github.com/espressif/esp-bsp/actions/runs/19232686231

PR #687, opened 2025-12-02
feat(lcd_ili9341): Selectable default initial commands for LCD from three presets.

espzav:

- [x] CI passing

# Change description
- Based on PR: #513 - Closes #513


espzav:
@tore-espressif @PetrESP PTAL

components/lcd/esp_lcd_ili9341/esp_lcd_ili9341.c
```diff
@@ -41,6 +41,7 @@ typedef struct {
     uint8_t fb_bits_per_pixel;
     uint8_t madctl_val; // save current value of LCD_CMD_MADCTL register
     uint8_t colmod_val; // save current value of LCD_CMD_COLMOD register
+    ili9341_lcd_init_cmds_t init_cmds_sel;
```
PetrESP: For the purpose of future maintainability: Are we sure that another API is really needed and will be used often? Can't we just use KConfig instead? Which would also get rid of the switch which selects the default optioni

components/lcd/esp_lcd_ili9341/esp_lcd_ili9341.c
```diff
@@ -183,7 +186,7 @@ static esp_err_t panel_ili9341_reset(esp_lcd_panel_t *panel)
     return ESP_OK;
 }
 
-static const ili9341_lcd_init_cmd_t vendor_specific_init_default[] = {
+static const ili9341_lcd_init_cmd_t vendor_specific_init_default_1[] = {
```
PetrESP: Sorry for the nitpick but why not index from 0?

components/lcd/esp_lcd_ili9341/esp_lcd_ili9341.c
```diff
@@ -231,6 +234,65 @@ static const ili9341_lcd_init_cmd_t vendor_specific_init_default[] = {
...
+    {0xE0, (uint8_t []){0x0F, 0x31, 0x2B, 0x0C, 0x0E, 0x08, 0x4E, 0xF1, 0x37, 0x07, 0x10, 0x03, 0x0E, 0x09, 0x00}, 15, 0}, // Adjusted for ILI9341_2_DRIVER
+    /* Negative gamma correction */
+    {0xE1, (uint8_t []){0x00, 0x0E, 0x14, 0x03, 0x11, 0x07, 0x31, 0xC1, 0x48, 0x08, 0x0F, 0x0C, 0x31, 0x36, 0x0F}, 15, 0}, // Adjusted for ILI9341_2_DRIVER
+    /* Entry mode set */
+    {0xB7, (uint8_t []){0x07}, 1, 0},
+    /* Display function control */
+    {0xB6, (uint8_t []){0x08, 0x82, 0x27}, 3, 0},
+};
+
+/* Init commands used in ESP-BOX-3 */
```
PetrESP: It is already well done using the `init_cmds` in the BOX 3 so why are we adding a functionality which copies a one which is already there?

components/lcd/esp_lcd_ili9341/esp_lcd_ili9341.c
espzav: In these drivers are not used usually KConfig. This is the reason, why I wanted to do it in runtime.

components/lcd/esp_lcd_ili9341/esp_lcd_ili9341.c
espzav: No reason, I started from 1 :-)

components/lcd/esp_lcd_ili9341/esp_lcd_ili9341.c
espzav: Because, when we added some other configuration, this is the right place to add more than one. Customer can use only this driver and can select some known configuration for his display. It can help to fast start woth this driver (without long search on the Internet)

PetrESP: But in that case I am missing information about which driver should be picked for which display. Do you plan to add some comments which would help with choosing the right preset?

components/lcd/esp_lcd_ili9341/esp_lcd_ili9341.c
PetrESP: Yeah I see that none of them use KConfig. I think that using it would also save on some flash memory and would overall make more sense because you will always only use one init preset so there is no reason to compile all of them.

espzav: Reworked by yesterday discussion.

components/lcd/esp_lcd_ili9341/esp_lcd_ili9341.c
espzav: Reworked by yesterday discussion.

components/lcd/esp_lcd_ili9341/README.md
```diff
@@ -64,3 +64,29 @@ Alternatively, you can create `idf_component.yml`. More is in [Espressif's docum
 ```
 
 There is an example in ESP-IDF with this LCD controller. Please follow this [link](https://github.com/espressif/esp-idf/tree/master/examples/peripherals/lcd/spi_lcd_touch).
+
+## Predefined init vendor commands
+
+This driver contains some predefined vendor inititalization commands. It can help improve screen in some cases. Usage is in example below.
```
PetrESP:
```suggestion
This driver contains various predefined vendor initialization commands which in some cases can improve image quality. Usage is in example below.
```
Tiny nitpick but I think that it clarifies why these commands are useful.

espzav: Thank you, udpated.

Issue #615, opened 2025-07-31
esp_io_expander: how to use with APIs with GPIO number parameter


dmartauz:
### Related area

esp_io_expander

### Hardware specification

custom ESP32-P4 board

### Is your feature request related to a problem?

In our application with ESP32-P4 as main host we need to use IO expander due to lack of GPIOs. We would like to use it for slow and rarely used signals such as RESET pins of external circuits, BOOT pins of external RCPs, etc.

Example - we would like to connect RESET pin of external Ethernet PHY to an IO expander.
https://docs.espressif.com/projects/esp-idf/en/latest/esp32p4/api-reference/network/esp_eth.html#create-mac-and-phy-instance
`phy_config.reset_gpio_num = CONFIG_EXAMPLE_ETH_PHY_RST_GPIO; // alter the GPIO used for PHY reset`
`eth_phy_config_t` struct does not allow to define callbacks, only GPIO pin number. 

### Describe the solution you'd like

It is possible to set pin number parameter to -1 and handle RESET pin of Ethernet PHY outside of Ethernet API but it would be better if GPIOs provided by IO expander could be used the same way as native GPIOs available on the SoC.

### I have checked existing list of Feature requests and the Contribution Guide

- [x] I confirm I have checked existing list of Feature requests and Contribution Guide.

tore-espressif:
@dmartauz Could you please check if this PR fixes your problem? https://github.com/espressif/esp-bsp/pull/703

Using GPIO API for IO expander feature was released in [`esp_io_expander` v 1.2.0](https://components.espressif.com/components/espressif/esp_io_expander/versions/1.2.0)

dmartauz:
@tore-espressif  thanks, need to prepare suitable hardware for testing. Is this going to work also for CS pins of SPI devices?

tore-espressif:
@dmartauz the SPI driver code does not use GPIO API, so this will not work out-of-the-box.

cc @songruo Is there a plan to control SPI's CS signal through IO expanders?

dmartauz:
@tore-espressif 
Based on the wrapper documentation following APIs are supported:
- gpio_set_level
- gpio_get_level
- gpio_set_direction
- gpio_set_pull_mode

In our application I would like to move RESET and BOOT signals of connected coprocessors to the IO expander. However, looking at esp-serial-flasher code it makes use of currently unsupported API:
```
// Initialize boot pin selection pins
gpio_reset_pin(s_reset_trigger_pin);
[...]

PR #703, opened 2026-01-07
feat(io_expander): add GPIO API wrapper for ESP IO Expander (BSP-764)

songruo:

- [x] CI passing

# Change description

Usage Example:

```
esp_io_expander_handle_t io_expander_handle = NULL;
if (esp_io_expander_new_i2c_xxx(0x20, I2C_NUM_0, GPIO_NUM_5, GPIO_NUM_6, &io_expander_handle) == ESP_OK && io_expander_handle) {
    // Map expander pins to virtual GPIO numbers starting at GPIO_NUM_MAX
    esp_gpio_append_expander_handler(io_expander_handle, GPIO_NUM_MAX);
} else {
    printf("Failed to initialize xxx IO Expander\n");
    return;
}

gpio_set_direction(31, GPIO_MODE_OUTPUT);
gpio_set_direction(39, GPIO_MODE_INPUT);

gpio_set_level(31, 1);
int level = gpio_get_level(39);
```

https://github.com/espressif/esp-idf/issues/17638

---

> [!NOTE]
> Introduces an optional GPIO API wrapper enabling IO expander pins to be controlled through ESP-IDF `gpio_*` APIs via virtual GPIO numbers.
> 
> - New `esp_io_expander_gpio_wrapper.c/.h`: link-time wrap of `gpio_set_level/get_level/set_direction/set_pull_mode`; dynamic mapping of virtual GPIO ranges to an `esp_io_expander_handle_t` with append/remove functions
> - New Kconfig `IO_EXPANDER_ENABLE_GPIO_API_WRAPPER` and README usage docs/examples
> - `CMakeLists.txt`: conditional `PRIV_REQUIRES` (`esp_driver_gpio` for IDF >= 5.3; `driver` otherwise) and wrapper source/linker flags when enabled
> - Component version bumped to `1.2.0` in `idf_component.yml`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4890915b019926a33342908f13835cd095185e36. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/components/io_expander/esp_io_expander/esp_io_expander_gpio_wrapper.c
+++ b/components/io_expander/esp_io_expander/esp_io_expander_gpio_wrapper.c
@@ -76,7 +76,7 @@ esp_err_t esp_io_expander_gpio_wrapper_append_handler(esp_io_expander_handle_t h
     uint32_t end = start_io_num + cnt;
     // Pre-allocate node before taking the lock to minimize time spent in critical section
     ioexp_range_node_t *node = (ioexp_range_node_t *)heap_caps_malloc(sizeof(ioexp_range_node_t),
-                                                                      MALLOC_CAP_INTERNAL | MALLOC_CAP_8BIT);
+                               MALLOC_CAP_INTERNAL | MALLOC_CAP_8BIT);
     if (node == NULL) {
         return ESP_ERR_NO_MEM;
     }


espzav:
Hi @songruo, @suda-morris Thank you for this PR. This is a good idea to do it like this. But I have one question: When I want to use some pin from IO expander in ESP LCD or other component like this, is it possible? I think, when we do it inside IDF, it will be possible - right? And I think this is the main reason, why we do it. 

For example this issue: https://github.com/espressif/esp-bsp/issues/615

@igrr @tore-espressif @PetrESP  What do you think about it?

songruo:
@espzav I think the key consideration is we don't want IDF to directly rely on external components. If users need to expand IOs, they will have to pull this `esp_io_expander` component in their application. Then having the GPIO API wrapper implementation at here seems to be a good choice.

components/io_expander/esp_io_expander/include/esp_io_expander.h
```diff
@@ -396,6 +397,31 @@ esp_err_t esp_io_expander_reset(esp_io_expander_handle_t handle);
...
+ * @brief Append a range of GPIOs additional to the existing embedded GPIO range
+ * 
+ * @param start_num: Start number of the appended range
+ * @param cnt: Count of the appended range
+ * @param handler: IO Expander handle
+ *
+ * @return
+ *      - ESP_OK: Success, otherwise returns ESP_ERR_xxx
+ */
+esp_err_t esp_io_expander_append_num(uint32_t start_num, uint32_t cnt, esp_io_expander_handle_t handler);
```
espzav: What about remove `cnt` and get it from `esp_io_expander_config_t - io_count`. It is safer, because it must be put whole scope of one IO expander, otherwise, the calculation of `pin_mask` will be broken.

components/io_expander/esp_io_expander/include/esp_io_expander.h
```diff
@@ -396,6 +397,31 @@ esp_err_t esp_io_expander_reset(esp_io_expander_handle_t handle);
...
+/**
+ * @brief Remove a range of already appended GPIOs
+ * 
+ * @param start_num: Start number of the removed range
+ * @param cnt: Count of the removed range
+ *
+ * @return
+ *      - ESP_OK: Success, otherwise returns ESP_ERR_xxx
+ */
+esp_err_t esp_io_expander_remove_num(uint32_t start_num, uint32_t cnt);
```
espzav: Is it possible to make it like this `esp_err_t esp_io_expander_remove_num(esp_io_expander_handle_t handler);` ? When we remove `cnt` from append function, it will be safer. 
Otherwise, this is dangerous, when you put cnt or start somewhere inside the scope. And it is too complicated to calculation.

songruo: Yes, changing to handler will make the calculation much simpler, but it is a bit weird to have an API called "remove_num" having an input parameter that is an IO expander handler. Maybe I could change the APIs to

`esp_gpio_append_expander_handler(esp_io_expander_handle_t handler, uint32_t start_io_num);`

`esp_gpio_remove_expander_handler(esp_io_expander_handle_t handler);`

espzav: Yes, the API names will be changed too, it make sense :-) You can use proposed API :-) Thank you!

components/io_expander/esp_io_expander/esp_io_expander_gpio_wrapper.c
```diff
@@ -0,0 +1,264 @@
...
+        }
+        prev = curr;
+        curr = curr->next;
+    }
+    portEXIT_CRITICAL(&s_ioexp_lock);
+    if (to_free) {
+        free(to_free);
+    }
+    return ESP_OK;
+}
```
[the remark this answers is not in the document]
songruo: Not very necessary?

components/io_expander/esp_io_expander/include/esp_io_expander.h
songruo: Updated! PTAL again @espzav, thanks!

components/io_expander/esp_io_expander/include/esp_gpio_wrapper.h
```diff
@@ -0,0 +1,43 @@
...
+#include "sdkconfig.h"
+#include "esp_io_expander.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#if CONFIG_IO_EXPANDER_ENABLE_GPIO_API_WRAPPER
+/**
+ * @brief Append a range of GPIOs additional to the existing embedded GPIO range
```
espzav: Update `@brief` for the new API.

components/io_expander/esp_io_expander/include/esp_gpio_wrapper.h
```diff
@@ -0,0 +1,43 @@
...
+ * @param handler: IO Expander handle
+ * @param start_io_num: Start GPIO number that maps to the IO expander range
+ *
+ * @return
+ *      - ESP_OK: Success, otherwise returns ESP_ERR_xxx
+ */
+esp_err_t esp_gpio_append_expander_handler(esp_io_expander_handle_t handler, uint32_t start_io_num);
+
+/**
+ * @brief Remove a range of already appended GPIOs
```
espzav: Update `@brief` for the new API.

espzav (commented):
Some nitpicks added. 
And please, could you bump the version of `io_expander` component to : "1.2.0"?
I am thinking of the name of the `esp_gpio_wrapper.c/h` Isn't it better to use something like `esp_io_expander_gpio.h` or `esp_io_expander_gpio_wrapper.h`? When some other component wants to use it too.

The best, it will be to add "how to use it" into all Readmes of IO expanders. And bumb the versions of IO expanders too - I can do it in another PR. I have example too, I will add it to `esp_io_expander_tca9554`.

espzav:
@songruo I made an example for TCA9554 (esp32_s3_korvo_2 board). We can use it :-) We can add it later, it depneds on you. 
[examples.zip](https://github.com/user-attachments/files/24528029/examples.zip)

espzav:
@songruo And I am so sorry for the CI - it is broken for now. We will solve it ASAP.

songruo:
Yah, I don't have an IO expander chip that is supported in the io_expander component, but I have tried with a MCP23017 chip, and it works fine with the gpio wrapper as well :)

components/io_expander/esp_io_expander/esp_io_expander_gpio_wrapper.c
```diff
@@ -0,0 +1,264 @@
...
+        break;
+    case GPIO_FLOATING:
+        pud = IO_EXPANDER_PULL_NONE;
+        break;
+    default:
+        pud_valid = false;
+    }
+    if (!handler->write_pullup_en_reg || (pud == IO_EXPANDER_PULL_DOWN && !handler->write_pullup_sel_reg)) {
+        pud_valid = false;
+    }
```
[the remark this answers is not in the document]
songruo: This is something might needs to be improved in `esp_io_expander_set_pullupdown`. There are some chips that only supports pull-up, no pull-downs, so one register is enough (according to AI...).

espzav (approved):
@songruo Thank you very much for your work! LGTM

Please, could you update pre-commit again (latest changes in repository changed some rules for pre-commit). Thank you

components/io_expander/esp_io_expander/esp_io_expander_gpio_wrapper.c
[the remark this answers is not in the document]
songruo: Likewise, I think `esp_io_expander_set_pullupdown` should handle the case of `IO_EXPANDER_PULL_NONE` to some chips where `pull_en_reg` and `pull_sel_reg` are both not supported.

PR #721, opened 2026-02-10
feat(bsp_selector): Make all BSPs public and add NoGlib selection.

espzav:

- [x] CI passing

# Change description
- Make all BSPs in `bsp_selector` public - fix for using this component in another component.
- Added NoGlib selection

### DIFF
--- a/components/bsp_selector/Kconfig
+++ b/components/bsp_selector/Kconfig
@@ -101,7 +101,17 @@ menu "BSP Selector"
         int
         default 1 if BSP_SELECT_ESP_VOCAT
         default 0
+    
+    config BSP_SELECT_GLIB_SEL
+        int
+        default 1 if BSP_SELECT_GLIB
+        default 0
         
+    config BSP_SELECT_GLIB
+        bool "Use GLIB version (with LVGL)"
+        default y
+        help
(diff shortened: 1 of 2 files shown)


components/bsp_selector/Kconfig
```diff
@@ -117,21 +127,23 @@ menu "BSP Selector"
             bool "ESP_BSP_NONE"
             
         config BSP_SELECT_ESP_BSP_DEVKIT
+            depends on BSP_SELECT_GLIB
```
PetrESP: Is this necessary here? Maybe I am missing something

espzav: Good question. I forgot remove it. I added it to all BSPs, which haven't got _noglib version. But here it is not necessary. Fixed.

components/bsp_selector/Kconfig
```diff
@@ -120,18 +130,19 @@ menu "BSP Selector"
             bool "ESP_BSP_DEVKIT"
             
         config BSP_SELECT_ESP_BSP_GENERIC
+            depends on BSP_SELECT_GLIB && BSP_SELECT_GLIB
```
tore-espressif: Nitpick

```suggestion
            depends on BSP_SELECT_GLIB
```

PR #730, opened 2026-02-25
bsp(EchoEar): Created new BSP for EchoEar

espzav:

- [x] Project [README.md](../README.md) updated
- [x] Component contains idf_component.yml file with `url` field defined and overriden paths for local components
- [x] New BSP definitions added to [bsp_ext.py](../examples/bsp_ext.py)
- [x] BSP was added to [SquareLine](https://github.com/espressif/esp-bsp/tree/master/SquareLine/common)
- [x] BSP was added to [.pre-commit-config.yaml](.pre-commit-config.yaml) doxybook list
- [x] BSP was added to BSP Selector [Kconfig](components/bsp_selector/Kconfig) and [idf_component.yml](components/bsp_selector/idf_component.yml)
- [ ] _Optional:_ BSP was added to Runner ([pytest.ini](pytest.ini), [conftest.py](conftest.py), [build-run-applications.yml](.github/workflows/build-run-applications.yml))
- [x] CI passing

# Change description
- New BSP - EchoEar - Generated by BSP Generator
- New Private API for Touch buttons
- Updated SquareLine manifest with display shape
- Will be added to runner later (still waiting for new runner)

### DIFF
--- a/SquareLine/boards/v9/echoear/manifest.json
+++ b/SquareLine/boards/v9/echoear/manifest.json
@@ -5,9 +5,10 @@
 
  "screen_width": "360",
  "screen_height": "360",
+ "screen_circle": true,
  "screen_color_swap": false,
 
- "supported_lvgl_version": "9.1.*, 9.2.*",
+ "supported_lvgl_version": "9.1.*, 9.2.*, 9.3.*",
 
  "short_description": "EchoEar BSP for SquareLine Studio.",
  "long_description": "EchoEar is an intelligent AI development kit. It is suitable for voice interaction products that require large model capabilities, such as toys, smart speakers, and smart central control systems. The device is equipped with a 1.85-inch QSPI circular touch screen, dual microphone array, and supports offline voice wake-up and sound source localization algorithms. Combined with the large model capabilities provided by OpenAI，Xiaozhi AI, Gemini, etc., EchoEar can achieve full-duplex voice interaction, multimodal recognition, and intelligent agent control, providing a solid foundation for developers to create complete edge-side AI application experiences.",
--- a/SquareLine/boards/v9/m5dial/manifest.json
+++ b/SquareLine/boards/v9/m5dial/manifest.json
@@ -1,13 +1,14 @@
 {
     "name":"M5Dial",
-    "version":"1.0.2",
+    "version":"1.0.3",
     "mcu":"ESP32S3",
 
     "screen_width":"240",
     "screen_height":"240",
+    "screen_circle": true,
     "screen_color_swap":false,
 
-    "supported_lvgl_version":"9.1.*, 9.2.*",
+    "supported_lvgl_version":"9.1.*, 9.2.*, 9.3.*",
 
     "short_description":"As a versatile embedded development board, M5Dial integrates the necessary features and sensors for various smart home control applications. It features a 1.28-inch round TFT touchscreen, a rotary encoder, an RFID detection module, an RTC circuit, a buzzer, and under-screen buttons, enabling users to easily implement a wide range of creative projects.",
     "long_description":"As a versatile embedded development board, M5Dial integrates the necessary features and sensors for various smart home control applications. It features a 1.28-inch round TFT touchscreen, a rotary encoder, an RFID detection module, an RTC circuit, a buzzer, and under-screen buttons, enabling users to easily implement a wide range of creative projects.",
--- a/SquareLine/boards/v8/custom_waveshare_7inch/manifest.json
+++ b/SquareLine/boards/v8/custom_waveshare_7inch/manifest.json
@@ -5,6 +5,7 @@
 
     "screen_width":"800",
     "screen_height":"480",
+    "screen_circle": false,
     "screen_color_swap":true,
 
     "supported_lvgl_version":"8.2.0, 8.3.*",
--- a/SquareLine/boards/v8/esp32_c3_lcdkit/manifest.json
+++ b/SquareLine/boards/v8/esp32_c3_lcdkit/manifest.json
@@ -5,6 +5,7 @@
 
     "screen_width":"240",
     "screen_height":"240",
+    "screen_circle": true,
     "screen_color_swap":true,
 
     "supported_lvgl_version":"8.2.0, 8.3.*",
--- a/SquareLine/boards/v8/esp32_s2_kaluga_kit/manifest.json
+++ b/SquareLine/boards/v8/esp32_s2_kaluga_kit/manifest.json
@@ -5,6 +5,7 @@
 
     "screen_width":"320",
     "screen_height":"240",
+    "screen_circle": false,
     "screen_color_swap":true,
 
     "supported_lvgl_version":"8.2.0, 8.3.*",
--- a/SquareLine/boards/v8/esp32_s3_eye/manifest.json
+++ b/SquareLine/boards/v8/esp32_s3_eye/manifest.json
@@ -5,6 +5,7 @@
 
     "screen_width":"240",
     "screen_height":"240",
+    "screen_circle": false,
     "screen_color_swap":true,
 
     "supported_lvgl_version":"8.2.0, 8.3.*",
--- a/SquareLine/boards/v8/esp32_s3_korvo_2/manifest.json
+++ b/SquareLine/boards/v8/esp32_s3_korvo_2/manifest.json
@@ -5,6 +5,7 @@
 
     "screen_width":"320",
     "screen_height":"240",
+    "screen_circle": false,
     "screen_color_swap":true,
 
     "supported_lvgl_version":"8.2.0, 8.3.*",
--- a/SquareLine/boards/v8/esp32_s3_lcd_ev_board/manifest.json
+++ b/SquareLine/boards/v8/esp32_s3_lcd_ev_board/manifest.json
@@ -5,6 +5,7 @@
 
     "screen_width":"800",
     "screen_height":"480",
+    "screen_circle": false,
     "screen_color_swap":false,
 
     "supported_lvgl_version":"8.2.0, 8.3.*",
--- a/SquareLine/boards/v8/esp32_s3_usb_otg/manifest.json
+++ b/SquareLine/boards/v8/esp32_s3_usb_otg/manifest.json
@@ -5,6 +5,7 @@
 
     "screen_width":"240",
     "screen_height":"240",
+    "screen_circle": false,
     "screen_color_swap":true,
 
     "supported_lvgl_version":"8.2.0, 8.3.*",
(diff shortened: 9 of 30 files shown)


bsp/echoear/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+if("${IDF_VERSION_MAJOR}.${IDF_VERSION_MINOR}" VERSION_GREATER_EQUAL "5.3")
```
tore-espressif: This will never be false, since in idf_component.yml we have idf: ">=5.5".

SquareLine/boards/v9/m5dial/manifest.json
```diff
@@ -5,6 +5,7 @@
 
     "screen_width":"240",
     "screen_height":"240",
+    "screen_circle": false,
```
tore-espressif: m5 dial is circle?

```suggestion
    "screen_circle": true,
```

SquareLine/boards/v9/echoear/manifest.json
```diff
@@ -0,0 +1,19 @@
+{
+ "name": "EchoEar",
+ "version": "1.0.0",
+ "mcu": "ESP32S3",
+
+ "screen_width": "360",
+ "screen_height": "360",
+ "screen_color_swap": false,
+
+ "supported_lvgl_version": "9.1.*, 9.2.*",
```
tore-espressif: FYI: Latest SquareLine supports also LVGL 9.3. However, I am not sure if they keep updating our BSP packages, because I don't see our board supported for LVGL9.2 in squareline... we can discuss later

bsp/echoear/CMakeLists.txt
espzav: Updated.

SquareLine/boards/v9/m5dial/manifest.json
espzav: Good point, thanks, I overlooked. Fixed.

SquareLine/boards/v9/echoear/manifest.json
espzav: I updated for LVGL 9.3 and updated the version of the board. We will see in the next update od SquareLine.

bsp/echoear/src/echoear.c
```diff
@@ -0,0 +1,1043 @@
...
+    /* Create channels with default config */
+    touch_channel_config_t chan_cfg = {
+        .active_thresh = {2000},
+        .charge_speed = TOUCH_CHARGE_SPEED_7,
+        .init_charge_volt = TOUCH_INIT_CHARGE_VOLT_DEFAULT,
+    };
+    for (int i = 0; i < BSP_BUTTON_NUM; i++) {
+        if (bsp_button_config[i].type == BSP_BUTTON_TYPE_TOUCHPAD) {
+            uint8_t chan = bsp_button_config[i].cfg.custom.key;
+            BSP_ERROR_CHECK_RETURN_ERR(touch_sensor_new_channel(bsp_touchpad_sens_handle,
```
PetrESP: I think that we should deinit/free all allocated memory created by `touch_sensor_new_controller` when any subsequent function calls fail.

bsp/echoear/src/echoear.c
```diff
@@ -0,0 +1,1043 @@
...
+
+    /* Enable the touch sensor to do the initial scanning, so that to initialize the channel data */
+    BSP_ERROR_CHECK_RETURN_ERR(touch_sensor_enable(bsp_touchpad_sens_handle));
+
+    /* Scan the enabled touch channels for several times, to make sure the initial channel data is stable */
+    for (int i = 0; i < BSP_TOUCHPAD_RETRIES; i++) {
+        ESP_ERROR_CHECK(touch_sensor_trigger_oneshot_scanning(bsp_touchpad_sens_handle, 2000));
+    }
+
+    /* Disable the touch channel to rollback the state */
```
PetrESP: Is this still necessary? The touch sensor is disabled to apply configuration but we don't seem to be changing it. We only read the one shot data for which we should not need to stop the touch sensor

espzav:
Hi @tore-espressif @PetrESP @igrr I decided to divide SRC files in BSP by feature - what do you think? Check my latest commit in this PR.
WHY: As a discussion on Mattermost in the BSP channel, I am thinking about versioning feature files, when there are more board versions. It can be `bsp_display_v1.c` or something like that and in CMake allow you to add both files or only one by `menuconfig`. When there will be more version files, I am looking for some solution (something like Ivan proposes)

bsp/echoear/src/bsp_button.c
```diff
@@ -0,0 +1,206 @@
...
+    BSP_ERROR_CHECK_RETURN_ERR(touch_sensor_start_continuous_scanning(bsp_touchpad_sens_handle));
+
+    bsp_touchpad_sens_initialized = true;
+    ESP_LOGI(TAG, "Touch sensor initialized.");
+    return ESP_OK;
+}
+
+static esp_err_t  bsp_touchpad_custom_deinit(button_driver_t *button_driver)
+{
+    (void)button_driver;
```
PetrESP: Just a nitpick but I think that we can use `ESP_UNUSED` here if you'd like. But it is pretty obvious what you are trying to do here

bsp/echoear/src/bsp_button.c
```diff
@@ -0,0 +1,206 @@
...
+    }
+    touch_sensor_del_controller(bsp_touchpad_sens_handle);
+    bsp_touchpad_sens_handle = NULL;
+    return ESP_OK;
+}
+
+static uint8_t bsp_touchpad_custom_get_key_value(button_driver_t *button_driver)
+{
+    /* Init, if not */
+    bsp_touchpad_custom_init();
```
PetrESP: Here you are not checking the error return value

bsp/echoear/src/bsp_button.c
```diff
@@ -0,0 +1,206 @@
...
+
+    if (touchpad_ch < 0 || touchpad_ch > TOUCH_MAX_CHAN_ID || bsp_touchpad_chan_handles[touchpad_ch] == NULL) {
+        return 0;  /* released */
+    }
+
+    uint32_t data[TOUCH_SAMPLE_CFG_NUM];
+    if (touch_channel_read_data(bsp_touchpad_chan_handles[touchpad_ch], TOUCH_CHAN_DATA_TYPE_RAW, data) != ESP_OK) {
+        return 0;
+    }
+    /* Read RAW value and compare with initial read */
```
PetrESP: Just for thought: I think that it works as a simple demo. But what if you hold the device wrong when plugging it in? Then the initial value is messed up and the button will never work.

PetrESP (commented):
I will be able to test it on hardware tomorrow. For now I really like the idea of separating features into separate files because IMO the bsp source file was also getting really big. Also doesn't it help with the BSP generator? You can just have a template source file for each feature and include portions of it.
I've left some comments especially about the new touch controller which I've reviewed more than the rest of this PR. The addition is very nice. Thanks

PetrESP:
> Hi @tore-espressif @PetrESP @igrr I decided to divide SRC files in BSP by feature - what do you think? Check my latest commit in this PR. WHY: As a discussion on Mattermost in the BSP channel, I am thinking about versioning feature files, when there are more board versions. It can be `bsp_display_v1.c` or something like that and in CMake allow you to add both files or only one by `menuconfig`. When there will be more version files, I am looking for some solution (something like Ivan proposes)

I really like this idea of separating features into files **but** having multiple files which have almost the same content with just few changes here and there seems like a future maintenance issue.

I can't imagine having display_v1 and display_v2 where the only change is a different touch controller driver but when there are some inevitable IDF changes we need to edit both files (multiply that by number of new boards and maintained revisions).

I would much prefer having separate files for each feature but even if there are multiple board versions we keep the same functionality and only change what is necessary using IFDEFs etc.

igrr:
I think a per-feature split is nice, it makes it easy to understand and copy code into custom BSPs.

When we have multiple display versions, we don't have to keep common display code in v1 and v2 files, we can keep common code in display.c, while display_v1.c and display_v2.c only contain the small amount of code which is different.

espzav:
Thank you guys for your ideas and feedback. I will keep this feature (split source files) in BSP Generator and I will make this change in all re-generated BSPs - this is not breaking change. 
I agree with Ivan, we can keep everything in main `bsp_display.c` file and only changes add into separate files. The question is, how to do it simple (for generator and maintanance). 
- We should add one function for detection and functions for v1, v2, ...
- I proposing call the function for example `**b**sp_lcd_new_panel()` where will be `bsp_get_board_version()` detection called only once (like `bsp_i2c_init`, return static saved value if calling again); And then put it only on right places. Or something like this.
- Still thinking about using `IDF plugins` system, but I think it is not really good for this. 
- For BSP generator, I can change from one LCD/Touch to array and when there is more than one, I can generate board versions. Or only checkbox for pre-generate board versions in CMake and KConfig.

PetrESP:
I see. 

@espzav

> CMake allow you to add both files or only one by menuconfig

Okay so as I understand it we want to have:

display_v1.c
```
display_init(){
........ Some V1 specific code here ....
}
```
display_v2.c
```
display_init(){
........ Some V2 specific code here ....
}
```
CMakeList.txt
```
if(CONFIG_BSP_VERSION STREQUAL "v2")
list(APPEND SRCS src/display_v2.c)
else()
list(APPEND SRCS src/display_v1.c)
endif()
```
So we can just call `display_init` and let the build system handle this. Do I understand this correctly? Because in this case there is no run time configuration.

bsp/echoear/src/bsp_button.c
espzav: Reworked.

espzav:
@PetrESP Thank you for the review. I fixed all comments and reworked process of detection of touch button. PTAL again

espzav:
@tore-espressif Thank you for the revie. All comments were resolved and reworked touch buttons processing. PTAL again

bsp/echoear/src/bsp_button.c
```diff
@@ -0,0 +1,213 @@
...
+#define BSP_TOUCHPAD_RETRIES        (3)
+#define BSP_TOUCHPAD_THRESH_RATIO   (0.015f)
+#define BSP_TOUCHPAD_CHAN_COUNT     (2)
+
+static bool bsp_touchpad_sens_initialized = false;
+static touch_sensor_handle_t bsp_touchpad_sens_handle = NULL;
+static touch_channel_handle_t bsp_touchpad_chan_handles[TOUCH_MAX_CHAN_ID + 1] = { NULL };
+static uint32_t bsp_touchpad_last_val[TOUCH_MAX_CHAN_ID];
+
+static esp_err_t bsp_touchpad_custom_init()
```
tore-espressif, line 74: Well done navigating through this crazy init sequence!

tore-espressif (approved):
@espzav LGTM, however the touch pad driver v2 is beyond my comprehension...

PetrESP (approved):
I've tested it on hardware and everything works great.

PR #741, opened 2026-03-18
Expand sensor example and add BMI270 sensor driver

PetrESP:

- [ ] CI passing

# Change description
Added display visualization to the sensor example.
Added BMI270 sensor driver

### DIFF
--- a/.github/workflows/squareline.yml
+++ b/.github/workflows/squareline.yml
@@ -8,6 +8,10 @@ on:
       - 'SquareLine/**'
       - '.github/workflows/squareline.yml'
 
+# Required for uploading the release
+permissions:
+  contents: write
+
 jobs:
   create-squareline-packages:
     runs-on: ubuntu-24.04


components/sensors/bmi270/idf_component.yml
```diff
@@ -0,0 +1,12 @@
+version: 1.0.0
+description: I2C driver for BMI270 inertial measurement unit
+url: https://github.com/espressif/esp-bsp/tree/master/components/bmi270
+tags:
+  - i2c
+  - spi
```
espzav: SPI is not supported - please remove this tag

components/sensors/bmi270/include/bmi270.h
```diff
@@ -0,0 +1,232 @@
...
+
+#include "driver/i2c_types.h"
+#include "driver/spi_master.h"
+
+#define BMI270_CHIP_ID                      0x24
+#define BMI270_I2C_ADDRESS_L                0x68
+#define BMI270_I2C_ADDRESS_H                0x69
+
+// SPI interface is currently not supported
+typedef enum {
```
espzav, line 24: Please add commenst to enums, and structs.

components/sensors/bmi270/include/bmi270.h
```diff
@@ -0,0 +1,232 @@
...
+ * @param[in]  config       BMI270 sensor driver configuration
+ * @param[out] dev_handle   Pointer to the allocated memory for the BMI270 handle
+ *
+ * @return
+ *     - ESP_OK             Success
+ *     - ESP_ERR_NO_MEM     Memory allocation failure
+ *     - ESP_ERR_NOT_FOUND  Device was not found on the data bus
+ *     - Or other errors from the underlying I2C driver
+ */
+esp_err_t bmi270_init(const bmi270_driver_config_t *config, bmi270_handle_t **dev_handle);
```
espzav:
```suggestion
esp_err_t bmi270_create(const bmi270_driver_config_t *config, bmi270_handle_t **dev_handle);
```

espzav: We have another API for other sensors like tihis: `esp_err_t aht30_create(i2c_master_bus_handle_t i2c_bus, const uint8_t dev_addr, aht30_handle_t *handle_ret);`

Do we want to change it in all sensors? Or why this is different?

components/sensors/bmi270/include/bmi270.h
```diff
@@ -0,0 +1,232 @@
...
+ * @param[in]  dev_handle   Initialized BMI270 handle
+ * @param[out] x            X axis accelerometer value in g
+ * @param[out] y            Y axis accelerometer value in g
+ * @param[out] z            Z axis accelerometer value in g
+ *
+ * @return
+ *     - ESP_OK             Success
+ *     - Or other errors from the underlying I2C driver
+ */
+esp_err_t bmi270_get_acc_data(const bmi270_handle_t *dev_handle, float *x, float *y, float *z);
```
espzav: In `icm42670`, we are using names like `acce` and `gyro`. Can we make it same? - in all API functions.

components/sensors/bmi270/src/bmi270.c
```diff
@@ -0,0 +1,463 @@
...
+static const char *TAG = "BMI270";
+
+/*******************************************************************************
+* Public functions
+*******************************************************************************/
+
+esp_err_t bmi270_init(const bmi270_driver_config_t *config, bmi270_handle_t **dev_handle)
+{
+
+    assert(config != NULL);
```
espzav: Can we return error, when `config` is `NULL`?

espzav: Add check for `dev_handle` too.

components/sensors/bmi270/src/bmi270.c
```diff
@@ -0,0 +1,463 @@
...
+    if (config->interface == BMI270_USE_I2C) {
+        const i2c_device_config_t i2c_dev_cfg = {
+            .dev_addr_length = I2C_ADDR_BIT_LEN_7,
+            .device_address = config->addr,
+            .scl_speed_hz = I2C_CLK_SPEED,
+        };
+        ESP_GOTO_ON_ERROR(i2c_master_bus_add_device(config->i2c_bus, &i2c_dev_cfg, &handle->i2c_handle), err, TAG,
+                          "Failed to add new I2C device");
+        assert(handle->i2c_handle);
+    } else if (config->interface == BMI270_USE_SPI) {
```
espzav, line 65: Return error, not supported or something like that.

components/sensors/bmi270/src/bmi270.c
```diff
@@ -0,0 +1,463 @@
...
+    ESP_GOTO_ON_ERROR(bmi270_get_chip_id(handle, &data), err, TAG,
+                      "Failed to read ID");
+    ESP_GOTO_ON_FALSE(data == BMI270_CHIP_ID, ESP_ERR_NOT_FOUND, err, TAG, "Unexpected chip ID");
+
+    // Start the BMI270 and prepare it to load the configuration file
+    ESP_GOTO_ON_ERROR(soft_reset(handle), err, TAG, "Failed to request a soft reset");
+    ESP_GOTO_ON_ERROR(set_adv_pwr_save(handle, false), err, TAG, "Failed to disable power saving");
+    ESP_GOTO_ON_ERROR(set_config_load(handle, false), err, TAG, "Failed to disable configuration loading");
+
+    file_size = sizeof(bmi270_config_file);
```
espzav: Could you put the whole file uploading into separate function?

components/sensors/bmi270/src/bmi270.c
```diff
@@ -0,0 +1,463 @@
...
+    remainder = file_size % BMI270_FILE_WRITE_LEN;
+
+    // Pick write approach depending on it's divisibility by the write length
+    if (!remainder) {
+        for (file_index = 0; (file_index < file_size) && (ret == ESP_OK); file_index += BMI270_FILE_WRITE_LEN) {
+            ret = upload_file(handle, &bmi270_config_file[file_index], file_index, BMI270_FILE_WRITE_LEN);
```
espzav: When one transaction fails and next passes, you get `ESP_OK` in `ret`. You can use `ret |= ...` or return immediately, when failed.

components/sensors/bmi270/src/bmi270.c
```diff
@@ -0,0 +1,463 @@
...
+        }
+    } else {
+        chunk_size = file_size - remainder;
+        // Bulk write parts of the file
+        for (file_index = 0; (file_index < chunk_size) && (ret == ESP_OK); file_index += BMI270_FILE_WRITE_LEN) {
+            ret = upload_file(handle, &bmi270_config_file[file_index], file_index, BMI270_FILE_WRITE_LEN);
```
espzav: When one transaction fails and next passes, you get `ESP_OK` in `ret`. You can use `ret |= ...` or return immediately, when failed.

components/sensors/bmi270/src/bmi270.c
```diff
@@ -0,0 +1,463 @@
...
+        }
+
+        // Switch to half-word granular write
+        for (file_index = chunk_size; (file_index < file_size) && (ret == ESP_OK); file_index += 2) {
+            ret = upload_file(handle, &bmi270_config_file[file_index], file_index, BMI270_FILE_WRITE_LEN);
```
espzav: When one transaction fails and next passes, you get ESP_OK in ret. You can use ret |= ... or return immediately, when failed.

components/sensors/bmi270/src/bmi270.c
```diff
@@ -0,0 +1,463 @@
...
+
+    // Re-enable loading of the configuration file
+    ESP_GOTO_ON_ERROR(set_config_load(handle, true), err, TAG, "Failed to enable configuration loading");
+
+    // Wait for the BMI270 to initialize
+    status_checks = BMI270_STATUS_CHECK_CNT;
+    do {
+        read_register(handle, BMI270_INTERNAL_STATUS_REG, &data, 1);
+
+        ets_delay_us(BMI270_STATUS_CHECK_PERIOD);
```
espzav: Is it possible to use `vTaskDelay`? It can help to sleep this task and call another during waiting.

components/sensors/bmi270/src/bmi270.c
```diff
@@ -0,0 +1,463 @@
...
+        // Bulk write parts of the file
+        for (file_index = 0; (file_index < chunk_size) && (ret == ESP_OK); file_index += BMI270_FILE_WRITE_LEN) {
+            ret = upload_file(handle, &bmi270_config_file[file_index], file_index, BMI270_FILE_WRITE_LEN);
+        }
+
+        // Switch to half-word granular write
+        for (file_index = chunk_size; (file_index < file_size) && (ret == ESP_OK); file_index += 2) {
+            ret = upload_file(handle, &bmi270_config_file[file_index], file_index, BMI270_FILE_WRITE_LEN);
+        }
+    }
```
espzav: There is no check fo `ret` after any `for`.

components/sensors/bmi270/src/bmi270.c
```diff
@@ -0,0 +1,463 @@
...
+    ESP_LOGI(TAG, "Successfully initialized the BMI270 sensor driver");
+
+    return ret;
+
+err:
+    ESP_RETURN_ON_ERROR(bmi270_delete(&handle), TAG, "Clean up failure") ;
+    return ret;
+}
+
+esp_err_t bmi270_delete(bmi270_handle_t **dev_handle)
```
espzav:
```suggestion
esp_err_t bmi270_delete(bmi270_handle_t *dev_handle)
```
?

components/sensors/bmi270/src/bmi270.c
```diff
@@ -0,0 +1,463 @@
...
+{
+    assert(dev_handle != NULL);
+    assert(chip_id != NULL);
+    return read_register(dev_handle, BMI270_CHIP_ID_REG, chip_id, 1);
+}
+
+esp_err_t bmi270_start(bmi270_handle_t *dev_handle, const bmi270_config_t *config)
+{
+
+    assert(dev_handle != NULL);
```
espzav: I think that we talked about it in some my review. We can check it and return error instead of assert. 
It is ok to use for private functions.

components/sensors/bmi270/src/bmi270.c
```diff
@@ -0,0 +1,463 @@
...
+
+    uint8_t data;
+
+    if (!dev_handle->initialized) {
+        return ESP_ERR_INVALID_STATE;
+    }
+
+    // Disable all internal BMI270 blocks
+    data = 0;
```
espzav: You can set `data` on top.

components/sensors/bmi270/src/bmi270.c
```diff
@@ -0,0 +1,463 @@
...
+
+static esp_err_t soft_reset(const bmi270_handle_t *dev_handle)
+{
+    assert(dev_handle != NULL);
+
+    uint8_t data = BMI270_CMD_SOFT_RESET;
+
+    ESP_RETURN_ON_ERROR(write_register(dev_handle, BMI270_CMD_REG, &data, 1), TAG,
+                        "Failed to soft reset the device");
+    ets_delay_us(BMI270_SOFT_RESET_TIME);
```
espzav: Is it possible to use `vTaskDelay`?

components/sensors/bmi270/src/bmi270_sensor_hub.c
```diff
@@ -0,0 +1,98 @@
...
+
+/*******************************************************************************
+* Public functions
+*******************************************************************************/
+
+esp_err_t bmi270_impl_init(bus_handle_t bus_handle, uint8_t addr)
+{
+    const bmi270_driver_config_t driver_config = {
+        .addr = addr,
+        .interface = BMI270_USE_I2C,
```
espzav, line 38: Sensor HUB using only I2C?

examples/sensors/main/main.c
```diff
@@ -47,6 +57,7 @@ void sensor_event_handler(void *handler_args, esp_event_base_t base, int32_t id,
                  sensor_data->sensor_name,
                  sensor_data->sensor_addr,
                  sensor_data->humidity);
+        chart_add_value(humid_series, (int32_t)(sensor_data->humidity * 10.0f));
```
espzav: 1. When this example using display, could you rename it to `display_sensors`? 
2. I think, all functions for LVGL should be in `#if BSP_CAPS_DISPLAY`

components/sensors/bmi270/include/bmi270.h
PetrESP: I don't think that we need to change it in all sensors right now. But if I remember correctly we have discussed and decided that we want to abstract the interface configuration so that we can choose either SPI or I2C using the same function. And since SPI does not require an address, I've decided to put it also in the configuration structure

components/sensors/bmi270/src/bmi270.c
PetrESP: Do you mean instead of the assert or keep the assert and also return?

components/sensors/bmi270/src/bmi270.c
PetrESP: It would not be possible to the do this:
`*dev_handle = NULL;`
But if you want to omit this then we can leave it up to the integration to NULL this pointer.

components/sensors/bmi270/src/bmi270.c
PetrESP: I think we can. It is just 2 ms but I think that is by default 20 slices which is enough to make sense

components/sensors/bmi270/src/bmi270_sensor_hub.c
PetrESP: Currently there is no way to pass user data through to the initialization function in sensor hub. Probably for now the plan for sensors with configurable interfaces is to have multiple sensor hub implementation structures.

examples/sensors/main/main.c
PetrESP: 1. There already is an example called `display_sensors`
2. Yeah looking at it I think that there are multiple improvements I can make. I will do it

components/sensors/bmi270/include/bmi270.h
espzav: ok

examples/sensors/main/main.c
espzav: ad 1. we can remove the old `display_sensors` example. Azure kit is deprecated.

components/sensors/bmi270/src/bmi270.c
espzav: You can use `ESP_RETURN_ON_ERROR` or somethink like that with printed error. It is enough. Assert can be disabled in release build.

components/sensors/bmi270/src/bmi270.c
```diff
@@ -0,0 +1,481 @@
...
+        };
+        ESP_GOTO_ON_ERROR(i2c_master_bus_add_device(config->i2c_bus, &i2c_dev_cfg, &handle->i2c_handle), err, TAG,
+                          "Failed to add new I2C device");
+        assert(handle->i2c_handle);
+    } else if (config->interface == BMI270_USE_SPI) {
+
+        // TODO: Reminder when implementing the SPI interface
+        // Dummy SPI read which sets the device to SPI mode
+        // ESP_GOTO_ON_ERROR(bmi270_get_chip_id(handle, &data), err, TAG,
+        // "Failed to perform a dummy read");
```
espzav, line 70: I still don't see any return value here. 
When the return will be here, don't forgot free `handle`.
- Or you can add this condition to top.

components/sensors/bmi270/src/bmi270.c
espzav: This is not usual to set NULL to pointer inside function for deinit.

components/sensors/bmi270/src/bmi270_sensor_hub.c
```diff
@@ -0,0 +1,98 @@
...
+                        "Failed to initialize");
+
+    // Set a reasonable default configuration
+    const bmi270_config_t config = {
+        .acce_odr = BMI270_ACC_ODR_100_HZ,
+        .acce_range = BMI270_ACC_RANGE_4_G,
+        .gyro_odr = BMI270_GYR_ODR_100_HZ,
+        .gyro_range = BMI270_GYR_RANGE_1000_DPS
+    };
+    ESP_RETURN_ON_ERROR(bmi270_start(sensor_hub_bmi270_handle, &config), TAG, "Failed to start");
```
espzav: When it fails, you need to call `bmi270_delete`

components/sensors/bmi270/test_app/main/test_app_bmi270.c
```diff
@@ -0,0 +1,141 @@
...
+    ret = bmi270_get_acce_data(bmi270, &x, &y, &z);
+    TEST_ASSERT_EQUAL(ESP_OK, ret);
+    ret = bmi270_get_gyro_data(bmi270, &x, &y, &z);
+    TEST_ASSERT_EQUAL(ESP_OK, ret);
+
+    ret = bmi270_stop(bmi270);
+    TEST_ASSERT_EQUAL(ESP_OK, ret);
+
+    ret = bmi270_delete(&bmi270);
+    TEST_ASSERT_EQUAL(ESP_OK, ret);
```
espzav, line 87: Missing I2C deinit.

examples/display_sensors/README.md
```diff
@@ -1,20 +1,10 @@
...
-At the start of the program, magnetometer calibration is performed for 10 seconds.
-Turn the board in every axis during this time to achieve best magnetometer results.
-
-## LED and buzzer
-On every press of KEY_IO0 button, the buzzer beeps and AZURE LED blinks.
-
-## uSD card
-If a uSD card is successfully mounted a hello.txt file is created and WIFI LED is turned on.
+All sensors are sampled depending on the configured sampling period.
+Measurement results are printed on the terminal.
```
espzav:
```suggestion
Measurement results are printed on the terminal or showed on the display if available.
```

Or something like this. There will be nice to add screen image. And make same style like other display examples.

components/sensors/bmi270/src/bmi270.c
```diff
@@ -0,0 +1,481 @@
+/*
+ * SPDX-FileCopyrightText: 2026 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <string.h>
+#include "driver/i2c_master.h"
```
espzav, line 8: Missing include for `vTaskDelay`:
```
#include "freertos/FreeRTOS.h"
```

examples/display_sensors/main/main.c
```diff
@@ -1,305 +1,248 @@
...
+                 sensor_data->sensor_name,
+                 sensor_data->sensor_addr);
+        break;
+    case SENSOR_STOPED:
+        ESP_LOGI(TAG, "Timestamp = %llu - %s_0x%x STOPPED",
+                 sensor_data->timestamp,
+                 sensor_data->sensor_name,
+                 sensor_data->sensor_addr);
+        break;
+    case SENSOR_HUMI_DATA_READY:
```
espzav, line 59: This case should be visible only when `BSP_CAPS_HUMITURE`. Otherwise there is missing `chart_add_value` if there is only IMU.

components/sensors/bmi270/README.md
```diff
@@ -0,0 +1,93 @@
...
+```
+After a completed measurement data can be read using:
+``` c
+bmi270_get_acce_data(bmi270_handle_pointer, &x, &y, &z);
+```
+Disable measurements and clean up the BMI270 driver using:
+``` c
+bmi270_stop(bmi270_handle_pointer);
+bmi270_delete(bmi270_handle_pointer);
+free(bmi270_handle_pointer);
```
espzav: I don't understand - why?

PetrESP: I've copied the wrong thing :facepalm: It was supposed to be `bmi270_handle_pointer = NULL;`
It's fixed now

espzav:
@PetrESP We forgot to add this component to upload component job and this is missing in Component registry.

espzav:
@PetrESP One additional thing - there will be nice to add one additional tag: "humiture/imu/..." for BSP generator.

Issue #743, opened 2026-03-21
Macro for setup of I2C results in "i2c_master_bus_add_device(1177): invalid scl frequency"


sandynomad:
### Board

Waveshare AMOLED 2.41

### Hardware Description

None.

### IDE Name

ESP-IDF

### Operating System

OSX

### Description

When using this code:

`esp_lcd_panel_io_i2c_config_t io_config = ESP_LCD_TOUCH_IO_I2C_FT5x06_CONFIG();
ESP_ERROR_CHECK( esp_lcd_new_panel_io_i2c(  i2c_handle,  &io_config, &tp_io_handle ) );
`

the ESP32-S3 fails with:

`E (1426) i2c.master: i2c_master_bus_add_device(1177): invalid scl frequency
E (1426) lcd_panel.io.i2c: esp_lcd_new_panel_io_i2c_v2(70): i2c add device fail
ESP_ERROR_CHECK failed: esp_err_t 0x102 (ESP_ERR_INVALID_ARG) at 0x42006d85
file: "src/chips/touch.cpp" line 30
func: void touch_init()
expression: esp_lcd_new_panel_io_i2c( i2c0.handle(), &io_config, &tp_io_handle )

abort() was called at PC 0x4037a753 on core 0
`

This appears to be due to   <io_config.scl_speed_hz> not being defined.

### Sketch

```cpp
esp_lcd_panel_io_i2c_config_t io_config = ESP_LCD_TOUCH_IO_I2C_FT5x06_CONFIG();
ESP_ERROR_CHECK( esp_lcd_new_panel_io_i2c(  i2c_handle,  &io_config, &tp_io_handle ) );
```

### I have checked existing issues, README.md and ESP32 Forum

- [x] I confirm I have checked existing issues, online documentation and Troubleshooting guide.

PetrESP:
Hello @sandynomad,

Thank you for reporting this issue. 
Currently the correct initialization flow is this:
```
#define FT5x06_I2C_SPEED 400000     
esp_lcd_panel_io_i2c_config_t io_config = ESP_LCD_TOUCH_IO_I2C_FT5x06_CONFIG();
io_config.scl_speed_hz = FT5x06_I2C_SPEED;
ESP_ERROR_CHECK( esp_lcd_new_panel_io_i2c(  i2c_handle,  &io_config, &tp_io_handle ) );
```
I will check if there is a reason to not set the speed in the define and I will either fix the documentation or the define so that it is clear.

[...]

PR #744, opened 2026-03-23
fix(lcd_touch): Add a default I2C clock speed to the configuration macro

PetrESP:

- [ ] CI passing

# Change description
Added default assignment to `scl_speed_hz` in all configuration macros for LCD touch controllers
Relates to #743

### DIFF
--- a/components/esp_lvgl_port/idf_component.yml
+++ b/components/esp_lvgl_port/idf_component.yml
@@ -2,7 +2,7 @@ version: "2.7.2"
 description: ESP LVGL port
 url: https://github.com/espressif/esp-bsp/tree/master/components/esp_lvgl_port
 dependencies:
-  idf: ">=5.1"
+  idf: ">=5.2"
   lvgl/lvgl:
     version: ">=8,<10"
     public: true
--- a/components/lcd_touch/esp_lcd_touch_ft5x06/idf_component.yml
+++ b/components/lcd_touch/esp_lcd_touch_ft5x06/idf_component.yml
@@ -2,7 +2,7 @@ version: "1.1.0~1"
 description: ESP LCD Touch FT5x06 - touch controller FT5x06
 url: https://github.com/espressif/esp-bsp/tree/master/components/lcd_touch/esp_lcd_touch_ft5x06
 dependencies:
-  idf: ">=4.4.2"
+  idf: ">=5.2"
   esp_lcd_touch:
     version: "^1.2.0"
     public: true
--- a/components/lcd_touch/esp_lcd_touch_stmpe610/idf_component.yml
+++ b/components/lcd_touch/esp_lcd_touch_stmpe610/idf_component.yml
@@ -2,7 +2,7 @@ version: "1.0.8"
 description: ESP LCD Touch STMPE610 - touch controller STMPE610
 url: https://github.com/espressif/esp-bsp/tree/master/components/esp_lcd_touch_stmpe610
 dependencies:
-  idf: ">=5.0"
+  idf: ">=5.2"
   esp_lcd_touch:
     version: "^1.2.0"
     public: true
--- a/SquareLine/boards/v8/custom_waveshare_7inch/main/idf_component.yml
+++ b/SquareLine/boards/v8/custom_waveshare_7inch/main/idf_component.yml
@@ -1,6 +1,6 @@
 description: ESP-BSP SquareLine LVGL Example
 dependencies:
-  idf: ">=4.4"
+  idf: ">=5.2"
   ws_7inch:
     version: "1.0"
     override_path: "../components/ws_7inch"
--- a/SquareLine/boards/v8/esp32_c3_lcdkit/main/idf_component.yml
+++ b/SquareLine/boards/v8/esp32_c3_lcdkit/main/idf_component.yml
@@ -2,5 +2,5 @@ description: ESP-BSP SquareLine LVGL Example
 targets:
     - esp32c3
 dependencies:
-  idf: ">=4.4"
+  idf: ">=5.2"
   esp32_c3_lcdkit: "==1.0.1"
--- a/SquareLine/boards/v8/esp32_s3_eye/main/idf_component.yml
+++ b/SquareLine/boards/v8/esp32_s3_eye/main/idf_component.yml
@@ -2,5 +2,5 @@ description: ESP-BSP SquareLine LVGL Example
 targets:
     - esp32s3
 dependencies:
-  idf: ">=4.4"
+  idf: ">=5.2"
   esp32_s3_eye: "==3.0.4"
--- a/SquareLine/boards/v8/esp32_s3_korvo_2/main/idf_component.yml
+++ b/SquareLine/boards/v8/esp32_s3_korvo_2/main/idf_component.yml
@@ -2,5 +2,5 @@ description: ESP-BSP SquareLine LVGL Example
 targets:
     - esp32s3
 dependencies:
-  idf: ">=4.4"
+  idf: ">=5.2"
   esp32_s3_korvo_2: "==2.1.2"
--- a/SquareLine/boards/v8/esp32_s3_usb_otg/main/idf_component.yml
+++ b/SquareLine/boards/v8/esp32_s3_usb_otg/main/idf_component.yml
@@ -2,5 +2,5 @@ description: ESP-BSP SquareLine LVGL Example
 targets:
     - esp32s3
 dependencies:
-  idf: ">=4.4"
+  idf: ">=5.2"
   esp32_s3_usb_otg: "==1.5.1"
--- a/SquareLine/boards/v8/espbox_3/main/idf_component.yml
+++ b/SquareLine/boards/v8/espbox_3/main/idf_component.yml
@@ -2,5 +2,5 @@ description: ESP-BSP SquareLine LVGL Example
 targets:
     - esp32s3
 dependencies:
-  idf: ">=4.4"
+  idf: ">=5.2"
   esp-box-3: "==1.1.3"
--- a/SquareLine/boards/v8/espwroverkit/main/idf_component.yml
+++ b/SquareLine/boards/v8/espwroverkit/main/idf_component.yml
@@ -2,5 +2,5 @@ description: ESP-BSP SquareLine LVGL Example
 targets:
     - esp32
 dependencies:
-  idf: ">=4.4"
+  idf: ">=5.2"
   esp_wrover_kit: "==1.5.1"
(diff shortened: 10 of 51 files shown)


components/lcd/esp_lcd_sh1107/include/esp_lcd_sh1107.h
```diff
@@ -41,17 +41,18 @@ esp_err_t esp_lcd_new_panel_sh1107(const esp_lcd_panel_io_handle_t io,
...
-        .dc_bit_offset = 0,                 \
-        .lcd_cmd_bits = 8,                  \
-        .lcd_param_bits = 8,                \
-        .flags =                            \
-        {                                   \
-            .disable_control_phase = 1,     \
-        }                                   \
+#define ESP_LCD_IO_I2C_SH1107_CONFIG()              \
+    {                                               \
+        .scl_speed_hz = 100000,                     \
```
espzav, line 46: Usually, we are using fast mode in BSP (400kHz). Why do you using 100k in all files?

PetrESP: I felt like it was a safe default value. We can do 400k if you want.

espzav: Yes, I undrestand. It is safe, but usually customers will take default config and use it. I don't know, which solution is better...

components/lcd_touch/esp_lcd_touch_stmpe610/idf_component.yml
```diff
@@ -1,4 +1,4 @@
-version: "1.0.8"
+version: "1.0.8~1"
```
espzav: There is no change in stmpe610

components/lcd/esp_lcd_sh1107/include/esp_lcd_sh1107.h
```diff
@@ -41,17 +41,18 @@ esp_err_t esp_lcd_new_panel_sh1107(const esp_lcd_panel_io_handle_t io,
  * @brief Touch IO configuration structure
  *
  */
-#define ESP_LCD_IO_I2C_SH1107_CONFIG()             \
```
espzav, line 44: Missing version update in this component.

bsp/m5stack_tab5/src/m5stack_tab5.c
```diff
@@ -792,7 +792,7 @@ esp_err_t bsp_touch_new(const bsp_touch_config_t *config, esp_lcd_touch_handle_t
     esp_lcd_panel_io_i2c_config_t tp_io_config = ESP_LCD_TOUCH_IO_I2C_GT911_CONFIG();
     tp_io_config.dev_addr = ESP_LCD_TOUCH_IO_I2C_GT911_ADDRESS_BACKUP;
     tp_io_config.scl_speed_hz =
-        CONFIG_BSP_I2C_CLK_SPEED_HZ; // This parameter was introduced together with I2C Driver-NG in IDF v5.2
+        CONFIG_BSP_I2C_CLK_SPEED_HZ; // Overwrite the default value with a configurable one
```
espzav: When speed is per device, I think we can remove this Kconfig option in future.

components/lcd/esp_lcd_sh1107/include/esp_lcd_sh1107.h
PetrESP: It depends. I think going for 100k as the default means that it will work even if you have some weird design which uses internal pull-ups but also you are not maximizing the I2C speed. But I would say that for a touch controller 100k is already enough.
On the other hand 400k is supported by most (probably all that we support) controllers and is faster.
I don't mind either way I've just picked what made more sense to me.

espzav: ok, we can keep it as it is.

PetrESP:
@espzav I've realized that we don't even test lower versions than 5.2 (we can't make sure that they even work) so I think that we can bump all components to at least 5.2.
Do I need to update their versions as well? Or you wanna keep it like this? I don't think that here is a hard reason to upload it right now

bsp/esp32_p4_function_ev_board/esp32_p4_function_ev_board.c
```diff
@@ -854,7 +854,6 @@ esp_err_t bsp_touch_new(const bsp_touch_config_t *config, esp_lcd_touch_handle_t
         },
     };
     esp_lcd_panel_io_i2c_config_t tp_io_config = ESP_LCD_TOUCH_IO_I2C_GT911_CONFIG();
-    tp_io_config.scl_speed_hz = CONFIG_BSP_I2C_CLK_SPEED_HZ;
```
espzav, line 857: Remove from Kconfig of this board too.

bsp/esp32_p4_function_ev_board/idf_component.yml
```diff
@@ -1,4 +1,4 @@
-version: "5.2.2"
+version: "5.2.3"
```
espzav, line 1: Why here changed only bugfix version?

bsp/esp_bsp_generic/src/esp_bsp_generic.c
```diff
@@ -641,36 +641,26 @@ esp_err_t bsp_touch_new(const bsp_touch_config_t *config, esp_lcd_touch_handle_t
 #if CONFIG_BSP_TOUCH_DRIVER_TT21100
     ESP_LOGI(TAG, "Initialize LCD Touch: TT21100");
     esp_lcd_panel_io_i2c_config_t tp_io_config = ESP_LCD_TOUCH_IO_I2C_TT21100_CONFIG();
-    tp_io_config.scl_speed_hz =
```
espzav, line 644: Remove from Kconfig too

bsp/esp_vocat/src/bsp_display.c
```diff
@@ -266,7 +266,7 @@ esp_err_t bsp_touch_new(const bsp_touch_config_t *config, esp_lcd_touch_handle_t
     esp_lcd_panel_io_handle_t tp_io_handle = NULL;
     esp_lcd_panel_io_i2c_config_t tp_io_config = ESP_LCD_TOUCH_IO_I2C_CST816S_CONFIG();
     tp_io_config.scl_speed_hz =
-        CONFIG_BSP_I2C_CLK_SPEED_HZ; // This parameter was introduced together with I2C Driver-NG in IDF v5.2
+        CONFIG_BSP_I2C_CLK_SPEED_HZ; // Overwrite the default value with a configurable one
```
espzav: Remove it - or should I regenerate?

bsp/m5stack_core_s3/m5stack_core_s3.c
```diff
@@ -523,7 +523,7 @@ esp_err_t bsp_touch_new(const bsp_touch_config_t *config, esp_lcd_touch_handle_t
     esp_lcd_panel_io_handle_t tp_io_handle = NULL;
     esp_lcd_panel_io_i2c_config_t tp_io_config = ESP_LCD_TOUCH_IO_I2C_FT5x06_CONFIG();
     tp_io_config.scl_speed_hz =
-        CONFIG_BSP_I2C_CLK_SPEED_HZ; // This parameter was introduced together with I2C Driver-NG in IDF v5.2
+        CONFIG_BSP_I2C_CLK_SPEED_HZ; // Overwrite default value with a configurable one
```
espzav: Remove it

components/esp_lvgl_port/idf_component.yml
```diff
@@ -2,7 +2,7 @@ version: "2.7.2"
 description: ESP LVGL port
 url: https://github.com/espressif/esp-bsp/tree/master/components/esp_lvgl_port
 dependencies:
-  idf: ">=5.1"
+  idf: ">=5.2"
```
espzav, line 5: LVGL port has changelog - please add this change to it.

components/lcd_touch/esp_lcd_touch_ft5x06/idf_component.yml
```diff
@@ -1,8 +1,8 @@
-version: "1.1.0"
+version: "1.1.0~1"
```
espzav, line 1: You increased version only in some components, why?

bsp/esp32_p4_function_ev_board/esp32_p4_function_ev_board.c
PetrESP: I've left it intentionally because it is still used in the LCD panel configuration

bsp/esp32_p4_function_ev_board/idf_component.yml
PetrESP: Because the functionality is the same. I didn't remove any kconfig option but I've just removed the redundant setting of scl_speed_hz

bsp/esp_vocat/src/bsp_display.c
PetrESP: I've missed that one. Removed

bsp/m5stack_core_s3/m5stack_core_s3.c
PetrESP: I wanted to keep it in all board which also use I2C to configure other devices (for example the P4 function ev board and the S3 core devices) but looking back at it I will remove them.

components/esp_lvgl_port/idf_component.yml
PetrESP: I wrote a comment about this. Please respond to it

components/lcd_touch/esp_lcd_touch_ft5x06/idf_component.yml
PetrESP: I've increased it in those where I've removed the I2C change. For those with minimal supported version I did not bump it yet.

bsp/esp32_p4_function_ev_board/esp32_p4_function_ev_board.c
espzav: ok

bsp/esp32_p4_function_ev_board/idf_component.yml
espzav: ok

components/lcd_touch/esp_lcd_touch_ft5x06/idf_component.yml
espzav: ok

bsp/esp_vocat/src/bsp_display.c
```diff
@@ -265,8 +265,6 @@ esp_err_t bsp_touch_new(const bsp_touch_config_t *config, esp_lcd_touch_handle_t
     };
     esp_lcd_panel_io_handle_t tp_io_handle = NULL;
     esp_lcd_panel_io_i2c_config_t tp_io_config = ESP_LCD_TOUCH_IO_I2C_CST816S_CONFIG();
-    tp_io_config.scl_speed_hz =
-        CONFIG_BSP_I2C_CLK_SPEED_HZ; // This parameter was introduced together with I2C Driver-NG in IDF v5.2
```
espzav, line 269: Missing version change in VoCat

PR #751, opened 2026-04-02
ESP-BSP benchmark action

PetrESP:

- [ ] CI passing

# Change description
Created a separate action which generates benchmark results
Removed benchmarks from build-run action


SquareLine/gen.py
```diff
@@ -6,7 +6,6 @@
 import shutil
 import argparse
 from zipfile import ZipFile
-from typing import Optional
```
tore-espressif, line 9: This is here for old-style type comment on line 60 https://github.com/espressif/esp-bsp/blob/40e821ccb8a571cc532478f1470bb193b931a314/SquareLine/gen.py#L60

If you want to remove `from typing import Optional` you should update line 60 to modern type annotation

``` python
def color_print(message: str, color: str, newline: str | None = '\n') -> None:
```

.flake8
```diff
@@ -124,11 +124,16 @@ select =
...
-    E241,           # multiple spaces after ','
-    W503,           # line break before binary operator
-    W504            # line break after binary operator
+    # multiple spaces before operator
+    E221,
+    # missing whitespace after ',', ';', or ':'
+    E231,
+    # multiple spaces after ','
+    E241,
+    # line break before binary operator
```
tore-espressif, line 133: old format seems more readable to me, but I'll leave this up to you...

conftest.py
```diff
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: 2024-2025 Espressif Systems (Shanghai) CO LTD
+# SPDX-FileCopyrightText: 2024-2026 Espressif Systems (Shanghai) CO LTD
```
espzav, line 1: There is no reason to change year here.

espzav (approved):
@PetrESP Thank you for this work! That's great! 
I am not good person for make review of python and CI files. I looked at it and looks ok to me 👍

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,138 @@
...
+
+permissions:
+  checks: write
+
+jobs:
+  build_benchmarks:
+    env:
+      BENCHMARK_PROJECT_DIR: "examples/display_lvgl_benchmark"
+    runs-on: ubuntu-latest
+    container: espressif/idf:release-v5.5
```
tore-espressif, line 21: Question: Will we test only on one version of IDF? Or will this be expanded to a 'matrix' strategy in the future?

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,138 @@
...
+  pull_request:
+    types: [opened, reopened, synchronize, labeled]
+
+permissions:
+  checks: write
+
+jobs:
+  build_benchmarks:
+    env:
+      BENCHMARK_PROJECT_DIR: "examples/display_lvgl_benchmark"
```
tore-espressif: Does not have to be changed, just FYI:

you can define the env var at the top level like this

```
env:
  BENCHMARK_PROJECT_DIR: examples/display_lvgl_benchmark
```

Then, you do not have to define it separately for `build_benchmarks` and `run_benchmarks` jobs

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,138 @@
...
+          # Run benchmarks with custom config
+          pytest \
+            ${{ env.BENCHMARK_PROJECT_DIR }}/pytest_display_lvgl_benchmark.py \
+            -c pytest.ini \
+            --suppress-no-test-exit-code \
+            --junit-xml=${{ env.TEST_RESULT_NAME }}.xml \
+            -p no:idf-ci \
+            --ignore-glob '*/managed_components/*' \
+            --ignore=.github \
+            -n auto
```
tore-espressif, line 86: Nitpick, does not have to be changed:

The idea of `pytest.ini` file is to provide all pytest project wide configuration in one file. So we do not have to have inconveniently long command line.

Eg. `-p no:idf-ci` is already present in pytest.ini and `--ignore-glob '*/managed_components/*'` and `--ignore=.github` could be moved there because they are valid for all pytest runs in esp-bsp repository.

Also `-c pytest.ini` is a default value, but feel free to keep it there if you want to be explicit

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,138 @@
...
+      - name: Upload benchmark metrics
+        uses: actions/upload-artifact@v7
+        if: always()
+        with:
+          name: benchmark_metrics
+          path: bsp_metrics.json
+  upload_metrics:
+    runs-on: ubuntu-latest
+    # Dry run on pull requests and manual triggers
+    if: github.repository_owner == 'espressif' && (github.event_name == 'schedule' || github.event_name == 'push')
```
tore-espressif, line 112: Question: Will the PR author be able to see the benchmark results for his changes?

tore-espressif (approved):
@PetrESP great work! I think that moving the benchmark into a separate workflow is the right idea. Also kudos for result uploading with `esp-metrics-cli`!

I'm leaving just nitpicks; nothing blocking

.flake8
PetrESP: Sadly the newer versions of flake8 does not support inline comments and will cause an error on the old format. There were some issues with them so they stopped supporting them. That is why I had to change it. (I think it is from version 6.0.0.) 
I had to bump it because it was throwing `AttributeError("module 'ast' has no attribute 'Str'")` on newer python versions.

.github/workflows/benchmark.yml
PetrESP: Yes in the future we will. But at this point I wanted to keep it stable since we had some issues with 6.0

.github/workflows/benchmark.yml
PetrESP: Oh I did not know about the `-c pytest.ini` I though I had to specify it because I was running test in the specified directory.
Regarding the pytest.ini I have prepared changes to move a lot of these arguments there. I plan to merge it after this one with changes to this pytest call. https://github.com/espressif/esp-bsp/pull/785

.github/workflows/benchmark.yml
PetrESP: Currently you can manually check the artifacts if you think your changes will affect performance/binary size. For now I think doing some comparison with the latest result would be a bit flaky, because we don't have a good access to the database to set up a querry. It would be nice to do it in the future but for now I think that having a history of results is more valuable.

SquareLine/gen.py
PetrESP: Makes sense. Done

.flake8
tore-espressif: Thank you for the explanation

PR #772, opened 2026-04-29
Update build rules and add compilation tests 

PetrESP:
- [x] CI passing

# Change description
- Added/updated build rules for CI. Some components were missing and some were outdated
- Removed a test which was used to test compilation of certain components. This test was removed in place of multiple compilation tests per component which are used according to the build rules.


test_apps/.build-test-rules.yml
```diff
@@ -1,10 +1,5 @@
 # Noglib test_app: Build only in CI, where ENV_BUILD_NOGLIB is set
 test_apps/noglib:
   disable:
+    # TODO: Is this disabled anywhere?
     - if: ENV_BUILD_NOGLIB == 0
```
PetrESP: @espzav Do you have any ideas what could this be for? It does not seem to be se anywhere in the repository

espzav: @PetrESP I am not sure, where it was used. It seems it is missing somewhere. It should build noglib test app only with noglib BSP.

components/.build-test-rules.yml
```diff
@@ -1,24 +1,8 @@
...
-      reason: Requires I2C Driver-NG which was introduced in v5.2
-
-components/esp_lvgl_port/test_apps/lvgl_port:
-  depends_filepatterns:
-    - "components/esp_lvgl_port/**"
-  disable:
-    - if: IDF_VERSION < "5.2.0"
-      reason: Requires I2C Driver-NG which was introduced in v5.2
+    - "components/lcd_touch/**"
+    - "components/lcd/**"
```
espzav, line 5: `-"components/sensors/**"` missing here

test_apps/.build-test-rules.yml
espzav: It seems, we can remove this file. But it seems, we haven't got any test for noglib.

components/.build-test-rules.yml
PetrESP: I could be wrong but I don't think that `lvgl_port` has any dependencies on sensors.

espzav: Oh, I overlooked that it is for lvgl_port. I am sorry.

test_apps/.build-test-rules.yml
PetrESP: The `noglib` project was supposed to at least compile it but it seems that with this `disable` it was never compiled. If I remove it now the CI will fail because it can't be compiled. Maybe I leave a `TODO` and we will add at least a compilation after we do the LVGL noglib changes?

espzav: ok, after the split BSP, we will make new example for compile+runtime testing NoGlib versions

espzav (approved):
@PetrESP Good job! Thank you! LGTM

PR #773, opened 2026-04-29
Remove Jira workflow and automatically upload all components from a directory

PetrESP:

- [ ] CI passing

# Change description
Remove unused Jira workflow
Automate addition of components


PetrESP:
@espzav I made it so that all of our non-bsp component list gets updated automatically. So now we don't have to manually keep a list.
In the future we can do the same with BSPs and noglib but there are some deprecated BSPs which we don't want to automatically upload.
I am open to suggestions if you don't think that this is a good idea.

.github/workflows/upload_component.yml
```diff
@@ -65,48 +65,21 @@ jobs:
   upload_other_components:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@v5
```
tore-espressif: Was there an issue with checkout v6?

.github/workflows/upload_component.yml
```diff
@@ -65,48 +65,21 @@ jobs:
...
+      - name: Discover components to upload
+        id: discover_components
+        run: |
+          config_files=$(grep -rl '^version:' components/)
+          component_dirs=$(dirname $config_files)
```
espzav, line 73:
```suggestion
          component_dirs=$(dirname $config_files | sort -u)
```
It can solve duplicity.

.github/workflows/upload_component.yml
```diff
@@ -65,48 +65,21 @@ jobs:
   upload_other_components:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@v5
+      - name: Discover components to upload
+        id: discover_components
+        run: |
+          config_files=$(grep -rl '^version:' components/)
```
espzav, line 72: I am not sure if this is the right form of find all components. What about use "component.yml"?

espzav (commented):
@PetrESP This is good idea, thank you! 
I am thinking about some tag (or some comment) for deprecated components for skip it here. - we shouldn't make changes in deprecated components, then it is ok to keep it here...

tore-espressif (commented):
looks good, thanks!

.github/workflows/upload_component.yml
PetrESP: I am not looking for component YAML files because there are also test apps which contain them but they are not a component.

.github/workflows/upload_component.yml
PetrESP: There were on issues. I've messed up the rebase so thank you for catching that.

.github/workflows/upload_component.yml
PetrESP: This is an interesting idea. But don't we want to do the opposite? (which is crashing when there are 2 came components) Because otherwise we would be silently not uploading one of them.

espzav: Agree.

.github/workflows/upload_component.yml
espzav: ok

espzav (approved):
@PetrESP Thank you! LGTM

PR #774, opened 2026-04-30
feat(bsp_graphics_lvgl): Created new component for BSP graphics (LVGL) and removed LVGL from BSP

espzav:

- [ ] CI passing

# Change description


espzav:
Hi @tore-espressif, @igrr, @PetrESP, @georgik  this is the proposal of the splitting BSP to noglib part + new component graphics (LVGL). 
- This is a working part without big breaking changes (need to add component to example/project if not using BSP Selector AND add header file)
-  Next step will be add new API and mark this as DEPRECATED. (I will make it in the same PR)
- I tried it with CMake V2, it is not working now, because LVGL port has some issues with V2. 

Please, could you look at it and give me some feedback?

Thank you!

components/bsp_graphics/bsp_graphics_lvgl/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+idf_component_register(
+    SRCS "src/bsp_graphics_lvgl.c"
+    INCLUDE_DIRS "include"
+    PRIV_INCLUDE_DIRS "priv_include"
+)
+
+# Add BSP from project YML
+target_link_libraries(${COMPONENT_LIB} PUBLIC esp_bsp::core)
+target_compile_definitions(${COMPONENT_LIB} PUBLIC BSP_GRAPHICS)
+target_compile_definitions(${COMPONENT_LIB} PUBLIC BSP_GRAPHICS_LVGL)
```
PetrESP: What is this compile definition for?

components/bsp_graphics/bsp_graphics_lvgl/src/bsp_graphics_lvgl.c
```diff
@@ -0,0 +1,159 @@
...
+* Local variables
+*******************************************************************************/
+static const char *TAG = "BSP Graphics";
+static lv_display_t *disp;
+static lv_indev_t *disp_indev_touch = NULL;
+
+/*******************************************************************************
+* Private functions
+*******************************************************************************/
+static lv_display_t *bsp_display_lcd_init(const bsp_display_cfg_t *cfg)
```
tore-espressif, line 31: @espzav IIUC, this function should be 'generic' and working on all boards. However, it initializes DSI peripheral. Is there a plan to move the DSI specific part to the respective BSP?
Can we make this function more generic so it compile and run on all BSPs?

espzav: @tore-espressif This function returns `lv_display_t`, this is type from LVGL. It initializes LVGL, it cannot be generic.

tore-espressif: Sure, it is tight to LVGL.

But, it also initializes DSI. This will not work on eg. ESP32-S3. Will we have several versions of this funciton? (eg. for DSI, SPI and i8080 backed displays?)

georgik:
I was validating the concept.

- bsp_selector for board selection in menuconfig definitely makes sense.
- bsp_selector for the graphical library selection in menuconfig does not makes sense.

I recommend to remove the selection of NOGLIB and LVGL, because in the project you have a graphical library and they have different APIs, so it does not make sense to switch them in menuconfig, since each would handle graphics differently.

georgik:
@espzav Please check https://github.com/georgik/esp-idf-component-SDL/pull/19 . This it how the usage could look like with SDL3.

Do not worry about amount of changed files, this is because I needed to include the source for components bsp_selector and m5stack_tab5. The important is example `snow` which should demonstrate how to use the concept.
Since the bsp_selector is commited in components directory, it's disabled from idf_component.yml. Once it get published we can enable it. I've tried to pull it from repo, but Component Manager was not able to handle it.

I have similar solution for Raylib. Once we agree on concept, I can push also that one. It would be good if we can somehow publish some version of the selector, either under espressif-dev namespace or something like that, so that we do not need to commit components into component examples like snow.

PR #778, opened 2026-05-07
LVGL port feature updates

espzav:
# ESP-BSP Pull Request checklist

- [x] Version of modified component bumped
- [x] CI passing

# Description
- Add support for RGB LCD for ESP32P4 and other chips with RGB peripheral + updated example
- - Closes #631 
- Updated support for RGB565 swapped color format 
- - Closes #768
- Supported LCD_RGB_ISR_IRAM_SAFE for RGB panels 
- - Closes #732
- - Closes #755
- - Closes #515 
- Add const to event bits

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes display flush/ISR callback wiring and memory allocation behavior for RGB panels (including IRAM-safe paths), which can impact rendering stability and ISR safety across targets/IDF versions.
> 
> **Overview**
> Updates `esp_lvgl_port` to **generalize RGB and MIPI-DSI support using `SOC_*` capability checks** (instead of target-specific guards) and adds explicit IDF-version gating with clearer error messages.
> 
> Improves RGB panel robustness when `CONFIG_LCD_RGB_ISR_IRAM_SAFE` is enabled by moving the RGB vsync callback to IRAM, passing `disp_ctx` directly as ISR `user_ctx`, and allocating the display context from internal RAM in this mode.
> 
> Extends LVGL9 display handling to accept `LV_COLOR_FORMAT_RGB565_SWAPPED` (including for `swap_bytes` and DMA buffer constraints), bumps the component version to `2.8.0`, and updates CI/example coverage by enabling the `rgb_lcd` example for `esp32s3`/`esp32p4` with new per-target `sdkconfig.defaults` and ESP32P4-specific example pin/color-format setup (touch disabled on P4).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bb7bf85d4e2f977850d928f53063788a8d9985e9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #783, opened 2026-05-14
fix(lvgl_port/lvgl8): allow MALLOC_CAP_DMA|MALLOC_CAP_SPIRAM on PSRAM-DMA-capable SoCs

vikramdattu:
## Summary

The lvgl8 port unconditionally rejects `buff_dma && buff_spiram` even on SoCs whose GDMA can reach PSRAM directly (ESP32-S3, ESP32-P4). The lvgl9 port already gates the same check on `SOC_PSRAM_DMA_CAPABLE == 0`. This PR mirrors that pattern in lvgl8 so callers on capable targets can keep a PSRAM-resident canvas without the CPU memcpy detour.

Behavior on targets where `SOC_PSRAM_DMA_CAPABLE == 0` is unchanged — the existing `trans_size > 0` escape hatch still applies.

## Motivation

On ESP32-P4 with a SPIRAM-backed LVGL canvas, this restriction forces either a CPU memcpy through internal RAM each flush (CPU-bound) or staying in internal RAM (memory-pressured during streaming workloads).

## Test plan

- [x] ESP32-P4 (ESP32-P4-EYE board, P4 ≥ rev1, PSRAM enabled, SOC_PSRAM_DMA_CAPABLE=1): boot succeeds with `disp_cfg.flags.buff_dma=true, buff_spiram=true`. Previously asserted with `Alloc DMA capable buffer in SPIRAM is not supported!`.
- [ ] ESP32 / ESP32-S2 path unchanged (compile-only check; behavior identical to pre-patch).

## Notes

- Matches the same logic the PPA driver already does in `common/ppa/lcd_ppa.c`.
- lvgl9 port is unchanged — it already has this gate.

### DIFF
--- a/components/esp_lvgl_port/src/lvgl8/esp_lvgl_port_disp.c
+++ b/components/esp_lvgl_port/src/lvgl8/esp_lvgl_port_disp.c
@@ -279,13 +279,26 @@ static lv_disp_t *lvgl_port_add_disp_priv(const lvgl_port_display_cfg_t *disp_cf
         ESP_GOTO_ON_FALSE(trans_sem, ESP_ERR_NO_MEM, err, TAG, "Failed to create transport counting Semaphore");
         disp_ctx->trans_sem = trans_sem;
     } else {
-        uint32_t buff_caps = MALLOC_CAP_DEFAULT;
+        uint32_t buff_caps = 0;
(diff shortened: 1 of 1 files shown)


espzav (approved):
@vikramdattu Thank you for this update! LGTM
please sign CLA, I will merge it and it will be released after this #778

PR #785, opened 2026-05-18
Minor pytest updates

PetrESP:

- [ ] CI passing

# Change description
Moved common options from CI to the pytest.ini
Added bspwall marker

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -119,11 +119,6 @@ def pytest_generate_tests(metafunc):
         )
 
 
-@pytest.fixture
-def build_dir(build_dir: str) -> str:
-    return f'{build_dir}'
-
-
 # This fixing using cache when used "-n auto" (parallel)
 def pytest_configure(config):
     # If run pytest-xdist (parallel), set unique cache dir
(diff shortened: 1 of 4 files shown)


conftest.py
```diff
@@ -119,11 +119,6 @@ def pytest_generate_tests(metafunc):
         )
 
 
-@pytest.fixture
-def build_dir(build_dir: str) -> str:
```
espzav, line 123: Why we do not need this?

PetrESP: Because it does not seem to do anything. It is basically a wrapper for `pytest-embedded` fixture `build_dir` and it returns it's value. 
It would be used for [something like this](https://docs.pytest.org/en/stable/how-to/fixtures.html#override-a-fixture-on-a-folder-conftest-level). But since we don't do edit the string it does nothing.
Maybe it was supposed to be a placeholder for something or a fix for previous versions.

espzav: ok, when it is working :-)

espzav (approved):
@PetrESP Thank you for improvements :-) LGTM

PR #788, opened 2026-05-27
Add sensor hub to DS18B20

PetrESP:

- [ ] CI passing

# Change description
Added sensor hub API to the DS18B20 driver and an example of using it


components/ds18b20/examples/ds18b20_sensor_hub/main/ds18b20_sensor_hub_example.c
```diff
@@ -0,0 +1,73 @@
...
+#error "No 1-Wire backend selected in menuconfig"
+#endif
+
+    sensor_config_t config = {
+        .type = HUMITURE_ID,
+        .mode = MODE_POLLING,
+        .min_delay = 500,
+        .bus = bus,
+        // The address can be any arbitrary number since it is unused
+        .addr = 22,
```
suda-morris, line 56: actually ds18b20 does have a address, but need to do a enumeration on the bus first.

components/ds18b20/src/ds18b20_sensor_hub.c
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+#include "esp_log.h"
+#include "esp_err.h"
+#include "esp_check.h"
+#include "ds18b20.h"
+#include "iot_sensor_hub.h"
+
+static const char *TAG = "ds18b20_sensor_hub";
+static ds18b20_device_handle_t sensor_hub_ds18b20_handle;
```
suda-morris, line 14: can't the "sensor-hub" accommodate two or more DS18B20 devices? Is there somewhere to save the device handle in a sensor-hub device context? and the context can be passed to the functions like "acquire_temperature".

components/ds18b20/CMakeLists.txt
```diff
@@ -1,2 +1,4 @@
-idf_component_register(SRCS "src/ds18b20.c"
+idf_component_register(SRCS "src/ds18b20.c" "src/ds18b20_sensor_hub.c"
                        INCLUDE_DIRS "include")
+
+target_link_libraries(${COMPONENT_LIB} INTERFACE "-u ds18b20_impl_init")
```
suda-morris: I would suggest introducing a Kconfig gate for the DS18B20 device driver, which determines if the driver can be detected by the sensor-hub framework. And based on that Kconfig, we add this linkage

suda-morris: and in the Kconfig help, we should document the limitaion e.g. only one device can be used, if there're multiple DS18B20 on the bus, it's not supported by the sensor hub library.

components/ds18b20/examples/ds18b20_sensor_hub/main/Kconfig.projbuild
```diff
@@ -0,0 +1,45 @@
...
+            Stop searching after this many DS18B20 devices are found.
+
+    config EXAMPLE_ONEWIRE_ENABLE_INTERNAL_PULLUP
+        bool "Enable internal pull-up resistor on bus GPIO"
+        default y
+        help
+            Enable this when there is no external pull-up resistor on the
+            1-Wire data line.
+
+    choice EXAMPLE_ONEWIRE_BACKEND
```
suda-morris: Since the purpose of this example is to test  the sensor-hub frame work, there's no need to test it against different backend peripheral. Remove the choice, keep it simple and focus.

components/ds18b20/src/ds18b20_sensor_hub.c
PetrESP: Sadly there is no way to pass context to those functions. I will try to think of a way to add it since I think it would be very useful. 
For now I have specified that we will only support a single device for this use case and potentially expand it in the future

components/ds18b20/examples/ds18b20_sensor_hub/main/Kconfig.projbuild
PetrESP: You are right. I've simplified it

components/ds18b20/examples/ds18b20_sensor_hub/main/ds18b20_sensor_hub_example.c
PetrESP: Yes. I hope the new comment is clearer.
I  have to set it to some non zero address because sensor hub checks for it

components/ds18b20/src/ds18b20_sensor_hub.c
suda-morris: I think this is a framework limitation, the interface like

```c
    /**
     * @brief Acquire temperature data from the sensor
     *
     * @param[out] temperature Pointer to store the temperature data
     * @return esp_err_t Result of acquiring data
     */
    esp_err_t (*acquire_temperature)(float *temperature);
```

doesn't even accept a handle! I have reported the issue to them.

suda-morris (approved):

components/ds18b20/src/ds18b20_sensor_hub.c
PetrESP: Yeah this is very limiting. Thank you for that

espzav (approved):
@PetrESP nice work! Thank you, LGTM

PR #790, opened 2026-06-01
ci(doxygen): Install newer version

espzav:

- [x] CI passing

# description
- Install newer version of doxygen


bsp/esp-box-3/API.md
```diff
@@ -405,7 +405,7 @@ Variables:
 
 -  const sdspi\_device\_config\_t \* sdspi  
 
--  union [**bsp\_sdcard\_cfg\_t**](#struct-bsp_sdcard_cfg_t) slot  
+-  anonymous union slot
```
PetrESP, line 408: So this is basically a workaround for the issue where it showed numbers (i guess some  ID) instead of the documentation?

PetrESP (approved):
LGTM

bsp/esp-box-3/API.md
espzav: This is the part of struct (union or struct without name). The new doxygen put number instead of this. Hanxi fixed it in doxybook to `anonymous`.

tore-espressif (approved):
LGTM!

PR #807, opened 2026-07-20
Fix ST7123 display initialization delay parameters

artisdom:
# Change description

The ST7123 init table sends SLPOUT (0x11) and DISPON (0x29) with incorrect delay_ms = 0

MIPI DCS requires waiting 120 ms after Sleep Out: 
during that window the display controller starts its internal oscillator and charge pumps 
and reloads its analog defaults (gamma / VCOM) from NVM. 
Sending further commands inside that window races the analog bring-up, 
and the panel intermittently latches a corrupted analog state. 

On the M5Stack Tab5 (board v2, ST7123 panel) this shows as a randomly-occurring washed-out "white hue" 
across the whole screen while the image content itself is rendered correctly;

Give SLPOUT the DCS-mandated 120 ms and DISPON a 50 ms settle. 
These are the same values used by the esp_lcd_st7123 driver's own default init table

https://github.com/espressif/esp-iot-solution/blob/master/components/display/lcd/esp_lcd_st7123/esp_lcd_st7123.c#L219


espzav (approved):
@artisdom Thank you for fix. LGTM

PR #804, opened 2026-07-10
fix(m5stack_tab5): support ST7121 LCD variant

RiviaRammer:
## 问题背景

M5Stack Tab5 存在不同 LCD 方案。当前 BSP 在 I2C 地址 `0x55` 探测到 ST712x 触摸控制器后，会统一按 ST7123 LCD 初始化。

Tab5 UserDemo 会读取 ST712x 寄存器 `0x0000` 的固件版本区分屏幕：

- `FW version = 1`: ST7121
- `FW version = 3`: ST7123

因此，ST7121 屏幕版本需要使用不同的初始化命令和 DPI 时序。

## 修改内容

- 读取 ST712x FW version，区分 ST7121 和 ST7123
- 保留原有 GT911 / ILI9881C 路径
- ST7121 和 ST7123 均使用现有 `esp_lcd_new_panel_st7123()` driver
- ST7121 使用其专用初始化命令、965 Mbps DSI lane rate 和 DPI timing
- 不新增独立的 `esp_lcd_st7121` component

## 测试

- 已使用 ESP-IDF v5.4.3 重新构建 Tab5 bring-up demo
- ELF 仅包含 `esp_lcd_new_panel_st7123`，不包含 `esp_lcd_new_panel_st7121`
- 官方工程师已使用此实现方式在实体 ST7121 和 ST7123 Tab5 上测试，显示动画和触摸正常，未见花屏、闪烁或重启
- `git diff --check` 通过

---

> [!NOTE]
> **Medium Risk**
> Changes runtime hardware detection and MIPI/DPI init for display bring-up; mis-detection could break display on some Tab5 units, but scope is limited to the M5Stack Tab5 BSP.
> 
> **Overview**
> Adds a third M5Stack Tab5 hardware path for **ST7121** panels, which were previously lumped in with ST7123 after I2C touch probe at `0x55`.
> 
> **Board detection** now reads ST712x touch firmware from register `0x0000` (FW `1` → ST7121, FW `3` → ST7123) and replaces magic integers with a `bsp_tab5_board_version_t` enum. Unknown or failed probe paths return errors instead of asserting where appropriate.
> 
> **Display bring-up** selects ST7121-specific MIPI init tables (`disp_init_data_st7121.h`), DPI timing, and **965 Mbps** DSI lane rate while still using `esp_lcd_new_panel_st7123()`. GT911/ILI9881C and ST7123 behavior is unchanged. BSP package version bumps to **1.3.0**.
> 
[...]

PR #808, opened 2026-07-27
fix(m5stack_tab5): don't return NULL from int bsp_get_board_version()

vikramdattu:
`bsp_get_board_version()` in `bsp/m5stack_tab5/src/bsp_display.c` returns `int`, but uses `BSP_ERROR_CHECK_RETURN_NULL`. When `CONFIG_BSP_ERROR_CHECK` is disabled, that macro expands to `return NULL;`, so the file fails to compile with `-Wint-conversion` (an error with current toolchains):

```
bsp_display.c:138:5: error: returning 'void *' from a function with return type 'int' makes integer from pointer without a cast [-Wint-conversion]
```

Any project that sets `CONFIG_BSP_ERROR_CHECK=n` (return-error-code style instead of abort) cannot build the Tab5 BSP.

Fix: use `BSP_ERROR_CHECK(x, 0)`. With `CONFIG_BSP_ERROR_CHECK` enabled the behaviour is unchanged (`ESP_ERROR_CHECK`); with it disabled the function returns 0 ("version unknown") and, since `board_ver` only caches values > 0, the probe retries on the next call.

Hit in an app build (ESP-IDF v5.5.4, esp32p4, m5stack_tab5 1.2.0~1, `CONFIG_BSP_ERROR_CHECK=n`). Verified the pre-fix code reproduces the exact error under `-Werror=int-conversion` and the fixed code compiles clean.

---

> [!NOTE]
> **Low Risk**
> Localized BSP display/touch init fix and clearer error handling; behavior unchanged when error-check abort mode is enabled.
> 
> **Overview**
> Fixes a **build break** when `CONFIG_BSP_ERROR_CHECK=n`: `bsp_get_board_version()` now uses `BSP_ERROR_CHECK(..., 0)` instead of `BSP_ERROR_CHECK_RETURN_NULL`, so it no longer emits `return NULL` from an `int` function (`-Wint-conversion`).
> 
> Display and touch setup **cache** the board version once per path and **fail with `ESP_ERR_NOT_SUPPORTED`** if probing did not yield version 1 or 2, instead of calling `bsp_get_board_version()` repeatedly and branching with separate `== 2` checks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8789c2374af81c842b5e9d6312df48027e9417f2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/bsp/m5stack_tab5/src/bsp_display.c
+++ b/bsp/m5stack_tab5/src/bsp_display.c
@@ -135,7 +135,7 @@ static int bsp_get_board_version(void)
     }
 
     /* Enable Feature */
-    BSP_ERROR_CHECK_RETURN_NULL(bsp_feature_enable(BSP_FEATURE_TOUCH, true));
+    BSP_ERROR_CHECK(bsp_feature_enable(BSP_FEATURE_TOUCH, true), 0);
 
     vTaskDelay(pdMS_TO_TICKS(500));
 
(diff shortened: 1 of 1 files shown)


espzav:
@vikramdattu Thank you for this fix. You are right. 
We are waiting for this PR: #804 it can fix this issue too. But there is still discussion.

espzav:
Closing, due to fixed in #804

PR #814, opened 2026-08-10
fix(lcd_touch): Require esp_lcd_touch when the component manager is off

nebkat:

- [ ] CI passing

# Change description
Every touch driver includes esp_lcd_touch.h but declares the dependency only in idf_component.yml, which nothing acts on when the component manager is disabled, leaving the include path unset and the build broken.

The requirement is added only in that case: with the component manager enabled the dependency is registered as espressif__esp_lcd_touch, so an unconditional REQUIRES esp_lcd_touch would name a component that does not exist in that path (the same split esp_lvgl_port already handles via BUILD_COMPONENTS).

---

> [!NOTE]
> **Low Risk**
> Build-system dependency wiring only; no runtime or driver logic changes.
> 
> **Overview**
> Fixes builds of six LCD touch drivers when the IDF component manager is disabled. Those drivers include `esp_lcd_touch.h` but previously declared the dependency only in `idf_component.yml`, so include paths were missing without the manager.
> 
> Each driver's `CMakeLists.txt` now checks `IDF_COMPONENT_MANAGER` and adds `esp_lcd_touch` to `REQUIRES` only in that case (avoiding a name clash with the managed `espressif__esp_lcd_touch` component). Component versions are bumped accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4df6ea55593668302db2127b6b41760d5d1b5ea1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/components/lcd_touch/esp_lcd_touch_cst816s/CMakeLists.txt
+++ b/components/lcd_touch/esp_lcd_touch_cst816s/CMakeLists.txt
@@ -1,1 +1,10 @@
-idf_component_register(SRCS "esp_lcd_touch_cst816s.c" INCLUDE_DIRS "include" REQUIRES "driver" "esp_lcd")
+set(REQ "driver" "esp_lcd")
+
+# esp_lcd_touch is declared in idf_component.yml, which is only acted upon by the
+# component manager. Add it explicitly when the component manager is disabled.
+idf_build_get_property(idf_component_manager IDF_COMPONENT_MANAGER)
+if(NOT idf_component_manager)
+    list(APPEND REQ esp_lcd_touch)
+endif()
+
+idf_component_register(SRCS "esp_lcd_touch_cst816s.c" INCLUDE_DIRS "include" REQUIRES ${REQ})
--- a/components/lcd_touch/esp_lcd_touch_cst816s/idf_component.yml
+++ b/components/lcd_touch/esp_lcd_touch_cst816s/idf_component.yml
@@ -1,4 +1,4 @@
-version: "1.1.1~2"
+version: "1.1.2"
 description: ESP LCD Touch CST816S - touch controller CST816S
 url: https://github.com/espressif/esp-bsp/tree/master/components/lcd_touch/esp_lcd_touch_cst816s
 dependencies:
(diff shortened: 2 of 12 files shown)


espzav:
HI @nebkat, I am not sure, why component manager should be disabled. All Our projects are made for component manager. We are using component manager as main.

nebkat:
Hi @espzav,

In our application we do not use component manager as we develop agricultural equipment and often find ourselves in areas without internet - which causes issues with component manager when trying to build. In addition we prefer the strong guarantees of git submodule references rather than relying on the component manager versioning.

It is an officially supported [setting](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/tools/idf-component-manager.html#disabling-the-component-manager) and I have made similar [PRs](https://github.com/espressif/esp-usb/pull/509) before so would appreciate if you could consider it!

---

The one remaining problem to make it fully usable as a submodule is:
https://github.com/espressif/esp-bsp/blob/f0ef9497efce684997ce391edd19733483e250a5/components/lcd/esp_lcd_ili9341/CMakeLists.txt#L6-L7

`package_manager` is defined in [esp_iot_solution](https://github.com/espressif/esp-iot-solution/blob/master/tools/cmake_utilities/package_manager.cmake) which is the one repository we avoid submoduling as it has grown so large. We have temporarily solved this by breaking out these tools to their own [repo](https://github.com/nebkat/esp-cmake-utilities), but it would be great if they were somewhere shared (other than esp-iot-solution) if they are going to be relied on by many components.

espzav:
Hi @nebkat ok. Thank you for clarification and for contribution. We can merge it, but please update minor versions like:

from: 
version: "1.1.1~2"
to:
version: "1.1.2"

espzav (approved):
@nebkat Thank you! LGTM

espzav:
@nebkat Please submit CLA license, I cannot merge it.

nebkat:
Done now, thanks!

PR #818, opened 2026-08-18
fix(bsp): Replaced lv_disp_rotation_t with lv_display_rotation_t

axos88:
## Problem

`bsp_display_rotate()` takes an `lv_disp_rotation_t`, which no longer exists in LVGL. Every affected BSP therefore fails to compile against current LVGL `master`:

```
error: unknown type name 'lv_disp_rotation_t'; did you mean 'lv_display_rotation_t'?
  bsp/esp32_s31_korvo_1/include/bsp/esp32_s31_korvo_1.h:601
```

LVGL 9.5.0 (the newest release on the component registry as of August 2026) still builds, so this only bites people who track `master` - which, for ESP32-S31, is currently necessary if you want the PPA draw unit (see lvgl/lvgl#10481).

## Why the name disappeared

`lv_disp_rotation_t` was never part of LVGL v8. It was a mistake in LVGL's own v8 compatibility header, fixed by **lvgl/lvgl#10095** (commit `879e7b9f6`, merged 11 May 2026):

```diff
-typedef lv_display_rotation_t       lv_disp_rotation_t;
+typedef lv_display_rotation_t       lv_disp_rot_t;
```

v8's real identifier was `lv_disp_rot_t`, so the alias these BSPs relied on was simply wrong, and removing it was a legitimate fix rather than a breaking change.

## Fix

Use `lv_display_rotation_t`, which is correct under **both** major versions:

- LVGL 9: it is the native name.
- LVGL 8: `components/esp_lvgl_port/include/esp_lvgl_port_compatibility.h` already provides it via `typedef lv_disp_rotation_t lv_display_rotation_t;`.

That compatibility header deliberately defines the LVGL 8 shim, so it is **left untouched** and is now the only place in the repository that mentions the old name.

## Scope

- 18 BSPs (header, source and `API.md`), plus two `esp_lvgl_port` examples/tests
- Component revisions bumped for the BSPs, following the `~N` convention used for fixes
- `esp_lvgl_port` itself is unchanged (only its examples/tests), so its version is not bumped

## Verification

Built the LVGL demo project for **ESP32-S31-Korvo-1 V1.1** against LVGL `master` with the PPA draw unit enabled. Before this change the build required patching this identifier by hand; after it, it succeeds unmodified.

I only had ESP32-S31-Korvo-1 hardware to hand, so the other 17 BSPs are verified by inspection - the change is a pure identifier rename with no behavioural component.

Two notes for reviewers:

- **doxygen was not available on my machine**, so the `doxygen-api-md` pre-commit hook could not run. The `API.md` files were updated with the same textual rename instead of being regenerated; please regenerate if the output differs. All other hooks (astyle, copyright, codespell, README/board-table updaters) pass and changed nothing.
- **The 18 version bumps are easy to drop** if you would rather handle versioning at release time.

---

> [!NOTE]
> **Low Risk**
> Identifier-only API alignment with no logic changes; LVGL 8 still gets `lv_display_rotation_t` via esp_lvgl_port compatibility.
> 
> **Overview**
> Updates **`bsp_display_rotate()`** and related docs across **18 BSPs** so the rotation parameter uses **`lv_display_rotation_t`** instead of **`lv_disp_rotation_t`**, matching current LVGL and fixing builds against LVGL master where the old type was removed.
> 
> The same type rename is applied in **`esp_lvgl_port`** touchscreen/test code and the **`display_rotation`** example. Each touched BSP gets a **`idf_component.yml`** patch revision bump (`~N`); **`esp_lvgl_port`** itself is unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8dd20773ff660759be43106594812c236571e253. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/bsp/esp-box-3/idf_component.yml
+++ b/bsp/esp-box-3/idf_component.yml
@@ -1,4 +1,4 @@
-version: 3.2.0
+version: 3.2.0~1
 description: Board Support Package (BSP) for ESP32-S3-BOX-3
 url: https://github.com/espressif/esp-bsp/tree/master/bsp/esp-box-3
 
--- a/bsp/esp-box-3/API.md
+++ b/bsp/esp-box-3/API.md
@@ -933,7 +933,7 @@ Below are some of the most relevant predefined constants:
 |  lv\_indev\_t \* | [**bsp\_display\_get\_input\_dev**](#function-bsp_display_get_input_dev) (void) <br>_Get pointer to input device (touch, buttons, ...)._ |
 |  bool | [**bsp\_display\_lock**](#function-bsp_display_lock) (uint32\_t timeout\_ms) <br>_Take LVGL mutex._ |
 |  esp\_err\_t | [**bsp\_display\_new**](#function-bsp_display_new) (const [**bsp\_display\_config\_t**](#struct-bsp_display_config_t) \*config, esp\_lcd\_panel\_handle\_t \*ret\_panel, esp\_lcd\_panel\_io\_handle\_t \*ret\_io) <br>_Create new display panel._ |
-|  void | [**bsp\_display\_rotate**](#function-bsp_display_rotate) (lv\_display\_t \*disp, lv\_disp\_rotation\_t rotation) <br>_Rotate screen._ |
+|  void | [**bsp\_display\_rotate**](#function-bsp_display_rotate) (lv\_display\_t \*disp, lv\_display\_rotation\_t rotation) <br>_Rotate screen._ |
 |  lv\_display\_t \* | [**bsp\_display\_start**](#function-bsp_display_start) (void) <br>_Initialize display._ |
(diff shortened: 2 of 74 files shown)


bsp/esp-box-3/idf_component.yml
```diff
@@ -1,4 +1,4 @@
-version: 3.2.0
+version: 3.2.0~1
```
espzav, line 1: The code was changed --> 
```suggestion
version: 3.2.1
```
- same for all changes in BSPs

espzav (commented):
@axos88 Thank you for this PR. Please update versions by my comment and resolve conflicts + rebase from master. 
LGTM

